PR #78, opened 2021-03-26
Adding server debugging support in launch.json and extension.ts, added sourceMap to tsconfig

erinsinger93:
I've been using this locally to step through language server code while testing out interactions in the dummy apps. `autoAttachChildProcesses` in the launch config lets us setup debugger ports for the server processes.

If this feature isn't something that interests others I can keep this on my local branch.


packages/environment-ember-loose/-private/environment/index.js.map
```diff
@@ -0,0 +1 @@
+{"version":3,"file":"index.js","sourceRoot":"","sources":["index.ts"],"names":[],"mappings":";;AAEA,SAAwB,mBAAmB;IACzC,OAAO;QACL,QAAQ,EAAE;YACR,SAAS,EAAE,6CAA6C;YAExD,sBAAsB,CAAC,YAAY;gBACjC,IAAI,qBAAqB,CAAC,IAAI,CAAC,YAAY,CAAC,EAAE;oBAC5C,gBAAgB;oBAChB,OAAO,CAAC,YAAY,CAAC,OAAO,CAAC,gBAAgB,EAAE,cAAc,CAAC,CAAC,CAAC;iBACjE;qBAAM;oBACL,sBAAsB;oBACtB,OAAO,CAAC,YAAY,CAAC,OAAO,CAAC,QAAQ,EAAE,KAAK,CAAC,CAAC,CAAC;iBAChD;YACH,CAAC;YAED,wBAAwB,CAAC,UAAU;gBACjC,IAAI,qBAAqB,CAAC,IAAI,CAAC,UAAU,CAAC,EAAE;oBAC1C,gBAAgB;oBAChB,OAAO,CAAC,UAAU,CAAC,OAAO,CAAC,gBAAgB,EAAE,cAAc,CAAC,CAAC,CAAC;iBAC/D;qBAAM;oBACL,sBAAsB;oBACtB,OAAO,CAAC,UAAU,CAAC,OAAO,CAAC,OAAO,EAAE,MAAM,CAAC,CAAC,CAAC;iBAC9C;YACH,CAAC;SACF;KACF,CAAC;AACJ,CAAC;AA1BD,sCA0BC"}
```
dfreeman, line 1: Can we `.gitignore` these?

jamescdavis:
I actually set this up locally too. I think we should commit it. I'm good with turning on source maps, but don't think we want to commit the map files.

.vscode/launch.json
```diff
@@ -3,8 +3,9 @@
   "configurations": [
     {
       "name": "Debug Extension",
-      "type": "extensionHost",
+      "type": "pwa-extensionHost",
```
dfreeman, line 6: My google skills are failing me—what's the difference between `extensionHost` and `pwa-extensionHost` here?

dfreeman (commented):
Thanks @erinsinger93, this seems like a super useful addition!

.vscode/launch.json
jamescdavis: I'm curious too, but hadn't even tried to Google it yet. 🙂

packages/environment-ember-loose/-private/environment/index.js.map
jamescdavis: +1

.vscode/launch.json
erinsinger93: It targets a newer javascript debugger that has a richer feature set -- the regular `extensionHost` type doesn't support the `autoAttachChildProcesses` configuration. The newer debugger is installed by default in recent versions of vscode -- people on really old versions of vscode would need to install it manually

Honestly, the documentation seems to be pretty sparse. Here's the newer debugger library: https://github.com/microsoft/vscode-js-debug

Let me know if you have more questions -- I can dig into the specifics if there are concerns.

dfreeman: Nope, that makes sense—thanks for the explainer! 😄

dfreeman (approved):

PR #79, opened 2021-03-27
Show using `glint` to replace `tsc --noEmit`

chriskrycho:
Previously, the README said that it was possible to do this, but did not make clear *how* to do this.


dfreeman (approved):
Thanks @chriskrycho!

PR #80, opened 2021-03-28
Fix `each` not working with a `readonly` array

boris-petrov:
I can't seem to run tests locally (i.e. most of them fail) so let's see what the CI says.


dfreeman (approved):
Good catch; thanks @boris-petrov!

dfreeman:
P.S. did you run `yarn build` before executing the tests locally? I'd expect most of them except for maybe the `@glint/config` suite to fail if not.

PR #112, opened 2021-03-30
Support inverse for {{#each-in}} - Fixes #108

### LINKED ISSUE #108: `{{#each-in}}` should support inverse
```
app/components/feature-flag-toggle.hbs:41:5 - error TS2345: Argument of type '{ default(key: string, flag: Feature): void; inverse(): void; }' is not assignable to parameter of type '{ default: (key: string, value: Feature) => void; }'.
  Object literal may only specify known properties, and 'inverse' does not exist in type '{ default: (key: string, value: Feature) => void; }'.

 41     {{#each-in this.features.flags as |key flag|}}
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 42       {{#if flag.isVisible}}
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 85       </p>
    ~~~~~~~~~~
 86     {{/each-in}}
```

```hbs
    {{#each-in this.features.flags as |key flag|}}
      {{#if flag.isVisible}}
        <div class="feature-group {{dasherize key}}">
          {{!-- snip --}}
        </div>
      {{/if}}
    {{else}}
      <p>
        There are no current experiments. New experiments will be announced via Insider emails.
      </p>
    {{/each-in}}
```

wagenet:


packages/environment-ember-loose/__tests__/intrinsics/each-in.test.ts
```diff
@@ -9,3 +9,38 @@ invokeBlock(eachIn({}, { a: 5, b: 3 }), {
     expectTypeOf(value).toEqualTypeOf<number>();
   },
 });
+
+// Can render inverse when undefined, null, or empty.
```
dfreeman, line 13: Can you add a test against a maybe-undefined object (e.g. `{ a: number } | undefined`)? I'm not sure we'll infer the correct `key` and `value` types the way the default block's params are specified right now.

dfreeman (commented):
Thanks @wagenet! Just one question around making sure we handle unions right

PR #115, opened 2021-03-31
Support SafeString - Fixes #93

### LINKED ISSUE #93: Support SafeString
```
app/components/trace-header-row.hbs:24:19 - error TS2345: Argument of type 'SafeString' is not assignable to parameter of type 'string | number | boolean | void | AcceptsBlocks<{}> | null'.

24             style={{bar.style}}
```

```hbs
          <div
            class="bar {{bar.category}}"
            style={{bar.style}}
          >
          </div>
```

wagenet:


packages/template/__tests__/custom-invokable.test.ts
```diff
@@ -68,6 +68,13 @@ invokeBlock(resolve(caseOf)({}, maybeValue), {
   },
 });
 
+// SafeString
+invokeEmit({
```
dfreeman, line 72: Can you also add a test to the `environment-ember-loose` suite to ensure that the result of an `htmlSafe()` call works?

It should, given how the types are defined today, but since the definitions are totally separate from each other, it would be good to have a check in place to prevent regression if they ever drift in the future.

dfreeman (approved):
Thanks!

PR #117, opened 2021-03-31
Fix Ember Modifier class

wagenet:
I'm not sure how to test this.


dfreeman (approved):
🤦‍♂️ thank you for fixing this!

What we'd discussed doing when setting up the reexports was including usage of each exported value in the `test-packages/ts-ember-app` package, which would cause that build to fail if something was broken. Apparently we never did that for modifiers, though, so I'm going to go take care of that right now.

PR #127, opened 2021-04-01
Better handling for null and undefined in {{get}}

wagenet:


packages/environment-ember-loose/-private/intrinsics/get.d.ts
```diff
@@ -3,6 +3,11 @@ import ObjectProxy from '@ember/object/proxy';
 
 export type GetHelper = DirectInvokable<{
   <T, K extends keyof T>(args: EmptyObject, obj: T, key: K): T[K];
-  <T extends object, K extends keyof T>(args: EmptyObject, obj: ObjectProxy<T>, key: K): T[K];
+  <T extends object, K extends keyof T>(
+    args: EmptyObject,
+    obj: ObjectProxy<T> | null | undefined,
+    key: K
+  ): T[K];
```
dfreeman, line 10: This should be `T[K] | undefined`, right?

dfreeman (approved):
Thanks for getting this sorted out!

PR #134, opened 2021-04-03
Properly handle attributes and modifiers for yielded components - Fixes #133

### LINKED ISSUE #133: Can't apply attributes or modifiers to yielded components
It generates something like this:

```ts
applyAttributes<ElementForComponent<typeof Foo?.MyComponent>>({ foo: 'bar' });
```

`typeof Foo?.MyComponent` does not work.

wagenet:


wagenet:
I'm not sure if my changes to `emitPathContents` are really ideal, but I believe this works.

dfreeman:
Since this issue should be taken care of in #138 (and the subsequent #149 followup), I'm going to go ahead and close this out. Thanks for your work on it!

PR #149, opened 2021-04-09
Ensure type narrowing propagates downward

### LINKED ISSUE #143: `if` type narrowing is lost inside further nested blocks
image

dfreeman:
## Background

In #138, we reworked the way components and vanilla HTML elements were emitted. This had several nice effects, but it also exacerbated a problem that was already lurking.

As a general rule, TypeScript discards any narrowing information about a mutable variable or field inside a closure that captures it, because there's no way for it to guarantee that the narrowing still applies. Concretely:

```ts
function myFunction(input: { foo: string | number }) {
  if (typeof input.foo === 'string') {
    // `input.foo` is a string here

    [1, 2, 3].forEach(() => {
      // `input.foo` is back to `string | number` here
    });
  }
}
```

In the example above, even though we as humans can say it's unlikely that `input.foo` will change before that `forEach` callback is invoked, TypeScript has no way of knowing that. For all it (and, technically, us) knows, `forEach` might stash that function away and call it again an hour from now, and `input.foo` could have been changed to a number in that time.

Prior to #138, we used functions to represent blocks passed to a component, so if you narrowed the type of a value outside a block and then tried to use it inside, the result of the narrowing would be lost. With #138, that problem also began to apply to attributes and modifiers passed to elements and components, which made the issue much more obvious, as seen in #143.

## The Fix

Since the code Glint generates is never actually executed, using callbacks the way we do isn't actually accomplishing much other than introducing a nested block scope, and that's something we can do ourselves instead.

The general change here goes from the form:

```ts
emitComponent(..., 𝛄 => {
  applyAttributes(𝛄.element, ...);
});
```

To:

```ts
{
  const 𝛄 = emitComponent(...);
  applyAttributes(𝛄.element, ...);
}
```

Here `𝛄` enters and leaves scope in the same places as before, and can continue to be shadowed by deeper blocks as appropriate. However, since there's nary a callback in sight, any type information produced by narrowing will continue to be available in further nested blocks.

Blocks passed to components work similarly, going from:

```ts
bindBlocks(𝛄.blockParams, {
  header() {
    // header stuff
  },
  body(item, index) {
    // body stuff
  }
});
```

To:

```ts
{
  const [] = 𝛄.blockParams.header;
  // header stuff
}
{
  const [item, index] = 𝛄.blockParams.body;
  // body stuff
}
```

### (No More) Mandatory Blocks

With this change we drop the notion of a 'mandatory' block. Previously if you wrote `Yields: { default: [] }`, a consumer would be _required_ to invoke your component with a default block. This always struck me as being of questionable value (the error message was awful), but it was a natural side effect of how we were passing blocks before so I left it.

Now we treat all declared blocks as available but not required, and it's up to the consumer to decide whether to make use of them, which maps neatly to the reality at runtime (`yield`-ing to a block you weren't passed is a no-op). Anecdotally this also feels better to me, as I've frequently found myself having to go back and stick a `?` on a block declaration because I'd inadvertently made it required.

That said, if others feel differently we can revisit and probably construct a synthetic way to enforce that mandatory blocks are actually used.

### Narrowing Safety

Because components can't store blocks and invoke them at some later point in time, this change is safe—any narrower type information we infer at one level _will_ still be true further down.

However, there's always the potential for that to change in the future. The original named blocks proposal did allow them to be reified into JS-accessible objects, but there were similar lifetime concerns for blocks that led to the trimmed down "yieldable-only" version that shipped.

If that changes in the future we may need to either revisit this specifically for blocks, or see if it's reasonable to advocate for a syntactic differentiator between yieldable blocks and reified ones.

Fixes #143, closes #146 (thanks for the failing tests, @wagenet! they're in the base commit here)

### DIFF
--- a/packages/core/__tests__/cli/check.test.ts
+++ b/packages/core/__tests__/cli/check.test.ts
@@ -40,6 +40,65 @@ describe('CLI: single-pass typechecking', () => {
     expect(checkResult.stderr).toEqual('');
   });
 
+  test('handles conditionals with yielding', async () => {
+    project.write('.glintrc', 'environment: ember-loose\n');
+
+    let script = stripIndent`
+      import Component from '@glint/environment-ember-loose/ember-component';
+
+      export type MyComponentArgs = {
+        foo?: () => {};
+        bar?: string;
+        baz?: { value: string };
+      };
+
+      export default class MyComponent extends Component<{ Args: MyComponentArgs }> { }
+    `;
+
+    let template = stripIndent`
+      {{#if @foo}}
+        <button {{on "click" @foo}} type="button"></button>
+      {{/if}}
+      {{#if @bar}}
+        <LinkTo @route="my-route">
+          <OtherComponent @value={{@bar}} />
+        </LinkTo>
+      {{/if}}
+      {{#if @baz}}
+        <LinkTo @route="my-route">
+          <OtherComponent @value={{@baz.value}} />
+        </LinkTo>
+      {{/if}}
+    `;
+
+    let otherScript = stripIndent`
+      import Component from '@glint/environment-ember-loose/ember-component';
+
+      export type OtherComponentArgs = {
+        value: string;
+      };
+
+      export default class OtherComponent extends Component<{ Args: OtherComponentArgs }> { }
+
+
+      declare module '@glint/environment-ember-loose/registry' {
+        export default interface Registry {
+          OtherComponent: typeof OtherComponent;
+        }
+      }
+    `;
+
+    project.write('my-component.ts', script);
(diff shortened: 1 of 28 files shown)


wagenet:
Thanks! I feel like we lose a little utility by not having required blocks, but at the same time, blocks aren't ever really required as far as HBS is concerned, so this seems acceptable.

dfreeman:
That's about where I landed. Making blocks requirable wasn't a feature that was ever specifically designed for, it just fell out of the implementation. Since there's nothing at runtime that actually cares either way, it feels ok to me to lose that even if it is a drop in expressiveness. And it's always something we can revisit down the line 🙂

PR #153, opened 2021-04-14
Fix `get` returning `unknown` for objects that can be `undefined`

### LINKED ISSUE #152: `get` returns `unknown` when the object could be `undefined`
Component:

```typescript

interface Foo {
  readonly [key: string]: number | undefined;
}

export default class GlimmerComponent extends Component<{}> {
  public foo?: Foo;
}
```

```hbs
{{test-helper-1 (get this.foo "asd")}}
```

Helper (just to have something that accepts `number | undefined`):

```typescript
import { helper } from '@glint/environment-ember-loose/ember-component/helper';

function testHelper1(params: [number | undefined]): number | undefined {
  return params[0];
}

const testHelper1Helper = helper(testHelper1);

export default testHelper1Helper;

    'test-helper-1': typeof testHelper1Helper;
  }
}
```

Instead of compiling, that errs with:

```
app/components/glimmer-component.hbs:1:17 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'number | undefined'.
  Type 'unknown' is not assignable to type 'number'.

1 {{test-helper-1 (get this.foo "asd")}}
                  ~~~~~~~~~~~~~~~~~~~~
```

Replacing `public foo?: Foo` with `public foo!: Foo` makes the error go away.

boris-petrov:
Fixes #152 

I think?

### DIFF
--- a/packages/environment-ember-loose/-private/intrinsics/get.d.ts
+++ b/packages/environment-ember-loose/-private/intrinsics/get.d.ts
@@ -3,6 +3,7 @@ import ObjectProxy from '@ember/object/proxy';
 
 export type GetHelper = DirectInvokable<{
   <T, K extends keyof T>(args: EmptyObject, obj: T, key: K): T[K];
+  <T, K extends keyof T>(args: EmptyObject, obj: T | null | undefined, key: K): T[K] | undefined;
   <T extends object, K extends keyof T>(
     args: EmptyObject,
     obj: ObjectProxy<T> | null | undefined,
--- a/packages/environment-ember-loose/__tests__/intrinsics/get.test.ts
+++ b/packages/environment-ember-loose/__tests__/intrinsics/get.test.ts
@@ -19,6 +19,22 @@ get(
   'hi'
 );
 
+interface Foo {
+  readonly [key: string]: number | undefined;
+}
+
+declare const maybeFoo: Foo | undefined;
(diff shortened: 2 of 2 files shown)


packages/environment-ember-loose/-private/intrinsics/get.d.ts
```diff
@@ -2,7 +2,7 @@ import { DirectInvokable, EmptyObject } from '@glint/template/-private/integrati
 import ObjectProxy from '@ember/object/proxy';
 
 export type GetHelper = DirectInvokable<{
-  <T, K extends keyof T>(args: EmptyObject, obj: T, key: K): T[K];
+  <T, K extends keyof T>(args: EmptyObject, obj: T | null | undefined, key: K): T[K];
```
dfreeman, line 5: Thanks for the PR! I think we need two separate signatures for this, because `{{get maybeString 'length'}}` won't give me a `number` if `maybeString` is undefined.

Probably something like this?
```suggestion
  <T, K extends keyof T>(args: EmptyObject, obj: T, key: K): T[K];
  <T, K extends keyof T>(args: EmptyObject, obj: T | null | undefined, key: K): T[K] | undefined;
```

And then we should update the test accordingly to cover both cases.

boris-petrov: Done, I force-pushed the changes. Thanks for the help!

dfreeman (approved):
Thank you!

PR #154, opened 2021-04-15
Introduce `PositionalArgs` for component signatures

### LINKED ISSUE #105: Support positional params on Ember Components
```
{{#async-await this.promise as |value|}}
    <SynchronousComponent @value={{value}} />
  {{else}}
    <LoadingSpinner />
  {{/async-await}}
```

https://github.com/tildeio/ember-async-await-helper/blob/master/addon/components/async-await.js

dfreeman:
While not super common, classic Ember components (as well as those that use custom component managers) can accept positional args when invoked with curlies. This PR adds `PositionalArgs` as a valid signature member for `EmberComponent` and `ComponentLike` in order to capture those cases.

Note: while the static property that's used to configure this at runtime on an Ember component is called `positionalParams`, I went with `PositionalArgs` for consistency with helper and modifier signatures (as well as `Args` itself on the component signature).

I also noticed in working on this that we had types tests for `EmberComponent`'s Glint integration, but not for `GlimmerComponent` or `ComponentLike`, so I added coverage for those as well.

Fixes #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,10 @@ export default class Greeting extends Component<GreetingSignature> {
 {{yield (concat @message ", " this.greetingTarget "!")}}
 ```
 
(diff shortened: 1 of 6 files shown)


chriskrycho (approved):
This makes good sense to me. From a practical POV, we may want to supply *lint rules* to help with when each of these is actually useful, but that's an Ember + TS concern, *not* a Glint concern, and is broader than this specific change.

PR #155, opened 2021-04-15
Remove `Record`-based constraints on signature members

dfreeman:
Currently, people who have `class MyComponent extends Component<MyArgs>` and are adopting Glint are naturally inclined to change that to `Component<{ Args: MyArgs }>`. Unfortunately, if the `MyArgs` type is declared as an `interface`, that change will fail with a confusing error:

```
Type '{ Args: MyArgs; }' does not satisfy the constraint 'ComponentSignature'.
  Types of property 'Args' are incompatible.
    Type 'MyArgs' is not assignable to type 'Record<string, unknown>'.
      Index signature is missing in type 'MyArgs'.(2344)
```

The underlying reason for this is that TS will infer an index signature for type aliases, but is unwilling to do so for interfaces because [that's more unsafe due to the potential for declaration merging](https://github.com/microsoft/TypeScript/issues/15300#issuecomment-332366024). Needing to migrate `interface`s to `type` declarations is an annoying adoption papercut, particularly given that it's not apparent from the error message that that's a viable fix.

In practice this constraint doesn't provide a _whole_ lot of value over using `object` instead, particularly relative to the number of folks who've already run into this index signature issue. We may consider some type-aware linting in the future, but for now this should remove some real pain for people working to adopt Glint.

### DIFF
--- a/packages/environment-ember-loose/-private/utilities.ts
+++ b/packages/environment-ember-loose/-private/utilities.ts
@@ -28,9 +28,9 @@ export type YieldsOf<C extends ComponentLike> = C extends Constructor<
  * README for further details.
  */
 export type ComponentSignature = {
-  Args?: Partial<Record<string, unknown>>;
+  Args?: object;
   PositionalArgs?: Array<unknown>;
-  Yields?: Partial<Record<string, Array<unknown>>>;
+  Yields?: object;
   Element?: Element | null;
 };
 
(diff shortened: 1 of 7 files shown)


chriskrycho (approved):

chriskrycho:
If only TS would actually let us describe these semantics usefully. 😩

PR #162, opened 2021-04-16
Add tests for `in-element`

boris-petrov:


dfreeman:
The docs say `element` is a [required parameter that "must exist at the time of rendering"](https://api.emberjs.com/ember/3.26/classes/Ember.Templates.helpers/methods/in-element?anchor=in-element). It sounds like the runtime behavior says otherwise, but can we confirm with someone on the VM or framework team that that's intentionally supported?

boris-petrov:
@dfreeman - funny, I always thought it was working with `null` or `undefined` and I just now hit the assertion that this is not true. I've changed the PR to just include tests.

dfreeman:
Awesome, thank you @boris-petrov!

PR #163, opened 2021-04-16
Don't prevent subclassing of components with companion templates

### LINKED ISSUE #161: Cannot extend a Glimmer component
Base component:

```typescript

export default class GlimmerComponent extends Component<{}> {}
```

Child component:

```typescript
import GlimmerComponent from './glimmer-component';

export default class ChildGlimmerComponent extends GlimmerComponent {}
```

Both should have an HBS file (empty for example). That leads to:

```
app/components/child-glimmer-component.ts:3:22 - error TS2417: Class static side 'typeof ChildGlimmerComponent' incorrectly extends base class static side 'typeof GlimmerComponent'.
  Types have separate declarations of a private property ''~template''.

3 export default class ChildGlimmerComponent extends GlimmerComponent {}
                       ~~~~~~~~~~~~~~~~~~~~~
```

dfreeman:
We evaluate companion templates from `.hbs` files by treating them as an implicit static property on their backing class. We synthesize this property as `private`, which causes problems if you subclass a component: TypeScript (reasonably) rejects inheritance hierarchies with the same private property declared at multiple levels.

For now we can fix this by changing `private` to `protected`, which also means we no longer need to emit an extra reference to it to avoid an "unused property" diagnostic. Longer term we'll hopefully be able to implement this in a [class static block](https://github.com/tc39/proposal-class-static-block) instead, assuming that proposal continues to advance.

Fixes #161


chriskrycho (approved):
💙

PR #166, opened 2021-04-16
Introduce `@glint-ignore` and `@glint-expect-error` directives

### LINKED ISSUE #82: [Feature request] Ignore lines in HBS with `{{! glint-ignore }}` or something
Ability to ignore lines in HBS would be great.

dfreeman:
In TypeScript, it's possible to tell the typechecker to either silence any errors that begin on a particular line (`@ts-ignore`) or to specifically expect an error and fail if none is produced (`@ts-expect-error`). While it's generally preferable to have code that typechecks correctly, these tools are useful to have available for cases where expressiveness of the type system falls short.

This PR introduces equivalent directives for Glint. Unfortunately we can't directly rely on the TS implementation of these, as there's no one-to-one mapping between lines in the input template and lines in our translated TypeScript.

Closes #82


chriskrycho (approved):
:shipit:

PR #167, opened 2021-04-16
Readme: Info on adding nested components to registry

### LINKED ISSUE #158: Nested component have to have their full nested name for the registry
This might just be a readme fix, but I wanted to report it in case something isn't working as it should.

Let's say I have a component in the following tree:

```
app/components/group/members-table/cells/name-and-email
```

In the component file, I have: 

```ts
export default class GroupMembersTableCellsNameAndEmail extends Component<Signature> {}

```

I would have thought I just did this to register it in the registry:
```ts
declare module "@glint/environment-ember-loose/registry" {
    GroupMembersTableCellsNameAndEmail: typeof GroupMembersTableCellsNameAndEmail
  }
}
```

but it appears I have to use the full nested name to get it to correctly type check:

```diff
declare module "@glint/environment-ember-loose/registry" {
  export default interface Registry {
-    GroupMembersTableCellsNameAndEmail: typeof GroupMembersTableCellsNameAndEmail
+   "Group::MembersTable::Cells::NameAndEmail": typeof GroupMembersTableCellsNameAndEmail
  }
}

```

Is that working as designed? If so, I can submit a PR to update the Readme to clarify this.

scottmessinger:
@dfreeman @wagenet Thanks for your comments! This should fix #158.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,30 @@ declare module '@glint/environment-ember-loose/registry' {
 }
 ```
 
+If you've nested your component inside folder(s), you'll need to add the full "strong" name of the component to the registry. And, if you're expecting to invoke your component with currlies or the `{{componet}}` helper, you'll need to add the component to the registry twice -- once using the `::` delimiter syntax and once using the `/` delimiter syntax. For example:
+
+```ts
+// app/components/grouping/my-component.ts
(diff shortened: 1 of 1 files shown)


chriskrycho (approved):
Thanks for this!

PR #175, opened 2021-04-20
Prefer `else` over `inverse`

wagenet:
Requires `Yields` to use `else`, but still allows for `{{yields to="inverse"}}`


wagenet:
See also https://github.com/glimmerjs/glimmer-vm/pull/1296

dfreeman:
Thanks @wagenet (and apologies for immediately creating a merge conflict!)

Can you add a note to the README where we talk about `Yields` mentioning that `inverse` is an alias for `else` (and that you should use the latter in your types?)

dfreeman (approved):

PR #177, opened 2021-04-21
Skip `actions/cache` for Windows CI

dfreeman:
The decompression step for `actions/cache` on Windows is [known to be slow](https://github.com/actions/cache/issues/442), and in our case it's significantly slower than just installing dependencies from scratch.

In representative sample runs, we spent:

|Step|Linux | Windows | Windows without `actions/cache` |
|-|-|-|-|
| Restore Cache | 0:07 | 3:03 | 0:00 |
| Install Dependencies | 0:22 | 0:40 | 2:18 |
| Save Cache | 0:06 | 0:11 | 0:00 |
| **Total** | **0:37** | **3:36** | **2:18** |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,15 +56,6 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 12
-      - name: Locate Yarn Cache
-        id: yarn-cache-dir-path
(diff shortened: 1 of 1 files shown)


chriskrycho (approved):
Much 🤦🏼‍♂️ , Windows.

PR #178, opened 2021-04-21
Support routes, controllers and classic-layout components

dfreeman:
## Overview

This PR introduces support for checking route/controller templates. It also enables resolution of classic-layout components.

## How Route/Controller Templates Work

The data available in a route template's context depends on whether a controller is available or not. 

In a route-only template, `this` will have type `void`, and there will only be one argument available: `@model`, defined to have type `Awaited<ReturnType<TheRoute['model']>>`.

In a template with a controller, `this` will be the instance type of the controller, and `@model` will be defined to be the type of the controller's `model` property.

## Internal Changes

When examining a given script, we need to be able to unambiguously determine whether its default export is the backing class for a template. Since the backing class for a route template is its controller if one exists or is the route otherwise, `@glint/environment-ember-loose` needs a way when looking at a route to say "this is a backing class unless a controller exists."

To enable such cases, a call to `getPossibleTemplatePaths('routes/foo.ts')` can now return a result like this:
```ts
[{ path: 'templates/foo.hbs', deferTo: ['controllers/foo.ts'] }]
```


chriskrycho (commented):
Noted this on Discord, but leaving it here for posterity/watchers/etc.:

> In a route-only template, `this` will have type `void`, and there will only be one argument available: `@model`, defined to have type `Awaited<ReturnType<TheRoute['model']>>`.

This isn't quite right; `this` should have a type roughly like `Controller & { model: Awaited<ReturnType<TheRoute['model']>> }`. Ember generates a controller at runtime if one isn't supplied (which is how e.g. `loading` and `error` templates work); you can see the resulting behavior in [this Twiddle](https://ember-twiddle.com/49d46eb3f10d8f8727e006d9abf540ec?openFiles=routes.application%5C.js%2C), which has an application route but no controller; `this.model` still resolves to the resulting model. :weary:

dfreeman:
Thanks for catching that @chriskrycho! That's now [fixed (and properly tested)](https://github.com/typed-ember/glint/compare/f0aef227e44bea9382f54770d1034be0d042202f..49b588c4a62bf3983e27ac3201bfe5d3b1f7bd9e)

PR #179, opened 2021-04-22
Avoid cascading errors on invalid component invocations

dfreeman:
Similar in spirit to #173, this PR ensures that trying to invoke an invalid component doesn't produce a cascade of errors that mask the true problem.

As an example of this, prior to this PR if I were to typo the name of a component in my template, in addition to the "that's not a valid key of `Globals`" error, I'd see a couple of others relating to trying to pass a default block to an unknown component.

image

This change ensures we produce `any` where appropriate in those situations so that we avoid a cascade and don't bury the real error.

### DIFF
--- a/packages/template/-private/dsl/emit.d.ts
+++ b/packages/template/-private/dsl/emit.d.ts
@@ -50,8 +50,8 @@ export declare function emitElement<Name extends string>(
 export declare function emitComponent<T extends AcceptsBlocks<any, any>>(
   component: T
 ): {
-  element: T extends AcceptsBlocks<any, infer El> ? El : null;
-  blockParams: T extends AcceptsBlocks<infer Yields, any> ? Required<Yields> : never;
+  element: T extends AcceptsBlocks<any, infer El> ? El : any;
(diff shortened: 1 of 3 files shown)


chriskrycho (commented):
Is there any risk from this that we will end up falling back to `any` in cases where we *don't* have an error like this?

chriskrycho (approved):
We chatted in Discord. For other readers, the key is:

```ts
export declare function emitComponent<T extends AcceptsBlocks<any, any>>(
```

If users fail to pass something which maps to *that* type, it will error at the component invocation site, and downstream *usage* of the resulting values (element and block params) will *not* error. By analogy (@dfreeman's explanation to me), if you do this:

```ts
for (let i in 123) {
  let j = i + 1;
}
```

…TypeScript will complain about the *actual* error, which is that `123` is not iterable, and then will treat `i` as `any` so that you don't get meaningless errors at `let j = i + 1`: the error there would be meaningless noise to the user, and TS can't possibly give a meaningful error at that point regardless because it has no way to understand what type `i` *should* have.

LET’S DO THIS.

PR #187, opened 2021-04-28
Add support for `bubbles` in `link-to`

boris-petrov:
https://github.com/emberjs/ember.js/blob/v3.26.1/packages/@ember/-internals/glimmer/lib/components/link-to.ts#L702

### DIFF
--- a/packages/environment-ember-loose/-private/intrinsics/link-to.d.ts
+++ b/packages/environment-ember-loose/-private/intrinsics/link-to.d.ts
@@ -14,6 +14,7 @@ type LinkToArgs = RequireAtLeastOne<
     disabled?: boolean;
     activeClass?: string;
     'current-when'?: string | boolean;
+    bubbles?: boolean;
     preventDefault?: boolean;
     tagName?: string;
   },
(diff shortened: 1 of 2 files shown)


dfreeman:
Can you get confirmation from anyone on the Ember team that that's meant to be public? It makes sense to me that it might be since `preventDefault` is, but it's [not documented](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/input?anchor=LinkTo).

boris-petrov:
I'm not sure who to ping for that (you probably would know better than me :smile: ) but right, I'm not exactly sure it's public... event though there [are](https://discuss.emberjs.com/t/disable-a-link-to-for-an-action-inside-it/13530/9) many [references](https://github.com/emberjs/ember.js/issues/1636) to [it](https://github.com/emberjs/ember.js/issues/19144) on the web.

I'll leave the decision to you. :D

dfreeman:
I would ask in Discord, either #general or possibly #dev-ember-js if you suspect it's a bug that this isn't documented

boris-petrov:
`bubbles` is actually deprecated as of Ember 3.27 so this PR is unneeded.

PR #194, opened 2021-05-27
Support standalone templates as components in GlimmerX

dfreeman:
## Overview

In GlimmerX, a standalone template is a valid component:

```js
const Foo = hbs`
  Hello, {{@target}}!
`;

export default hbs`
  <Foo @target="World" />
  <Foo @target="Everyone" />
`;
```

This PR enables these template components to work in Glint, using a type annotation to associate a signature where needed:

```ts
interface FooSignature {
  Args: { target: string };
}

const Foo: TemplateComponent<FooSignature> = hbs`
  Hello, {{@target}}!
`;

export default hbs`
  <Foo @target="World" />
  <Foo @target="Everyone" />
`;
```

Just as with class-based components, template components that don't declare a signature are assumed to accept no args, have no splattributes element, and never yield.

## Change Details

This PR tweaks the emit from `@glint/transform` slightly in order for type inference to work for the template contents.

Where previously we had:

```ts
(() => {
  let χ!: typeof import("dsl-path");
  return χ.template(function(𝚪: import("dsl-path").ResolveContext<MyComponent>) {
    // template contents
    𝚪;
  });
})();
```

Now we emit:

```ts
({} as typeof import("dsl-path")).template(
  function(
    𝚪: import("dsl-path").ResolveContext<MyComponent>,
    χ: typeof import("dsl-path")
  ) {
    // template contents
    𝚪; χ;
  }
);
```

Removing the IIFE layer means that the type of `𝚪` can be inferred based on where the template is being assigned/passed, which is how writing `const Foo: TemplateComponent<Signature> = hbs...` works above. Accordingly, we no longer annotate `𝚪` during emit if we don't have a definite context type (as we would in a class body).

## Other Notes

This work is the first step to supporting Ember's "template-only components", since it gets our ducks in a row for templates with no backing class. From here, we can also make the changes necessary to support typechecking for templates in rendering tests before tackling standalone template files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ TypeScript-powered tooling for Glimmer templates.
   - [With GlimmerX](#with-glimmerx)
     - [Import Paths](#import-paths)
     - [Component Signatures](#component-signatures)
+    - [Template Components](#template-components)
   - [With Ember.js](#with-emberjs)
     - [Import Paths](#import-paths-1)
     - [Component Signatures](#component-signatures-1)
@@ -130,6 +131,30 @@ export class Shout extends Component<ShoutSignature> {
 ```
 
 [yieldable named blocks rfc]: https://github.com/emberjs/rfcs/blob/master/text/0460-yieldable-named-blocks.md
+
+#### Template Components
+
+In GlimmerX, standalone templates can be invoked as components. Like class-based components, you can declare a signature for a template component in order for Glint to understand how it can be used.
+
+The class-based component above could as a template component like so:
+
+```ts
+import { TemplateComponent } from '@glint/environment-glimmerx/component';
+
+interface ShoutSignature {
+  /* same as above */
+}
+
+const shout = (message: string) => `${message.toUpperCase()}!`;
+
+const Shout: TemplateComponent<ShoutSignature> = hbs`
+  <div ...attributes>
+    {{yield (shout @message)}}
+  </div>
+`;
+```
+
+Note that, similar to React's `FunctionComponent` and `FC`, you can also import and use the `TC` type alias as a shorthand for `TemplateComponent`.
 
 ### With Ember.js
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -131,6 +132,30 @@ export class Shout extends Component<ShoutSignature> {
...
 [yieldable named blocks rfc]: https://github.com/emberjs/rfcs/blob/master/text/0460-yieldable-named-blocks.md
 
+#### Template Components
+
+In GlimmerX, standalone templates can be invoked as components. Like class-based components, you can declare a signature for a template component in order for Glint to understand how it can be used.
+
+The class-based component above could as a template component like so:
+
+```ts
+import { TemplateComponent } from '@glint/environment-glimmerx/component';
```
chriskrycho, line 142: Suggestion: we should recommend `import type` over `import` wherever we're importing a type-only type, just for clarity.

```suggestion
import type { TemplateComponent } from '@glint/environment-glimmerx/component';
```

README.md
```diff
@@ -131,6 +132,30 @@ export class Shout extends Component<ShoutSignature> {
...
+
+const Shout: TemplateComponent<ShoutSignature> = hbs`
+  <div ...attributes>
+    {{yield (shout @message)}}
+  </div>
+`;
+```
+
+Note that, similar to React's `FunctionComponent` and `FC`, you can also import and use the `TC` type alias as a shorthand for `TemplateComponent`.
+
```
chriskrycho, line 158: Suggested change (can't use an actual suggestion because nested code blocks like that aren't a thing!):

> Note that, similar to React's `FunctionComponent` and `FC`, you can also import and use the `TC` type alias as a shorthand for `TemplateComponent`:
> 
> ```ts
> import type { TC } from '@glint/environment-glimmerx/component';
> ```

chriskrycho (approved):

chriskrycho:
Thank you as always for the detailed PR writeup!

PR #200, opened 2021-06-25
Get TS nightly passing (once more, with feeling)

dfreeman:
There's been some churn lately in `typescript@next` around what gets included under `strict: true`.
 - `useUnknownInCatchVariables` is in
 - `exactOptionalPropertyTypes` (née `strictOptionalProperties`) is out

This has resulted in some churn for us in the "watch for breaking changes coming down the pipe" stesp in CI, but this PR should get us green again.


dfreeman:
Sorry for the conflicts this is going to produce @erinsinger93! It should get TS nightly passing on your branch the next time you merge `master` though :)

Issue #192, opened 2021-05-19
Deprecation warnings with Ember 3.27


boris-petrov:
Using [global](https://github.com/typed-ember/glint/blob/0.5.1/packages/environment-ember-loose/ember-component/helper.ts#L10) Ember is deprecated in Ember 3.27 and will be removed in Ember 4.

Issue #190, opened 2021-05-10
`link-to` cannot be used with the `component` helper


boris-petrov:
```hbs
{{#let (component "link-to" route="asd" model=123) as |Link|}}
  <Link></Link>
{{/let}}
```

Leads to:

```
app/components/bar.hbs:1:19 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '"link-to"' is not assignable to parameter of type '(new (...args: unknown[]) => Invokable<(args: unknown) => AcceptsBlocks<any, any>>) | null | undefined'.

1 {{#let (component "link-to" route="asd" model=123) as |Link|}}
                    ~~~~~~~~~
```

boris-petrov:
@dfreeman - I seem to be getting the exact same error even with Glint 0.6.1. Am I missing something or is this not fixed?

dfreeman:
You're right! See #211

Issue #186, opened 2021-04-28
Wrong error message when an attribute is a substring of another


boris-petrov:
Component `Foo`:

```typescript
import Component from '@glint/environment-ember-loose/glimmer-component';

export interface IFoo {
  readonly Args: {
    readonly a: number;
    readonly aa: string;
  };
}

export default class Foo extends Component<IFoo> {}

declare module '@glint/environment-ember-loose/registry' {
  export default interface Registry {
    Foo: typeof Foo;
  }
}
```

Component `Bar`:

```typescript
import Component from '@glint/environment-ember-loose/glimmer-component';

export default class Bar extends Component {}
```

```handlebars
<Foo @aa="asd" @a="qwe" />
```

Leads to:

```
app/components/bar.hbs:1:7 - error TS2322: Type 'string' is not assignable to type 'number'.

1 <Foo @aa="asd" @a="qwe" />
        ~

  app/components/foo.ts:5:14
    5     readonly a: number;
                   ~
    The expected type comes from property 'a' which is declared here on type '{ readonly a: number; readonly aa: string; }'
```

Note the first squiggly line - it's on `aa` instead of on `a`.

boris-petrov:
@dfreeman - there are two more cases where this breaks:

1)

```hbs
{{foo aa="qwe" a="qwe"}}
```

```
app/bar.hbs:1:7 - error TS2322: Type 'string' is not assignable to type 'number'.

1 {{foo aa="qwe" a="qwe"}}
        ~

  app/foo.ts:5:14
    5     readonly a: number;
                   ~
    The expected type comes from property 'a' which is declared here on type '{ readonly a: number; readonly aa: string; }'
```

2)

```hbs
<Foo
  @aa={{@a}}
  @a="qwe" />
```

Where `bar` now has this definition:

```ts
export default class Bar extends Component<{ Args: { a: string } }> {}
```

```
app/bar.hbs:2:10 - error TS2322: Type 'string' is not assignable to type 'number'.

2   @aa={{@a}}
           ~

  app/foo.ts:5:14
    5     readonly a: number;
                   ~
    The expected type comes from property 'a' which is declared here on type '{ readonly a: number; readonly aa: string; }'
```

dfreeman:
It would be so nice if `@glimmer/syntax` would give us the actual offsets for attrs rather than us having to reverse-engineer them 🙃 

Thanks for the additional report!

dfreeman:
[...]

Issue #181, opened 2021-04-26
`each-in` sets `string | number` for the key's type


boris-petrov:
```typescript
import Component from '@glint/environment-ember-loose/glimmer-component';

export default class Foo extends Component {
  public declare readonly c?: { readonly [key: string]: number };
}
```

```handlebars
{{#each-in this.c as |a|}}
  {{sum a 2}}
{{/each-in}}
```

And a generic `sum` helper:

```typescript
import Helper from '@glint/environment-ember-loose/ember-component/helper';

export interface ISum {
  readonly PositionalArgs: number[];
  readonly Return: number;
}

export default class Sum extends Helper<ISum> {
  public compute(numbers: ISum['PositionalArgs']): ISum['Return'] {
    return numbers.reduce((a, b) => a + b, 0);
  }
}

declare module '@glint/environment-ember-loose/registry' {
  export default interface Registry {
    sum: typeof Sum;
  }
}
```

Leads to:

```
app/components/foo.hbs:2:9 - error TS2345: Argument of type 'string | number' is not assignable to parameter of type 'number'.
  Type 'string' is not assignable to type 'number'.

2   {{sum a 2}}
          ~
```

Ignore the specific error - the problem is that `a` is typed as `string | number` instead of what I've said it can be - `key: string`.

dfreeman:
It looks like this is intentional behavior of `keyof` in TypeScript itself: https://github.com/microsoft/TypeScript/issues/29249

I want to read a bit more about their justification there, but I think for `{{#each-in}}` it probably makes sense for us to paper over that behavior and do something like `keyof T & string`.

PR #203, opened 2021-07-19
Assorted bugfixes

dfreeman:
This PR contains bugfixes for a handful of outstanding issues:
 - #181 
 - #186 
 - #190
 - #192

It also includes tweaks to `toDebugString` to make it more useful and an upgrade to Ember 3.27 in `ts-ember-app` to ensure everything works cleanly with the deprecated `Ember` global.


jamescdavis (approved):
I went through each of the the bugfixes and they all make sense and look good. It took me a minute to understand the change made in https://github.com/typed-ember/glint/pull/203/commits/39ae1efbcf68ca067601390a284cf26e253b6344 but I see now that, IIUC, it's to report the location of tokens in tagged templates relative to the whole file rather than relative to the start of the template string.

dfreeman:
> it's to report the location of tokens in tagged templates relative to the whole file

Yep, that's it exactly. That code isn't ever actually used in 'production', but it's super helpful during dev and debugging, and I've been meaning to fix that reported offset discrepancy for a long time—it makes it super annoying to diagnose what's going on in failing tests 😄

Issue #201, opened 2021-07-12
Update README for js support for glimmer language server


erinsinger93:

dfreeman:
Closed by #419 (thank you @chriskrycho!)

PR #207, opened 2021-07-20
Add `checkStandaloneTemplates` configuration option

dfreeman:
In most scenarios, a template inherits the "typedness" of the script it's attached to. If the backing module is TS, anything Glint sees as illegal usage in the template will get a red squiggle. If it's JS, we'll never give any squiggles, only provide go-to-definition, hover info, etc.

With template-only components, though, we have no way of knowing whether the author intends for the template to be typechecked or not. Currently we always treat standalone templates as targets for typechecking, but that's of course undesirable in JS-only projects.

This PR adds a `checkStandaloneTemplates` flag that users can specify in `.glintrc` to specify whether standalone templates should be typechecked or not.

This option should likely be documented as part of the larger "how to use Glint in JS projects" docs effort described in #201, as it doesn't really make sense outside that context. Does that seem reasonable to you @erinsinger93?

### DIFF
--- a/packages/core/src/common/document-cache.ts
+++ b/packages/core/src/common/document-cache.ts
@@ -45,7 +45,7 @@ export default class DocumentCache {
     }
 
     if (isTemplate(path)) {
-      return synthesizedModulePathForTemplate(path);
+      return synthesizedModulePathForTemplate(path, this.glintConfig);
     }
   }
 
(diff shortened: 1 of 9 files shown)


erinsinger93 (approved):
makes sense to me

packages/core/src/common/document-cache.ts
```diff
@@ -127,8 +127,9 @@ export function templatePathForSynthesizedModule(path: string): string {
   return path.replace(SCRIPT_EXTENSION_REGEX, '.hbs');
 }
 
-export function synthesizedModulePathForTemplate(path: string): string {
-  return path.replace(TEMPLATE_EXTENSION_REGEX, '.ts');
+export function synthesizedModulePathForTemplate(path: string, config: GlintConfig): string {
+  let extension = config.checkStandaloneTemplates ? '.ts' : '.js';
```
erinsinger93, line 131: One thought: Is this naming a bit confusing? In the event that a consumer has a JS-only project but they want `checkJs: true`, they'd want the JS extension on their standalone templates but they'd also want to typecheck the file still

PR #208, opened 2021-07-21
Update README to include documentation for Glint in projects using JS

erinsinger93:
First stab at adding documentation for folks onboarding to Glint who are either using a JS-only environment or an environment in the process of migrating to TS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ TypeScript-powered tooling for Glimmer templates.
     - [Template-Only Components](#template-only-components)
     - [Rendering Tests](#rendering-tests)
     - [Contextual Components](#contextual-components)
+  - [With JS rather than TS](#with-js-rather-than-ts)
 - [Known Limitations](#known-limitations)
   - [Environment Re-exports](#environment-re-exports)
   - [Ember-Specific](#ember-specific)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -474,6 +475,40 @@ const Glimmer: AcceptsAFooString = class extends GlimmerComponent<{ Args: { foo:
...
+```js
+// SimpleComponent.js
+
+import Component from '@glint/environment-glimmerx/component';
+import { hbs } from '@glimmerx/component';
+import { helper } from '@glint/environment-glimmerx/helper';
+
+const or = helper(
+  /** @param {[a: *, b: *]} param */
+  ([a, b]) => a || b
```
dfreeman, line 491: Not sure if it's worth getting this exact in an example like this, but writing out the type parameters for this helpler would allow the language server to understand that `{{or}}`'s return value will be the type of one of its inputs (inferring `A | B` as the return type):

```js
const or = helper(
  /**
   * @template A
   * @template B
   * @param {[a: A, b: B]} positional
   */
  ([a, b]) => a || b
);
```

dfreeman (commented):
Thank you, this is looking great!

dfreeman:
Oh one other thing that just occurred to me! I believe we still say somewhere at the top of the README that we don't yet support JS projects—that's no longer true 🙂

chriskrycho:
@erinsinger93 if you happen to see this, just wanted to say thanks again for writing it up! We merged all of your content into our docs in an update today. Really appreciate it!

PR #218, opened 2021-09-21
Allow explicitly passing `undefined` for `disabled` on `link-to`

### LINKED ISSUE #217: Glint doesn't work with the `exactOptionalPropertyTypes` TypeScript option
```ts
import type { ModelOrObjectPromiseProxy } from 'frontend/services/store/promise-proxies';

export interface IFoo {
  readonly Args: {
    readonly disabled?: boolean | undefined;
  };
}

export default class Foo extends Component<IFoo> {
}
```

```hbs
<LinkTo @route="foo" @disabled={{@disabled}}></LinkTo>
```

This leads to (be sure to use the `exactOptionalPropertyTypes` option):

```
app/components/foo.hbs:1:4 - error TS2322: Type 'boolean | undefined' is not assignable to type 'boolean'.                            
  Type 'undefined' is not assignable to type 'boolean'.                                                                                                       
                                                                                                                                                              
4   @disabled={{@disabled}}                                                                                                                                   
     ~~~~~~~~                                                                                                                                                 
                                                                                                                                                              
  node_modules/@glint/environment-ember-loose/-private/intrinsics/link-to.d.ts:14:5
    14     disabled?: boolean;
           ~~~~~~~~
    The expected type comes from property 'disabled' which is declared here on type 'LinkToArgs'
```

boris-petrov:
Fixes #217 

Not sure how to test this. Ideas?

### DIFF
--- a/packages/environment-ember-loose/-private/intrinsics/link-to.d.ts
+++ b/packages/environment-ember-loose/-private/intrinsics/link-to.d.ts
@@ -11,7 +11,7 @@ type LinkToArgs = RequireAtLeastOne<
     model?: unknown;
     models?: unknown[];
     query?: Record<string, unknown>;
-    disabled?: boolean;
+    disabled?: boolean | undefined;
     activeClass?: string;
     'current-when'?: string | boolean;
     preventDefault?: boolean;


dfreeman:
I'm going to close out this PR since the types as implemented match the Ember docs (see also the discussion on #217). If you want to PR a docs update to explicitly make `undefined` a valid value for `disabled`, I'm happy to revisit this once that lands 🙂

PR #235, opened 2021-11-22
Add overloads to `fn` with more arguments

boris-petrov:
... because sometimes you have functions with more arguments. :)


dfreeman:
👍  Seems reasonable! If you pull in the most recent changes on `master` that should hopefully get CI green so we can merge 🙂

boris-petrov:
@dfreeman - done!

PR #251, opened 2022-01-20
Support custom file extensions

dfreeman:
## Overview

This PR enables Glint environments to declare that additional file extensions should be supported when typechecking projects.

Note that this change does _not_ allow for custom syntax yet—files with custom extensions must nevertheless still be valid JS/TS or Handlebars, depending on the extension's declared kind. Support for preprocessing of custom extensions will come in a following PR; there was enough complexity in this change to warrant landing it on its own.

## Details

There are two major changes here, and the rest is largely adapting to those changes as they ripple out through the codebase.

The first is introducing a notion of custom extension configuration to `GlintEnvironment`, allowing the active environment to answer questions like "is this file a template?". This in turn means the rest of Glint needs to refer to the environment to determine how to treat different files, rather than using static utility functions.

The second change is updating the `DocumentCache` to allow a many-to-one mapping of paths to a single document, which corresponds to an actual resolvable module. For instance, with the GJS proposal, `components/foo.gts` and `components/foo.ts` would both correspond to the import specifier `components/foo`, and at runtime only one of them would 'win out' if both were present. Even putting the potential for conflict resolution aside, the TS language service deals exclusively in `.js` and `.ts` files, so when we see `foo.gts` on disk, we present it to the language service as `foo.ts`, meaning we must later be able to map back to `foo.gts` when we get e.g. a diagnostic for that module from the language service.

In addition to those two changes in Glint itself, this PR also introduces `test-packages/glint-environment-custom-test` and `test-packages/ts-custom-app`. The former is a Glint environment that defines a `.custom` file extension which is treated as TS, and the latter is a test application that makes use of that environment.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,9 +10,7 @@
       "runtimeExecutable": "${execPath}",
       "args": [
         "--extensionDevelopmentPath=${workspaceFolder}/packages/vscode",
-        "${workspaceFolder}/test-packages/ts-ember-app",
-        "${workspaceFolder}/test-packages/ts-glimmerx-app",
-        "${workspaceFolder}/test-packages/js-glimmerx-app"
+        "${workspaceFolder}/test-packages/glint-test-packages.code-workspace"
       ]
     }
   ]
--- a/packages/config/__tests__/environment.test.ts
+++ b/packages/config/__tests__/environment.test.ts
@@ -11,7 +11,9 @@ describe('Environments', () => {
         },
       });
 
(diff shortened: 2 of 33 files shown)


dfreeman:
WINDOWSSSSSS you can't be this slow _and_ this flaky 😫

image

Putting this on hold while I try to figure out a strategy to avoid losing Windows coverage entirely while also keeping CI actually, like, useful.

dfreeman:
CI is passing, and #252 (might have) improved the overall situation, so: onward!

PR #253, opened 2022-01-21
Overhaul global scope management

### LINKED ISSUE #182: Globals are shadowed by block parameters whose name they are a prefix of
```typescript

export default class Foo extends Component {
  public get bars(): Array<{ text: string }> | undefined {
    return undefined;
  }
}

import Helper from '@glint/environment-ember-loose/ember-component/helper';

declare class OrHelper<T> extends Helper<{
  readonly PositionalArgs: T[];
  readonly Return: T;
}> {}

    or: typeof OrHelper;
  }
}
```

```handlebars
{{#let (get this.bars 1) as |original|}}
  {{or original.text ""}}
{{/let}}
```

Leads to:

```
app/components/foo.hbs:2:5 - error TS2304: Cannot find name 'or'.

2   {{or original.text ""}}
      ~~
```

The crazy thing is that changing the name of the `original` variable to something else that doesn't start with `original` (say `bar`) - the error goes away. :smile:

dfreeman:
## Overview

This change overhauls the way environments declare how their exported tags handle assumptions about scope capture and template globals.

Today, we track the set of statically-known identifiers in scope as we traverse a template, potentially seeding that scope with the available set of JS identifiers depending on whether the environment's tag config declares `capturesOuterScope: true`. We then assume anything that _doesn't_ appear in that scope must be global.

## The Problem

Today's approach largely works, but by default it means autocomplete doesn't work for non-global identifiers, since upon seeing `{{fo}}` we'd emit `Globals['fo']` even if `foo` is an in-scope block param. We've baked in some heuristic workarounds for this, but they lead to issues like #182.

Additionally, this approach requires us to do a full scope analysis of every JS/TS file we parse in order to populate the initial set of in-scope variables for a template. Babel spends a nontrivial amount of time doing this analysis for us.

## This Change

With this change, environments that support scope capture (like GlimmerX) can now supply a static list of known globals, allowing us to assume anything else is a local identifier without needing to do scope analysis. Environments that have no scope capture and instead rely on global resolution (like loose-mode Ember) can provide no `globals` array and have everything assumed global as today (though without the problematic heuristic mentioned above).

## Technical Details

This is a breaking change in that it alters the public interface that environment packages expose to Glint in a backwards-incompatible way. It will also produce slightly different behavior for completions in strict-mode environments.

One notable caveat to this change is that strict-mode environments must be able to statically enumerate the template globals they make available. If a consumer adds additional globals (either by registering them with the VM at runtime or as build-time template macros), they need a way to inform the environment of that. Accordingly, this PR also allows environments to accept custom configuration.

```yml
# Today's setup; will continue to work:
environment: glimmerx

# With this change, users can also do the following:
environment:
  glimmerx:
    additionalGlobals: ['abc', 'def']
```

In anticipation of environment-specific configuration, environments today already export a function that returns their definition. With this change, that function will receive any custom user config (or `{}` otherwise) as its first argument when invoked. The GlimmerX environment has been updated to accept exactly the `additionalGlobals` config as shown here, but any arbitrary key-value data will be passed along.

Fixes #182

### DIFF
--- a/packages/config/__tests__/environment.test.ts
+++ b/packages/config/__tests__/environment.test.ts
@@ -194,28 +194,28 @@ describe('Environments', () => {
       expect(env.getConfiguredTemplateTags()).toEqual({ internal: {} });
     });
 
+    function createEnvironment(config: string): string {
+      let name = Math.random().toString(36).slice(2);
+      let dir = `${testDir}/node_modules/@glint/environment-${name}`;
+
+      fs.mkdirSync(dir, { recursive: true });
+      fs.writeFileSync(`${dir}/env.js`, `module.exports = ${config}`);
+      fs.writeFileSync(
+        `${dir}/package.json`,
+        JSON.stringify({
+          name: `@glint/environment-${name}`,
+          'glint-environment': 'env',
+        })
+      );
+
+      return name;
+    }
+
     describe('merging multiple environments', () => {
-      function createEnvironment(config: string): string {
-        let name = Math.random().toString(36).slice(2);
-        let dir = `${testDir}/node_modules/@glint/environment-${name}`;
-
-        fs.mkdirSync(dir, { recursive: true });
-        fs.writeFileSync(`${dir}/env.js`, `module.exports = () => (${config})`);
-        fs.writeFileSync(
-          `${dir}/package.json`,
-          JSON.stringify({
-            name: '@glint/environment-a',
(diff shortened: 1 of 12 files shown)


chriskrycho (approved):
This is a really nice improvement – not only in the fundamentals but in the way it simplifies the internals. 👍🏼

PR #273, opened 2022-03-31
Fix Helper types compat

jamescdavis:
This patches `packages/environment-ember-loose/ember-component/helper.ts` to have it's version of `Helper` include the same `Opaque` that `@ember/component/helper` extends, allowing it to continue to work with `@types/ember__component@4.0.6`.

### DIFF
--- a/packages/environment-ember-loose/ember-component/helper.ts
+++ b/packages/environment-ember-loose/ember-component/helper.ts
@@ -1,3 +1,4 @@
+import { Opaque } from 'ember/-private/type-utils';
 import { Ember } from './-ember';
 import type { Invoke, Invokable, EmptyObject } from '@glint/template/-private/integration';
 import type { AsObjectType } from '../-private/utilities';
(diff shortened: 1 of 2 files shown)


packages/environment-ember-loose/ember-component/helper.ts
```diff
@@ -30,7 +31,7 @@ export interface HelperSignature {
 const Helper = EmberHelper as AsObjectType<typeof EmberHelper> &
   (new <T extends HelperSignature>(
     ...args: ConstructorParameters<EmberHelperConstructor>
-  ) => Helper<T>);
+  ) => Helper<T> & Opaque<T>);
```
dfreeman, line 34: I think casting through `unknown` would be a bit more consistent with what we're doing with `helper` and in other declarations, and is probably a slightly safer way to handle this than adding properties on to the type here until TS considers it 'similar enough' to be castable.

```ts
const Helper = EmberHelper as unknown as AsObjectType<typeof EmberHelper> & ...
```

With the signature information the upstream type has, we could actually clean this up a lot and remove most or all of the casts entirely, but since the plan is to get rid of all these reexports in the next day or two anyway, it's probably not worth the effort 🙂

dfreeman: In retrospect, the second part of my comment maybe makes the first part irrelevant—this fixes solves the immediate issue, so let's :shipit:

dfreeman:
Thanks @jamescdavis!

PR #275, opened 2022-04-01
Ensure compatibility with latest upstream Ember/Glimmer types

dfreeman:
This PR ensures Glint is compatible with the recent signature-aware releases of `@glimmer/component`, `ember-modifier`, and `@types/ember__component`. ~It's intended to be the last release in the `0.7.x` line.~

The only actual bugfix is a change to our reexports for `ember-modifier`. Prior to this PR, `tsc` (for whatever reason) was inlining `(...args: ConstructorParameters<typeof Modifier>)` to `(owner: unknown, args: import("ember-modifier").ModifierArgs<unknown>)`, which breaks with `ember-modifier@3`, where `ModifierArgs` doesn't take any type parameters.

Pulling `ModifierConstructor` into its own type prevents this behavior, for reasons I don't 100% understand. Normally that would concern me, but this _should_ be the penultimate release of Glint which includes these reexports, and I've verified emit works as intended.


jamescdavis (approved):
Weird. 😆

PR #279, opened 2022-04-05
Fix issue where Component super args were incorrect

wagenet:
This makes types less safe, but previously, TypeScript would squash
the type param making it incorrect.

Before
```ts
declare type ComponentConstructor = {
    new <T extends ComponentSignature = {}>(...args: ConstructorParameters<GlimmerComponentConstructor>): Component<T>;
};
declare const Component: StaticSide<typeof import("@glimmer/component").default> & ComponentConstructor;
```
After:
```ts
declare type ComponentConstructor = {
    new <T extends ComponentSignature = {}>(owner: unknown, args: unknown): Component<T>;
};
declare const Component: StaticSide<typeof import("@glimmer/component").default> & ComponentConstructor;
```


wagenet:
Instead of merging into main, this should be its own branch for 0.7.3.

dfreeman:
I'm going to close this since #280 landed. Thanks again!

PR #280, opened 2022-04-05
Fix issue where Component super args were incorrect

wagenet:
This makes types less safe, but previously, TypeScript would squash
the type param making it incorrect.


dfreeman:
Thanks, @wagenet!

PR #281, opened 2022-04-06
Update readme with example of third-party types

ef4:
It took me a bit to figure out the best way to extend the Registry type from a .d.ts file for both components and helpers, so I figured this may be helpful in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ TypeScript-powered tooling for Glimmer templates.
     - [Component Signatures](#component-signatures-1)
     - [Ember Components](#ember-components)
     - [Template Registry](#template-registry)
+    - [Typing Your Dependencies](#your-dependencies)
     - [Route and Controller Templates](#route-and-controller-templates)
     - [Template-Only Components](#template-only-components)
     - [Rendering Tests](#rendering-tests)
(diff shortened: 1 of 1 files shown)


boris-petrov:
I believe a better way to type `page-title` is:

```ts
import Helper from '@glint/environment-ember-loose/ember-component/helper';

declare class PageTitleHelper extends Helper<{
  PositionalArgs: [string | undefined, ...ReadonlyArray<string | undefined>];
  Return: void;
}> {}

...

'page-title': typeof PageTitleHelper;
```

dfreeman:
In the past we've encouraged folks to declare template entities with a class declaration in the way @boris-petrov outlined here to avoid committing to any particular details about what template entities look like beyond what strictly needed to be treated as public. Earlier on in design and development of Glint, we wanted to leave the door open to the potential that we might store rules for resolving type information on the static side of the `Helper` class, for example, which would be lost with a bare constructor type `new () => Helper`.

At this point, though, I think it's safe to say that that isn't going to happen, and in fact there are concrete advantages to working with bare constructors in some situations. We've committed to that approach in the upstream helper and modifier types in the past couple of days specifically because of those advantages.

Given that, I think `new () => Helper<...>` is a reasonable thing to teach. The disparity with `ComponentLike` maybe points toward a use for `HelperLike` and `ModifierLike` types as well (particularly taking the `{{helper}}` and `{{modifier}}` helpers into account), but that's something we can look into after the current native-imports integration work is done.

Thank you, @ef4!

ef4:
@boris-petrov thanks for that example, that does read somewhat nicer so I can imagine preferring it in some situations.

PR #283, opened 2022-04-06
Support native import paths & signatures

dfreeman:
## Overview
Since its initial release, Glint has required users to use customized reexported versions of items from the `@ember/component`, `@glimmer/component` and `ember-modifier` packages in order for developers to be able to provide the granularity of type information needed to be able to typecheck those entities.

Now that https://github.com/emberjs/rfcs/pull/748 has been merged and implemented, we can instead augment those modules themselves, accounting for the signature structure that was ultimately agreed upon there.

This change allows users of `@glint/environment-ember-loose` to opt into native integration by importing `@glint/environment-ember-loose/native-integration` somewhere in their ambient types.

## Change Details
Note for reviewers: this should be much more consumable commit-by-commit than looking at the total diff.

At a high level, this PR:
 - updates our local copies of `@types/ember__component` and `ember-modifier` to get their most recent type accommodations for Glint
 - updates Glint to account for their new use of abstract constructor types
 - rehomes the existing types tests for our reexported entities to make it clear that that's what they apply to
 - adds the aforementioned `native-integration` module with appropriate declarations
 - fixes up our reexports so that their types don't clash with the newly-added native declarations
 - copies over the types tests for reexported entities and reworks them to test the native entities in an equivalent way

## Migration
As users move from the `environment-ember-loose` reexports to the native import paths, they'll also need to update their signatures. Developers can make this migration incrementally or all at once, updating to our eventual 0.8 release that drops the reexports when they're done.

For components, there are two key changes (see [the relevant section in the RFC](https://github.com/dfreeman/ember-rfcs/blob/glimmer-component-signature/text/0748-glimmer-component-signature.md#invokablecomponentsignature)):
 - `Yields` has become `Blocks`.  This key has a more complex notional desugaring, but the shorthand is compatible with how `Yields` worked before.
 - `Args` and `PositionalArgs` have been merged into `Args: { Named: ...; Positional: ... };`. If your component only has named args, you can skip the wrapping layer and continue to use `Args: MyNamedArgs` instead.

For helpers and modifiers, `NamedArgs` and `PositionalArgs` have been merged into `Args: { Named: ...; Positional: ... }` similar to the change for components. However, since neither named nor positional args are more privileged in today's implementation of helpers and modifiers, there is no shorthand.

## Still To Come
This diff was large enough already that I felt it merited a PR, but additional work is still pending for this to be ready for consumption. In particular, we need:
 - An equivalent of `environment-ember-loose`'s `ComponentLike` that accounts for the new signature structure (and doesn't require an import from `environment-ember-loose`)
 - README updates to reflect the new import paths and signature structures
 - A migration guide for folks who are moving from the old imports and signatures to new ones

At some point we also need to update `@glint/environment-glimmerx` as well, but that depends on that set of packages updating their own dependencies on `@glimmer/*` packages first.

At some point after the above has been released, we'll cut Glint 0.8.0 which will:
 - Remove the reexports entirely
 - Remove the `native-integration` opt-in import and instead always provide the integration declarations when included
 - Add `ember-modifier` and `@glimmer/component` at their minimum signature-compatible versions as `peerDependencies` to `@glint/environment-ember-loose`.


chriskrycho (commented):
Retroactive :shipit:

PR #289, opened 2022-04-07
Don't pretend the CLI can check individual files

### LINKED ISSUE #274: Passing file to CLI does not work
`glint --help` suggest you can pass the file you want to check:

```bash
> glint --help
glint [options] [file...]
```

But whatever you pass, a path to an existing (invalid!) hbs file or just `.`, it will return without any console output and with an exit code of zero!

dfreeman:
Fixes #274


chriskrycho (approved):
💙

PR #290, opened 2022-04-07
Add a migration guide for native imports

dfreeman:
This adds a guide for users to migrate from the `environment-ember-loose` reexports to the native import paths.

Note that this does not touch the baseline documentation everywhere else, which still uses the reexports. I'll update that (along with the remaining unmigrated tests) as part of 0.8.0.

### DIFF
--- a/docs/migrating.md
+++ b/docs/migrating.md
@@ -1,0 +1,301 @@
+# Migrating
+
+## Native Signatures and Imports
+
+This guide provides direction for migrating from custom Glint-specific import paths and signature formats to the
+native imports and standardized signatures for `@ember/component`, `@glimmer/component` and `ember-modifier`.
+
+Note that this guide applies to `@glint/environment-ember-loose`, but `@glint/environment-glimmerx` doesn't yet support
+native imports, pending an upgrade of GlimmerX's own dependencies.
+
+### Background
+
+Prior to version `0.7.4`, Glint required users to import the factories and base classes for components, helpers and
+modifiers from custom environment-specific paths. This was because the "native" versions of those entities couldn't
+capture enough information for Glint to typecheck them in a template with reasonable fidelity.
+
+Consider `@glimmer/component` as an example.
+
+```typescript
+import Component from '@glimmer/component';
+
+export interface MyComponentArgs {
+  message: string;
+}
+
+export default class MyComponent extends Component<MyComponentArgs> {
+  // ...
+}
+```
+
+Knowing only about the `MyComponent`'s expected `@arg` values, Glint couldn't provide any validation or assistance
+for what blocks the component accepts, what parameters it yields to those blocks, or what kind of modifiers would
+be valid to apply to it.
+
+### Signatures
+
+Ember RFC [#748](https://github.com/emberjs/rfcs/pull/748) formalized a notion of "component signatures" based on
+exploratory work Glint had done to find ways of expression the type information that `Args` alone couldn't capture.
+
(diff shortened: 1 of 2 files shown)


docs/migrating.md
```diff
@@ -0,0 +1,301 @@
...
+
+This guide provides direction for migrating from custom Glint-specific import paths and signature formats to the
+native imports and standardized signatures for `@ember/component`, `@glimmer/component` and `ember-modifier`.
+
+Note that this guide applies to `@glint/environment-ember-loose`, but `@glint/environment-glimmerx` doesn't yet support
+native imports, pending an upgrade of GlimmerX's own dependencies.
+
+### Background
+
+Prior to version `0.7.4`, Glint required users to import the factories and base classes for components, helpers and
```
chriskrycho, line 13:
```suggestion
Prior to version `0.7.4`, Glint required users to import the factories and base classes for components, helpers, and
```

OXFORD

docs/migrating.md
```diff
@@ -0,0 +1,301 @@
...
+incrementally or all at once, depending on the size of your project.
+
+#### Components
+
+For components, there are two key changes (see [the relevant section](https://github.com/dfreeman/ember-rfcs/blob/glimmer-component-signature/text/0748-glimmer-component-signature.md#invokablecomponentsignature) in the RFC):
+
+- `Yields` has become `Blocks`. This key has a more complex notional desugaring, but the shorthand is compatible with
+  how `Yields` worked before.
+- `Args` and `PositionalArgs` have been merged into `Args: { Named: ...; Positional: ... }`. If your component only has
+  named args, the wrapping layer can be skipped and you can continue to use `Args: MyNamedArgs` as before.
```
chriskrycho, line 68: Maybe adding a note here that this only particular affects Classic Component users, since it's the same for Glimmer Component users?

docs/migrating.md
```diff
@@ -0,0 +1,301 @@
...
+{% endtabs %}
+
+#### Helpers
+
+For helpers, `NamedArgs` and `PositionalArgs` have been merged into `Args: { Named: ...; Positional: ... }`, similar to
+the change for components. Unlike components, however, since neither type of argument is privileged over the other in
+the way helpers are defined today, there is no shorthand.
+
+For helpers and modifiers, NamedArgs and PositionalArgs have been merged into Args: { Named: ...; Positional: ... } similar to the change for components. However, since neither named nor positional args are more privileged in today's implementation of helpers and modifiers, there is no shorthand.
+
```
chriskrycho, line 163:
```suggestion
```

chriskrycho (changes requested):
Couple small tweaks, but 🤩

docs/migrating.md
dfreeman: BOOOOOOOO

docs/migrating.md
dfreeman: Oops! I had this notion in the back of my brain that I'd never deleted that after I was done copying from it, but then it wasn't where I looked 😂

docs/migrating.md
dfreeman: Less because I have religious convictions about the Oxford comma and more just because I know I've not used it in several other places in this document, I'm going to leave this as is for internal consistency.

chriskrycho: That seems reasonable.

Also, did we finally just discover a significant disagreement?!? 😱

chriskrycho (approved):

PR #291, opened 2022-04-07
Deprecate `environment-ember-loose` reexports

dfreeman:
This marks all of the exported types and values from `@glint/environment-ember-loose` as deprecated with pointers to alternatives.

### DIFF
--- a/packages/environment-ember-loose/-private/utilities.ts
+++ b/packages/environment-ember-loose/-private/utilities.ts
@@ -26,6 +26,8 @@ export type YieldsOf<C extends ComponentLike> = C extends Constructor<
 /**
  * The basic shape of a valid component signature. See the
  * README for further details.
+ *
+ * @deprecated Define signatures with no parent interface.
  */
 export type ComponentSignature = {
   Args?: object;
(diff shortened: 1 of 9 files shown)


dfreeman:
Ugh noooo test failures. Who would have thought the test suite that carefully checks the diagnostics emitted from using these files would object to new diagnostics suddenly showing up 🙃

kpfefferle:
Should https://typed-ember.gitbook.io/glint/using-glint/ember/imports be removed from the docs site?

dfreeman:
Yep! But it needs to happen in concert with updating the rest of the docs to use the new import paths and (slightly less find-and-replacably) signature formats 🙂 

I'm aiming to get that done by the end of the week!

kpfefferle:
@dfreeman In my experiments this week, I'm finding that I often still **DO** need to use the deprecated re-exports in order to avoid certain type errors. Hoping this is something that's just still settling so I can quickly move away from them 🤞

kpfefferle:
@dfreeman It turns out I missed adding the native opt-in before using the standardized imports:

```
import '@glint/environment-ember-loose/native-integration';
```

Now things are looking 🤩

PR #296, opened 2022-04-12
Use `SilentError` when no config is found

simonihmig:
Throwing a regular error gave an IMO confusing error message, with the full stack trace etc.

image


dfreeman:
Thank you, @simonihmig!

jamescdavis:
Released in [v0.7.5](https://github.com/typed-ember/glint/releases/tag/0.7.5)

Issue #297, opened 2022-04-14
Invalid parse after use of <template> as expression


ef4:
Given the file:

```js
import Component from '@glimmer/component';

console.log(<template></template>);

export class A extends Component {
  <template>Hello</template>
}
```

Glint's parse ends up off:

```
host/tests/integration/components/example-test.gts:6:3 - error TS1068: Unexpected token. A constructor, method, accessor, or property was expected.

6   <template>Hello</template>
    ~
```

(more errors follow that one, but they all seem to be fallout from the parser being off)

If I comment out the the `console.log` line, the whole thing parses and type checks OK.

The runtime is working fine, so I don't think this is a problem with ember-template-imports. The code it emits looks correct. This is only a type-checking problem.

chriskrycho:
Ohhhh, interesting. Thank you!

ef4:
I figured out how to extract the transformed intermediate representation, which makes the syntax error visible:

```ts
import Component from '@glimmer/component';

console.log(({} as typeof import("glint-environment-ember-template-imports/-private/dsl")).template(function(𝚪, χ: typeof import("glint-environment-ember-template-imports/-private/dsl")) {
  𝚪; χ;
}));

export class A extends Component {
  <static { ({} as typeof import("glint-environment-ember-template-imports/-private/dsl")).template(function(𝚪: import("glint-environment-ember-template-imports/-private/dsl").ResolveContext<A>, χ: typeof import("glint-environment-ember-template-imports/-private/dsl")) {
  𝚪; χ;
}) as unknown }
}
```

Looks like an off-by-one error during the handling of the first template leaves a stray `<` character before the `static {}` block of the second template.

ef4:
I understand better what is failing here. I think it's probably also causing https://github.com/typed-ember/glint/issues/263.

The bug starts with [this code](https://github.com/typed-ember/glint-environment-ember-template-imports/blob/7c60ab19252f307a3e3e592e28ef3e78e7f319e4/-private/environment/transform.ts#L196-L199
), which is doing a risky thing because the positions that it's assigning are positions in the *original* source code. But the node they're assigning them on is about to get appended to an AST representing the "preprocessed-but-not-yet-transformed" form that I pasted above.

[...]

PR #299, opened 2022-04-19
Fix preprocessed template locations

ef4:
This is a fix for #297.

It allows environment-defined transforms to emit template locations via meta instead of using `node.pos` (which is unreliable when the source has been preprocessed into an intermediate form).

### DIFF
--- a/packages/config/src/environment.ts
+++ b/packages/config/src/environment.ts
@@ -25,6 +25,10 @@ export type GlintExtensionPreprocess<T> = (
 export type GlintEmitMetadata = {
   prepend?: string;
   append?: string;
+  templateLocation?: {
+    start: number;
+    end: number;
+  };
 };
 
 export type GlintExtensionTransform<T> = (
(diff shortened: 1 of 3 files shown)


jamescdavis:
This works great! I'd like to see a test for using `templateLocation`. Maybe something like: https://github.com/ef4/glint/compare/fix-preprocessed-locations...typed-ember:fix-preprocessed-locations_with-preprocessing-test

ef4:
Thanks, I merged your test.

I did look into those tests and held off diving into them until after discussing the possibility of replacing the test suite's contrived preprocessor environment with the real glint-environment-ember-template-imports. It sounds like there is appetite for that too. 👍

jamescdavis:
> ...the possibility of replacing the test suite's contrived preprocessor environment with the real glint-environment-ember-template-imports. It sounds like there is appetite for that too.

Definitely! I just wanted a test for now so we can ship this (that will likely morph into tests for the real environment soon).

dfreeman (approved):

jamescdavis:
Released in [v0.7.5](https://github.com/typed-ember/glint/releases/tag/0.7.5)

PR #300, opened 2022-04-19
TypeScript upgrade

ef4:
This upgrades the glint monorepo's typescript version.

The reason I want to do this is because I'm preparing to move glint-environment-ember-template-imports into here, and that package relies on typescript's support for class static blocks that is not available in the present version.


ef4:
@typescript-eslint/parser prints a disclaimer about its lack of official support for this typescript version, but everything seems to lint OK, and to solve the warning we need get over the eslint 8 major version, so I think that can go in a separate PR.

dfreeman:
Thank you!

PR #301, opened 2022-04-19
Integrate and test ember-template-imports environment

ef4:
This ports https://github.com/typed-ember/glint-environment-ember-template-imports/tree/initial-implementation/-private into glint's monorepo.

The main benefit is that we can do integrated testing, and in most cases can replace the contrived custom-test environment with the real ember-template-imports environment.

Marking this as draft as I intend to port more of the tests over, but I wanted to invite early feedback.

Open questions:
 - [ ] should we rename it `@glint/environment-ember-template-imports`?


dfreeman:
>  in most cases can replace the contrived custom-test environment with the real ember-template-imports environment.

Are there cases where we can't? It would be great to drop that entirely (not necessarily in one fell swoop in this PR, but eventually) if we can.

> should we rename it @glint/environment-ember-template-imports?

I'm in favor 👍

ef4:
> Are there cases where we can't? 

Not that I know of, I just haven't gone through every test yet. If all the tests are really in support of ember-template-imports, then they can all switch and we can drop the custom-test environment.

dfreeman:
I think that should be the case—I'm happy to take a pass after this lands to clean up the custom-test environment and rework any remaining tests that are making use of it

ef4:
Yeah, my question was more whether those tests are all testing semantics that are exercised by ember-template-imports. From Dan's comment it sounds like they probably are.

jamescdavis:
I did a quick-and-dirty test refactor and all tests that currently use the `custom-test` environment work fine with `ember-template-imports`: https://github.com/ef4/glint/compare/ember-template-imports...typed-ember:ember-template-imports_remove_custom-test_environment

(cherry-pick if you like)

dfreeman:
Thank you for getting this rolling, @ef4! It's now been landed in #333 and released in `v0.8.2`

ef4:
Thanks for finishing it -- I had intended to and then had two weeks of unplanned Covid "vacation".

PR #305, opened 2022-04-20
Add class-based helper to ts-ember-app test package

jamescdavis:
I've had this lying around for a while and keep meaning to commit it.

### DIFF
--- a/test-packages/ts-ember-app/app/helpers/affix.ts
+++ b/test-packages/ts-ember-app/app/helpers/affix.ts
@@ -1,19 +1,22 @@
 import Helper from '@ember/component/helper';
 
+type Positional = [string];
+type Named = { prefix?: string, suffix?: string };
+
 export interface AffixHelperSignature {
   Args: {
-    Named: { prefix?: string, suffix?: string };
-    Positional: [string];
+    Positional: Positional;
+    Named: Named;
   },
   Return: string;
 }
 
 export default class AffixHelper extends Helper<AffixHelperSignature> {
   override compute(
-    params: AffixHelperSignature['Args']['Positional'],
-    { prefix, suffix }: AffixHelperSignature['Args']['Named']
-  ): AffixHelperSignature['Return'] {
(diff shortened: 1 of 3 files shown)


test-packages/ts-ember-app/app/helpers/affix.ts
```diff
@@ -0,0 +1,24 @@
...
+    Named: { prefix?: string, suffix?: string };
+    Positional: [string];
+  },
+  Return: string;
+}
+
+export default class AffixHelper extends Helper<AffixHelperSignature> {
+  override compute(
+    params: AffixHelperSignature['Args']['Positional'],
+    { prefix, suffix }: AffixHelperSignature['Args']['Named']
```
jamescdavis, line 14: Maybe I should break these out as separate named types?

dfreeman: I've waffled several times on which is the better default to nudge people toward. @chriskrycho do you have thoughts?

chriskrycho: Is “delete class-based-helpers’ and their annoying `compute` method with them and just teach people to use function-based-helpers with local state buckets they can set up manually if appropriate” an option? 😝 

More seriously, my own instinct here has been to just write the type as standalone aliases. Also *possibly* of interest is that TSLS will, as of recently, actually just expand the types for you to the concrete types the signature suggests (at least in VS Code).

For a teaching POV, though, I think doing it with named types is the least annoying?

```ts
type Positional = [string];
type Named = { prefix?: string, suffix?: string };

export interface AffixHelperSignature {
   Args: {
     Named: Named;
     Positional: Positional;
   },
   Return: string;
}

export default class Affix extends Helper<AffixHelperSignature> {
  override compute(
    [s]: Positional,
    { prefix, suffix }: Named
  ) {
    // ...
  }
}
```

dfreeman: Defaulting to named types seems reasonable to me 👍

dfreeman (approved):

PR #306, opened 2022-04-20
Update to use Node 14+

chriskrycho:
Since we are planning on cutting a 'major' with v0.8.0 shortly anyway, seems like a good time to upate the Node versions we are targeting. Among other things, this lets us take advantage of package `exports`, at least once we can require TS 4.7 (which we should be able to do in a 0.9 release in May).


dfreeman (approved):
Good call; thanks!

PR #307, opened 2022-04-21
Update documentation for native imports & signatures

dfreeman:
Currently all of our documentation is framed in terms of the custom re-exports we've had to use for much for Glint's lifetime, but with #283 native imports are now an option, and with #304 those imports will be the _only_ option. Accordingly, this PR updates the documentation site to use native imports and signatures in all of its example snippets.

This also removes the bulk of the contents of our README (from which the doc site contents were originally copied) in favor of nudging folks toward the more structured version of the content and avoiding the need to maintain the same information two places.


docs/ember/helper-and-modifier-signatures.md
```diff
@@ -0,0 +1,146 @@
...
+{% tabs %}
+{% tab title="Class-Based" %}
+
+```typescript
+import Helper from '@ember/component/helper';
+
+export interface AddSignature {
+  Args: {
+    Positional: Array<number>;
+    Named: { andThenMultiplyBy?: number };
```
jamescdavis, line 18: We should use type aliases for these.

docs/ember/helper-and-modifier-signatures.md
```diff
@@ -0,0 +1,146 @@
...
+
+```typescript
+import Modifier from 'ember-modifier';
+import { renderToCanvas, RenderOptions, VertexArray } from 'neat-webgl-library';
+
+export interface Render3DModelSignature {
+  Element: HTMLCanvasElement;
+  Args: {
+    Positional: [model: VertexArray];
+    Named: RenderOptions;
```
jamescdavis, line 96: Also here.

jamescdavis (dismissed):
We [decided](https://github.com/typed-ember/glint/pull/305#pullrequestreview-947396176) to teach type aliases for args in class-based helper and modifier signatures.

docs/ember/helper-and-modifier-signatures.md
dfreeman: Already on it locally 😜

docs/ember/helper-and-modifier-signatures.md
dfreeman: This is already an imported type alias

docs/ember/helper-and-modifier-signatures.md
```diff
@@ -0,0 +1,151 @@
...
+import Modifier from 'ember-modifier';
+import { renderToCanvas, VertexArray, RenderOptions } from 'neat-webgl-library';
+
+type Positional = [model: VertexArray];
+
+export interface Render3DModelSignature {
+  Element: HTMLCanvasElement;
+  Args: {
+    Positional: Positional;
+    Named: RenderOptions;
```
jamescdavis, line 101: nice

docs/ember/helper-and-modifier-signatures.md
```diff
@@ -0,0 +1,151 @@
...
+{% tabs %}
+{% tab title="Class-Based" %}
+
+```typescript
+import Helper from '@ember/component/helper';
+
+type Positional = Array<number>;
+type Named = { andThenMultiplyBy?: number };
+
+export interface AddSignature {
```
jamescdavis, line 18: Since signatures don't actually _need_ to be exported, is there any value in doing so?

dfreeman: People sometimes like to do type gymnastics with them, and reverse engineering them from the component type itself generally makes TS fall over. Given the number of times I've had to point people to just exporting the signature type instead of trying to extract it, it seemed reasonable to do it by default.

docs/ember/template-registry.md
```diff
@@ -2,12 +2,16 @@ Because Ember's template resolution occurs dynamically at runtime today, Glint n
 
 The recommended approach is to include a declaration in each component, modifier or helper module that adds it to the registry, which is the default export of `@glint/environment-ember-loose/registry`.
```
gitKrystan, line 3: This section could use an example of what the glint error message will look like if you forget to add something to the registry so that Google will point someone to this documentation when that happens.

docs/ember/contextual-components.md
```diff
@@ -20,32 +20,32 @@ Given a component like the one declared above, if you wrote this in some other c
...
-  default: [{ foo: ComponentWithBoundArgs<typeof MyComponent, 'value'> }]
+Blocks: {
+  default: [{ foo: WithBoundArgs<typeof MyComponent, 'value'> }]
 }
 ```
 
 If you had pre-bound multiple args, you could union them together with the `|` type operator, e.g. `'value' | 'count'`.
 
-Note that `ComponentWithBoundArgs` is effectively shorthand for writing out a `ComponentLike<Signature>` type, which is a generic type that any Glimmer component, Ember component or `{{component}}` return value is assignable to, assuming it has a matching signature.
+Note that `tWithBoundArgs` is effectively shorthand for writing out a `ComponentLike<Signature>` type, which is a generic type that any Glimmer component, Ember component or `{{component}}` return value is assignable to, assuming it has a matching signature.
```
gitKrystan, line 45:
```suggestion
Note that `WithBoundArgs` is effectively shorthand for writing out a `ComponentLike<Signature>` type (a generic type to which any Glimmer component, Ember component or `{{component}}` return value can be assigned, assuming it has a matching signature).
```

docs/ember/typing-your-dependencies.md
```diff
@@ -5,19 +5,34 @@ Here's an example that provides types for the `page-title` helper and the `Welco
 {% code title="types/global.d.ts" %}
```
gitKrystan, line 5: This section could use another reference to https://github.com/Gavant/glint-template-types.

gitKrystan (commented):
My review includes comments on this PR as well as comments on the state of the Glint+Ember docs in general (for you to take or leave).

docs/contents.md
```diff
@@ -7,8 +7,8 @@
 
 - Ember
```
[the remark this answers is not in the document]
jamescdavis: >Since Glint currently only works with TypeScript

Glint actually works with JS (#199), it's just not documented yet (#208).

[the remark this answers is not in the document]
gitKrystan: > Glint actually works with JS (#199), it's just not documented yet (#208).

Awesome! I do still think a link to e-cli-ts would be useful _somewhere_ since people frequently use them together. My concern is that someone might assume they only need Glint since the overview says "Glint's CLI provides a template-aware tool for performing end-to-end TypeScript typechecking"

README.md
```diff
@@ -65,557 +32,12 @@ The `glint` CLI can be used to typecheck your project in a similar manner to `ts
...
 
-You can use the `glint` executable in CI to ensure you maintain type safety in your templates. For example, you might replace an `ember-try` config's `tsc` invocation with `glint` like this:
-
-```diff
--     command: "tsc --noEmit",
-+     command: "glint",
-```
-
-You can also use the `glint` command locally with the `--watch` flag to monitor your project as you work!
+You can [use the `glint` executable in CI][using-glint] to ensure you maintain type safety in your templates. You can also use the `glint` command locally with the `--watch` flag to monitor your project as you work!
```
chriskrycho, line 35: The wording here is :thinking: – "maintain type safety in your templates" sort of reads to me as if it is *only* for templates. Trying to think of a rephrasing and so far failing, though.

docs/ember/component-signatures.md
```diff
@@ -1,18 +1,30 @@
-While Glimmer components accept `Args` as a type parameter, and Ember components accept no type parameters at all, the Glint version of each accepts `Signature`, which contains types for `Element`, `Args` and `Yields`.
+Since the implementation of [RFC 748], Glimmer and Ember components accept a `Signature` type parameter as part of their
```
chriskrycho, line 1: Do we care about the history? From a teaching POV, I think it’s more or less irrelevant now?

```suggestion
Glimmer and Ember components accept a `Signature` type parameter as part of their
```

dfreeman: My thinking in including this was that for many folks who don't necessarily follow the RFCs repo as closely as you or I do, this may be their first exposure to signatures as a concept. Nearly any existing documentation and code examples they'll see in the wild today will only show `Component<Args>`, including in many cases their own experience writing TS components in the past.

By calling out the RFC first thing on this page, my hope was to (a) reassure readers that this typing scheme is "just Ember" rather than something special Glint is inventing, and (b) provide easy access to the context behind how we got to this design from `Args` for those who may want to better understand it.

As TS is incorporated into the Ember docs themselves with guidance around signatures, and as code in the wild begins to be written this way, I agree that the history won't bring anything here from a teaching perspective. Or do you think it's already just a distraction today?

docs/ember/component-signatures.md
```diff
@@ -21,10 +33,11 @@ export interface SuperTableSignature<T> {
...
   };
-  // We accept two named blocks: an optional `header`, and a required
-  // `row`, which will be invoked with each item and its index.
-  Yields: {
-    header?: [];
+  // We accept two named blocks: a parameter-less `header` block
+  // and a `row` block which will be invoked with each item and
+  // its index sequentially.
+  Blocks: {
+    header: [];
```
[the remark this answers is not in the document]
dfreeman: This was a bit of intentional cleanup from a much earlier version of Glint's internals. Today Glint doesn't have a notion of required-ness for blocks, since `{{yield}}`-ing to a block that your consumer didn't actually pass is just a noop. https://github.com/typed-ember/glint/pull/149 has a bit more detail on that

dfreeman:
Thanks for the feedback, everyone! I REALLY appreciate the attention on the docs ❤️ 

In the interest of keeping the scope of this PR from blowing out (despite what the diff looks like, it really is 95% find/replacing code snippets and moving existing prose around between `.md` files), I'm going to merge this as it stands to accomplish the original goal of unblocking our 0.8.0 release.

I would _love_ to review PRs incorporating the other suggestions here (as well as anything else that can be improved!), but I know you're all busy and have already devoted time to writing up your commentary here, so I'll plan to come back and revisit after I travel next week if better writers than me don't get to it first 🙂

docs/ember/component-signatures.md
chriskrycho: Ah, this is a good point. I have a serious Curse of Knowledge™ going on, I think—your reasoning here makes good sense. Let's leave it in!

PR #309, opened 2022-04-21
Extend volta config in test packages

jamescdavis:
This makes working directly in test packages nicer when your default node version is < 14.


dfreeman (changes requested):
Thanks for this! I think we probably want to use Volta's `extends` here to keep things in sync with the root package version except in cases where we're intentionally diverging (which today I don't think we have any of) https://docs.volta.sh/advanced/workspaces

jamescdavis:
TIL Volta has `extends` :laughing:

chriskrycho (approved):

Issue #269, opened 2022-02-28
Make clear the path for incremental adoption


runspired:
Currently there is no ability to configure what templates might receive type checking, so applications migrating from JS to TS thinking of adopting glint see this as all-or-nothing.

dfreeman:
This is exactly the purpose of [the `include` glob(s)](https://github.com/typed-ember/glint#getting-started) in your `.glintrc` 🙂

runspired:
@dfreeman is that documented some place?

chriskrycho:
He linked directly to the docs. 😅

simonihmig:
I think an `exclude` option would also be helpful. Say you have `app/components/special/` that you want to exclude, but don't want to explicitly write dozens of `include` to include all the possible siblings of `special`, like:

```yaml
include:
- app/components/**
exclude:
- app/components/special/**
```

PR #310, opened 2022-04-23
Add @glint-nocheck directive.

jamescdavis:
This adds a `@glint-nocheck` directive, similar to [`@ts-nocheck` in .ts files](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#-ts-nocheck-in-typescript-files), that will cause Glint to not report errors for the whole template (while still providing autocompletion, types, and other diagnostics). It is basically a global `@git-ignore`. While `include` already provides a way to only check specified templates, workflows like [lint-to-the-future](https://github.com/mansona/lint-to-the-future) use file-based ignores. This provides a quick way to opt-out individual templates from type-checking that is easily discoverable by readers of the template.

### DIFF
--- a/packages/transform/src/template-to-typescript.ts
+++ b/packages/transform/src/template-to-typescript.ts
@@ -41,6 +41,14 @@ export function templateToTypescript(
   return mapTemplateContents(template, (ast, { emit, record, rangeForLine, rangeForNode }) => {
     let scope = new ScopeStack([]);
 
+    const firstNode = ast.body[0];
+    if (firstNode?.type === 'MustacheCommentStatement') {
+      let text = firstNode.value.trim();
+      if (/^@glint-nocheck/i.test(text)) {
+        return;
+      }
+    }
+
     emitTemplateBoilerplate(() => {
       for (let line of preamble) {
         emit.text(line);
--- a/packages/transform/src/template-to-typescript.ts
+++ b/packages/transform/src/template-to-typescript.ts
@@ -66,7 +66,7 @@ export function templateToTypescript(
 
         case 'CommentStatement':
         case 'MustacheCommentStatement':
-          return emitComment(node, template.length);
+          return emitComment(node);
 
         case 'MustacheStatement':
           return emitTopLevelMustacheStatement(node);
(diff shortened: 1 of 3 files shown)


packages/transform/src/template-to-typescript.ts
```diff
@@ -41,6 +41,14 @@ export function templateToTypescript(
   return mapTemplateContents(template, (ast, { emit, record, rangeForLine, rangeForNode }) => {
     let scope = new ScopeStack([]);
 
+    const firstNode = ast.body[0];
+    if (firstNode?.type === 'MustacheCommentStatement') {
+      let text = firstNode.value.trim();
+      if (/^@glint-nocheck/i.test(text)) {
+        return;
```
dfreeman, line 48: I think we may want to consider implementing this more like `@glint-ignore` and `@glint-expect-error` work, where we [capture the existence of the directive as part of transforming the module](https://github.com/typed-ember/glint/blob/main/packages/transform/src/template-to-typescript.ts#L151-L167) and then [apply it after the fact](https://github.com/typed-ember/glint/blob/main/packages/core/src/common/transform-manager.ts#L268-L273) when transforming diagnostics on the way out.

The advantage to that approach is that it still provides editor support (hover info, go-to-definition, etc) in templates with `@glint-nocheck`, whereas short-circuiting the transformation like this will opt out of having Glint do anything with the template at all.

dfreeman (commented):
Thank you for tackling this @jamescdavis! Adding this directive has been on my mind for a while now as part of having a better story than `exclude` for cases like https://github.com/typed-ember/glint/issues/269#issuecomment-1085717624

Part of that, I think, is that we _do_ want `@glint-nocheck` templates to still be transformed, we just don't want to emit errors from them.

jamescdavis:
@dfreeman that's a really good point! I've change it to basically be a global `@glint-ignore`.

packages/transform/src/template-to-typescript.ts
```diff
@@ -148,7 +148,10 @@ export function templateToTypescript(
       }
     }
 
-    function emitComment(node: AST.MustacheCommentStatement | AST.CommentStatement): void {
+    function emitComment(
+      node: AST.MustacheCommentStatement | AST.CommentStatement,
+      templateLength?: number
```
dfreeman, line 153: I think `template` itself is already in scope here?

jamescdavis: Indeed! 😆

dfreeman (approved):
Looks great; thanks again!

We should also add docs for all of the `@glint-*` directives at some point, but that was true before this PR too 😄

jamescdavis:
> We should also add docs for all of the @glint-* directives

That's the next thing I was planning to work on! 🙂

PR #313, opened 2022-04-29
Add addon documentation

simonihmig:
As discussed in https://github.com/typed-ember/glint/discussions/303.

I am not familiar with GitBook (can you preview the changes somehow?), and also not a native speaker, so please check carefully! 🧐


dfreeman:
Thank you @simonihmig! By default GitBook doesn't generate previews for pull requests from forks, but for the moment I'm just turning that security flag off, so if you push an empty commit (or amend the current one, anything to re-trigger checks) I _think_ we should get a preview now?

simonihmig:
I rebased and (force) pushed. Preview is available! Thanks!

dfreeman:
Perfect! I'm about to be mostly disconnected as I travel back to Europe this weekend, so in all likelihood I won't get the opportunity to thoroughly review this until next week. (@chriskrycho @jamescdavis if either of you has the opportunity in the meantime, though, of course feel free to merge once you're happy with the state of things)

Thank you again for all your work on this @simonihmig—I'm _deeply_ excited to see the ecosystem pieces for Glint beginning to come together!

simonihmig:
No problem, no reason to hurry this. I just want to do this as long as my memories of this addon experiment were still fresh...

> Thank you again for all your work on this @simonihmig—I'm deeply excited to see the ecosystem pieces for Glint beginning to come together!

Sure, same here! Thanks to you and the TS team. Seeing typed templates *actually* working is so exiting, almost too good to be true! 😍

docs/ember/authoring-addons.md
```diff
@@ -0,0 +1,53 @@
...
+}
+```
+
+{% endcode %}
+
+While this easy way, which only requires the app to import this single file, is sufficient for the majority of cases, it also has its limitations. Apps can replace or extend addon provided components (helpers, modifiers), by creating a component in the app itself with the same name. The addon provided types as proposed above will likely not be fully matching, so we also need to support a *flexible* way.
+
+This is given by letting the app manage the registry entries on its own, thereby being able to pick what to import from the addon directly and what to replace with its own custom types.
+
+There is one caveat though: it is generally recommended for app code to define the registry entries right in the actual file itself, so that it is nicely co-located with the signature and implementation. But when the app would import such a module from an addon, TypeScript would see that registry declaration and apply it, which is what we want to prevent, so the app can do that. So for addons it is recommended that all the registry entries are only added to the `addon/glint.ts` file as described above, and *not* to the file of the actual class!
```
dfreeman, line 36:
```suggestion
Note: in apps, we generally define the registry entries right in the actual file itself so that it is nicely co-located with the signature and implementation. If an app imported such a module from an addon, though, TypeScript would see the registry declaration there and apply it, which is exactly what we want to prevent. For addons it is therefore recommended that all the registry entries are only added to the `addon/glint.ts` file as described above, and *not* to the file containing the actual class!
```

docs/ember/using-addons.md
```diff
@@ -0,0 +1,93 @@
+Your app likely uses already a lot of addons, most if not all of them will not ship with Glint-compatible types yet. This means that for all the components, helpers or modifiers that you use, you will need to add some types so Glint can understand them.
```
dfreeman, line 1:
```suggestion
Your app likely uses already a lot of addons, and most if not all of them will not ship with Glint-compatible types yet. This means that for all the components, helpers or modifiers that you use, you will need to add some types so Glint can understand them.
```

docs/ember/using-addons.md
```diff
@@ -0,0 +1,93 @@
...
+
+{{outlet}}
+```
+
+{% endcode %}
+
+
+## Using Glint-enabled addons
+
+For addons that already ship with Glint-compatible types there is only little to do. Their README might contain some specific instructions. But if the addon follows Glint's recommendations for [authoring addons][authoring], the basic steps are as follows.
```
dfreeman, line 49:
```suggestion
For addons that already ship with Glint-compatible types there is only a little to do. Their README might contain some specific instructions, but if the addon follows Glint's recommendations for [authoring addons][authoring], the basic steps are as follows.
```

docs/ember/using-addons.md
```diff
@@ -0,0 +1,93 @@
...
+
+In case you are already using [strict mode] templates (via [first class component templates]), you are already explicitly importing all the components, helpers and modifiers in use by the template, so Glint already has all the typing information at hand, nothing more is required!
```
dfreeman, line 51:
```suggestion
If you are already using [strict mode] templates (via [first class component templates]), you are already explicitly importing all the components, helpers and modifiers in use by the template, so Glint already has all the typing information at hand, and nothing more is required!
```

docs/ember/using-addons.md
```diff
@@ -0,0 +1,93 @@
...
+For any other component, helper or modifier that you use from the addon as-is, you can extend the registry as described under [Typing your dependencies](#typing-your-dependencies) above. You don't need to write your own types though, as the addon already ships those. Just import the types from the addon itself:
+
+{% code title="types/global.d.ts" %}
+
+```typescript
+import type ResponsiveImageComponent from 'ember-responsive-image/components/responsive-image';
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    ResponsiveImage: typeof ResponsiveImageComponent;
```
dfreeman, line 81:
```suggestion
import type ResponsiveImage from 'ember-responsive-image/components/responsive-image';

declare module '@glint/environment-ember-loose/registry' {
  export default interface Registry {
    ResponsiveImage: typeof ResponsiveImage;
```

dfreeman (commented):
Thank you, @simonihmig! I've left some phrasing suggestions throughout, but overall I think this is very nice coverage of how addons fit into the Glint world.

@chriskrycho if you have the opportunity to give this a pass in the next few days, you're certainly a better crafter of words than I am, but I know you've got plenty going on so no worries if not!

docs/ember/authoring-addons.md
```diff
@@ -0,0 +1,53 @@
...
+{% code title="addon/glint.ts" %}
+
+```typescript
+import type ButtonComponent from 'awesome-addon/components/button';
+import type ModalComponent from 'awesome-addon/components/modal';
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    AwesomeButton: typeof ButtonComponent;
+    AwesomeModal: typeof ModalComponent;
```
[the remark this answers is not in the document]
simonihmig: Uh, really? I was actually developing that specific habbit, to disambiguate between different things (components, models, services etc.) that might share the same base name. But also I am not willing to die on that hill! :wink:

simonihmig:
> I've left some phrasing suggestions throughout

I applied them all. Thanks for the thorough check!

docs/ember/authoring-addons.md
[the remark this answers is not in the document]
chriskrycho: The reason it's *distinctly* not preferable for components is: when you *invoke* a component, you don't invoke it as `<AwesomeButtonComponent>`, you invoke it as `<AwesomeButton>`. Particular in a world with First-Class Component Templates and Glint-powered auto-import (…once we implement that 😉), you'll want to be able to type `<Awesome` and hit <kbd>^</kbd><kbd>␣</kbd> and have it autocomplete to `AwesomeButton` *not* to `AwesomeButtonComponent`.

Otherwise, imagine this:

```js
import AwesomeFormComponent from './awesome-form';
import AwesomeButtonComponent from './awesome-button';
import AwesomeInputComponent from './awesome-input';

<template>
  <AwesomeFormComponent as |form|>
    <AwesomeInputComponent @value={{@name}} @update={{@setName}} @label="Name" />
    <AwesomeButtonComponent @type='submit' @action={{form.submit}} />
  </AwesomeFormComponent>
</template>
```

😬 😬 😬 

This is clearly (to me at least!) very much preferable:

```js
import AwesomeForm from './awesome-form';
import AwesomeButton from './awesome-button';
import AwesomeInput from './awesome-input';

<template>
  <AwesomeForm as |form|>
    <AwesomeInput @value={{@name}} @update={{@setName}} @label="Name" />
    <AwesomeButton @type='submit' @action={{form.submit}} />
  </AwesomeForm>
</template>
```

Beyond components—it's quite reasonable to use that pattern to disambiguate models, services, controllers, etc., at least to a degree. (Routes and controllers are particularly in need of disambiguation even so, as you nearly always have matching pairs of them!) That said, I personally try to solve the overlap by giving things the best names that I can for the module they're coming from and then using modules as namespaces or *occasionally* by renaming it at the point of export. On the other hand, `Session` *might* be an object representing the session state separately from a `SessionService`, but unless it's actually ambiguous in my codebase, I would tend to name the service `Session` and give it a `state` property or some such. Then I would name the latter `State` inside that module and tak advantage of the fact that externally I can rename it on importing it if I *need* to disambiguate it:

```ts
import { type State as SessionState } from '../services/session';
```

Or, alternatively, I could do it when exporting it:

```ts
export { type State as SessionState };
```

But here I think it’s much more arguable than components, where the way we use them means it just basically *never* makes sense to have the suffix on them in my opinion!

dfreeman:
Oops, I meant to merge this last week! Thank you again, @simonihmig! 🙏

PR #315, opened 2022-05-03
Document @glint directives

jamescdavis:
Document the following directives in a new `@glint Directives` section:
- `@glint-ignore`
- `@glint-expect-error`
- `@glint-nocheck`

### DIFF
--- a/docs/contents.md
+++ b/docs/contents.md
@@ -20,5 +20,6 @@
   - [Imports](glimmerx/imports.md)
   - [Component Signatures](glimmerx/component-signatures.md)
   - [Template Components](glimmerx/template-components.md)
+- [@glint Directives](directives.md)
 - [Migrating](migrating.md)
 - [Known Limitations](known-limitations.md)
--- a/docs/directives.md
+++ b/docs/directives.md
@@ -1,0 +1,55 @@
+# @glint Directives
+
+Glint has several directives that can be used in template comments to control
+Glint's behavior. Additional comment text can follow directives to document
+their purpose.
+
+## @glint-ignore
+
+The `@glint-ignore` directive tells Glint to ignore the line that follows it.
+Glint will not report any errors encountered on the next line.
+
+Example:
+
+```hbs
+<MyComponent @expectedArg="foo" />
+
+{{! @glint-ignore: need to update signature to accept this arg }}
+<MyComponent @unexpectedArg="bar" />
+```
+
+## @glint-expect-error
+
+The `@glint-expect-error` directive operates similarly to `@glint-ignore` in
+that it will not report type errors it encounters, but it will also produce an
+error when an error is _not_ encountered. This is useful for tests where we
+expect an invocation not to type-check (e.g. due to bad arguments) and want to
+be alerted if it does.
+
+Example:
+
+```hbs
+<MyComponent @stringArg="foo" />
+
+{{! @glint-expect-error: let me know if this starts allowing numbers }}
+<MyComponent @stringArg={{123}} />
+```
+
+## @glint-nocheck
+
+The `@glint-noceck` directive will cause glint to not report errors for the
+entire template. The template is still processed by Glint such that
+auto-complete, type look-up, jump to definition, etc. are still functional,
+but any type errors will be ignored.
+
+Example:
+
+```hbs
+{{! @glint-nocheck: this whole teplate needs work }}
+
+<MyComponent @stringArg={{123}} />
+
+<AnotherComponent @badArg="foo" />
+
+{{two-arg-helper "bar"}}
+```


docs/contents.md
```diff
@@ -20,5 +20,6 @@
   - [Imports](glimmerx/imports.md)
   - [Component Signatures](glimmerx/component-signatures.md)
   - [Template Components](glimmerx/template-components.md)
+- [@glint Directives](directives.md)
```
dfreeman, line 23:
```suggestion
- [`@glint` Directives](directives.md)
```

docs/directives.md
```diff
@@ -0,0 +1,55 @@
+# @glint Directives
+
+Glint has several directives that can be used in template comments to control
+Glint's behavior. Additional comment text can follow directives to document
+their purpose.
```
dfreeman, line 5: Do you think it's worth calling out that these correspond to `@ts` directives of the same name to help with the mental model?

docs/directives.md
```diff
@@ -0,0 +1,55 @@
+# @glint Directives
```
dfreeman, line 1:
```suggestion
# `@glint` Directives
```

docs/directives.md
```diff
@@ -0,0 +1,55 @@
...
+
+Glint has several directives that can be used in template comments to control
+Glint's behavior. Additional comment text can follow directives to document
+their purpose.
+
+## @glint-ignore
```
dfreeman, line 7:
```suggestion
## `@glint-ignore`
```

docs/directives.md
```diff
@@ -0,0 +1,55 @@
...
+Example:
+
+```hbs
+<MyComponent @expectedArg="foo" />
+
+{{! @glint-ignore: need to update signature to accept this arg }}
+<MyComponent @unexpectedArg="bar" />
+```
+
+## @glint-expect-error
```
dfreeman, line 21:
```suggestion
## `@glint-expect-error`
```

docs/directives.md
```diff
@@ -0,0 +1,55 @@
...
+Example:
+
+```hbs
+<MyComponent @stringArg="foo" />
+
+{{! @glint-expect-error: let me know if this starts allowing numbers }}
+<MyComponent @stringArg={{123}} />
+```
+
+## @glint-nocheck
```
dfreeman, line 38:
```suggestion
## `@glint-nocheck`
```

docs/directives.md
```diff
@@ -0,0 +1,55 @@
...
+The `@glint-ignore` directive tells Glint to ignore the line that follows it.
+Glint will not report any errors encountered on the next line.
+
+Example:
+
+```hbs
+<MyComponent @expectedArg="foo" />
+
+{{! @glint-ignore: need to update signature to accept this arg }}
+<MyComponent @unexpectedArg="bar" />
```
dfreeman, line 18: This maybe seems like a more compelling example for `expect-error` to me (though to be fair, that's how I feel about almost every situation where you'd use one of those directives 😅)

Realistically the only case where I use `@{ts,glint}-ignore` is where I'm dealing with compatibility stuff, e.g.

```hbs
{{! @glint-ignore: this doesn't typecheck in TS 4.6 due to a bug, but we still test against that version in CI }}
<MyComponent @arg={{this.arg}} />
```

docs/directives.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+## @glint-expect-error
+
+The `@glint-expect-error` directive operates similarly to `@glint-ignore` in
+that it will not report type errors it encounters, but it will also produce an
+error when an error is _not_ encountered. This is useful for tests where we
+expect an invocation not to type-check (e.g. due to bad arguments) and want to
+be alerted if it does.
```
dfreeman, line 27: My inclination (though if you and/or @chriskrycho disagree I'm happy to defer!) would be to introduce `expect-error` first as the more general error-squashing solution

```suggestion
be alerted if it does, or if you know something currently won't typecheck due
to an issue elsewhere and want to know when it's been fixed and you can
remove the directive.
```

Then in the docs for `ignore`, we'd explicitly call out that `ignore` is more of a special-case thing that you typically wouldn't reach for.

docs/directives.md
```diff
@@ -0,0 +1,55 @@
...
+
+The `@glint-noceck` directive will cause glint to not report errors for the
+entire template. The template is still processed by Glint such that
+auto-complete, type look-up, jump to definition, etc. are still functional,
+but any type errors will be ignored.
+
+Example:
+
+```hbs
+{{! @glint-nocheck: this whole teplate needs work }}
```
dfreeman, line 48: Maybe worth calling out that `nocheck` is a useful migration tool?

```suggestion
{{! @glint-nocheck: none of this component's dependencies are typed yet }}
```

docs/directives.md
```diff
@@ -0,0 +1,55 @@
...
+```hbs
+<MyComponent @stringArg="foo" />
+
+{{! @glint-expect-error: let me know if this starts allowing numbers }}
+<MyComponent @stringArg={{123}} />
+```
+
+## @glint-nocheck
+
+The `@glint-noceck` directive will cause glint to not report errors for the
```
dfreeman, line 40:
```suggestion
The `@glint-nocheck` directive will cause Glint to not report errors for the
```

dfreeman (commented):
Thank you for writing these up! 👏

docs/directives.md
jamescdavis: Yes, I was actually adding that this morning!

docs/directives.md
chriskrycho: Yep, I agree here!

chriskrycho (commented):
I second Dan’s suggestions, but otherwise this looks great!

dfreeman:
@jamescdavis are you still planning on doing a round of revisions here at some point? I went to go link to these docs in #367 and then realized we'd never actually published it 🙂

chriskrycho:
Closing in favor of #401 – thanks for getting this up originally, @jamescdavis!

PR #318, opened 2022-05-03
Add docs for common error messages

chriskrycho:
We can expand this over time; for now, just the first couple items I know of!

### DIFF
--- a/docs/diagnosing-common-error-messages.md
+++ b/docs/diagnosing-common-error-messages.md
@@ -1,0 +1,74 @@
+Glint does its best to provide good error messages, but there are a number of cases where we are left with whatever TypeScript provides. This page documents the most common confusing type errors you might see.
+
+## Cannot assign an abstract constructor type
+
+```
+error TS2322: Type 'abstract new () => PartiallyAppliedComponent<...> is not assignable to type 'typeof <Your Component>'.
+  Cannot assign an abstract constructor type to a non-abstract constructor type.
+```
+
+This usually means that you have written a signature for a “contextual component” (or “higher-order component”) which accepts a component as an argument or yields it in one of its blocks, like this.
+
+```typescript
+import Component from '@glimmer/component';
+import type SomeOtherComponent from './some-other-component';
+
+interface MySignature {
+  Blocks: {
+    default: [typeof SomeOtherComponent];
+  };
+}
+
+export default class MyComponent extends Component<MySignature> {}
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    MySignature: typeof MySignature;
+    'my-signature': typeof MySignature;
+  }
+}
+```
+
+```handlebars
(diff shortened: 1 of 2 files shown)


docs/diagnosing-common-error-messages.md
```diff
@@ -0,0 +1,74 @@
...
+```handlebars
+{{yield (component 'my-signature' someArg=true)}}
+```
+
+In that case, you can use the `ComponentLike` or `WithBoundArgs` helpers as discussed in [Contextual Components](ember/contextual-components.md):
+
+```typescript
+import Component from '@glimmer/component';
+import type { WithBoundArgs } from './some-other-component';
+import type { ComponentLike } from '@glint/template';
```
nlfurniss, line 41:
```suggestion
import type { WithBoundArgs } from '@glint/template';
```

Your example uses `WithBoundArgs` and should import it from glint

nlfurniss (changes requested):

docs/diagnosing-common-error-messages.md
chriskrycho: Ah! Thank you. 😅

docs/diagnosing-common-error-messages.md
```diff
@@ -0,0 +1,73 @@
...
+    default: [WithBoundArgs<typeof SomeOtherComponent, 'someArg'>];
+  };
+}
+
+export default class MyComponent extends Component<MySignature> {}
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    MySignature: typeof MySignature;
+    'my-signature': typeof MySignature;
```
dfreeman, line 53:
```suggestion
    MyComponent: typeof MyComponent;
    'my-component': typeof MyComponent;
```

docs/diagnosing-common-error-messages.md
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+```
+
+```handlebars
+{{yield (component 'my-signature' someArg=true)}}
```
dfreeman, line 33:
```suggestion
{{yield (component 'my-component' someArg=true)}}
```

docs/diagnosing-common-error-messages.md
```diff
@@ -0,0 +1,73 @@
...
+    default: [typeof SomeOtherComponent];
+  };
+}
+
+export default class MyComponent extends Component<MySignature> {}
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    MySignature: typeof MySignature;
+    'my-signature': typeof MySignature;
```
dfreeman, line 27:
```suggestion
    MyComponent: typeof MyComponent;
    'my-component': typeof MyComponent;
```

dfreeman (commented):
Thank you @chriskrycho! This looks like a great starting point, and hopefully something we can use as input for including more helpful hints in the error messages we emit down the line.

PR #321, opened 2022-05-04
For Ember components, keep args in signature, use type alias

jamescdavis:
This updates the Ember component signature example in the docs to keep the args defined within the signature (like we teach for Glimmer components), and pull them out into a type alias that can be extended (`interface Greeting extends GreetingSignature['Args']` is not allowed, unfortunately). The motivation is that this form facilitates easier migration to Glimmer components because the signature is already in the form we want for Glimmer. Users would simply delete the `GreetingArgs` type alias and the `interface Greeting extends GreetingArgs {}` lines rather than needing to modify the signature by moving the arguments inside. I also think that the consistency between the Glimmer and Ember signatures is a win.

### DIFF
--- a/docs/ember/component-signatures.md
+++ b/docs/ember/component-signatures.md
@@ -77,18 +77,18 @@ Since Ember components don't have `this.args`, it takes slightly more boilerplat
 import Component from '@ember/component';
 import { computed } from '@ember/object';
 
-// We break out the args into their own interface to reference below:
-export interface GreetingArgs {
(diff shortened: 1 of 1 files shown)


chriskrycho (commented):
One thought: should we skip the intermediate type alias?

```ts
export default i interface Greeting extends GreetingSignature['Args'] {}
```

(I don't have a strong opinion!)

chriskrycho (approved):

PR #324, opened 2022-05-10
Add section for adding types to non-TS addons

jamescdavis:
This has come up recently so I thought it would be good to add a section on how to do it properly.


chriskrycho (approved):
Love it! Thank you!

PR #326, opened 2022-05-13
DOCS: Clarify Getting Started and Ember Installation pages

gitKrystan:
Clarifications to the Getting Started and Ember Installation pages per my comments on https://github.com/typed-ember/glint/pull/307. Specifically:

- Link directly to specific setup instructions in lieu of the somewhat confusing generic `.glintrc.yml` example.
- Link to ember-cli-typescript from Ember installation page since that package is so frequently used with this one. This is helpful if the user is not aware of the e-cli-ts package and got to the Glint docs by Googling "Ember TypeScript."
- Moves Glint Language Server docs to separate file to simplify the Getting Started page. The average user is not going to use the language server when they are just getting started.
- Some reformatting to make the page easier to skim.

### DIFF
--- a/docs/contents.md
+++ b/docs/contents.md
@@ -24,3 +24,4 @@
 - [Migrating](migrating.md)
 - [Diagnosing Common Error Messages](diagnosing-common-error-messages.md)
 - [Known Limitations](known-limitations.md)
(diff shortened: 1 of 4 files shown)


chriskrycho (approved):
Thank you!

PR #328, opened 2022-05-19
Fix name of environment-ember-loose main function

### LINKED ISSUE #327: Is this export from environment-ember-loose's environment misnamed?
Was poking down to see if I could figure out why `TemplateOnlyComponent` isn't working in an addon I'm working with, and came across this export from `environment-ember-loose/-private/environment`… which seems like it's probably misnamed?

https://github.com/typed-ember/glint/blob/eed2e4692d13b47bf1b040decc8abdfc018d1192/packages/environment-ember-loose/-private/environment/index.ts#L14

chriskrycho:
Fixes #327


chriskrycho:
Nightly failure seems unrelated.

PR #329, opened 2022-05-19
Make Ember template-only components ComponentLike

chriskrycho:
- Currently, template-only components do not work with `WithBoundArgs` because they are not `ComponentLike`. Introduce a failing test to demonstrate the problem (and prevent future regressions).
- Passing `args: never` instead of having an abstract constructor involved tripped up the assignability relation for `ComponentLike` with `TemplateOnlyComponent`: `new (arg: never) => Type` is not assignable to `new (...args: any) => Type` ([example][1]). Instead, introduce an abstract constructor type which the `TemplateOnlyComponent` interface can *extend*, using interface merging with the existing base definition in Ember itself (currently on DefinitelyTyped) to provide the desired semantics and constraints:
    - The type acts like a class for our purposes, meaning we can use it to get instance type info.
    - The type is *not* user-constructible (users cannot call `new` on it, since the constructor is `abstract`).

Thanks to @dfreeman for ID'ing the correct shape of the fix here!

[1]: https://tsplay.dev/WoGX8W

### DIFF
--- a/packages/environment-ember-loose/-private/dsl/integration-declarations.d.ts
+++ b/packages/environment-ember-loose/-private/dsl/integration-declarations.d.ts
@@ -41,10 +41,13 @@ interface TemplateOnlyComponentInstance<S> extends InstanceType<ComponentLike<S>
   [Context]: ComponentContext<void, S>;
 }
 
+// As with other abstract constructor types, this allows us to provide a class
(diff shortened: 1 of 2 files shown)


dfreeman (approved):
Thank you for tracking this down!

PR #330, opened 2022-05-22
docs: change return value of page-title helper

muziejus:
The code for the `page-title` helper includes:

```js
  compute(params, _hash) {
    let hash = {
      ..._hash,
      id: this.tokenId,
      title: params.join(''),
    };

    this.tokens.push(hash);
    this.tokens.scheduleTitleUpdate();
    return '';
  }
```

https://github.com/ember-cli/ember-page-title/blob/9f3a3b979a8bbca9e41f5c274993a964dc0fa461/addon/src/helpers/page-title.js#L24

The return value of the helper is an empty string, and not `void`, so
I've updated the signature typing in the glint documentation.

### DIFF
--- a/docs/ember/using-addons.md
+++ b/docs/ember/using-addons.md
@@ -18,7 +18,7 @@ declare module '@glint/environment-ember-loose/registry' {
     WelcomePage: ComponentLike;
     'page-title': HelperLike<{
       Args: { Positional: [title: string] };
-      Return: void;
+      Return: '';
     }>;
   }
 }


dfreeman:
Thank you for the PR—it's always wonderful to see improvements to the docs here! In this instance, however, I don't think we want to make this change.

While it's true that the implementation of that helper returns `''`, from a practical perspective what it's modeling is closer to [an `effect`](https://www.pzuraq.com/blog/introducing-use#effects). The [`page-title` docs](https://ember-cli.github.io/ember-page-title/) only ever direct users to invoke the helper as a top-level item in the template, not that its return value is something that should be consumed.

If a user were to write something like this:
```hbs
<MyComponent @aStringTypeArg={{page-title "hello"}} />
```
That would almost certainly be a misuse of the helper and indicate a misunderstanding on the user's part of how it works, but `Return: ''` would allow that to typecheck. Note that [`void` as a return type does not necessarily mean "returns `undefined`", but rather "returns something whose type won't be observed"](https://stackoverflow.com/questions/58885485/why-does-typescript-have-both-void-and-undefined/58885486#58885486).

muziejus:
Thanks, @dfreeman. I understand conceptually what you mean, and the stack overflow answer was illuminating, but I still get hung up over the fact that if I do:

```ts
const foo = ():void => ("")
```

I get an error from the TS checker. Yet from how I understand the Stack Overflow response, this isn't an "error" but rather a warning, or something:

> TypeScript will stop you from "accidently" returning a value, even though this wouldn't create a type system violation. This is helpful for catching bugs that appear from a refactoring

The real solution, as indicated [here](https://github.com/ember-cli/ember-page-title/issues/236#issuecomment-1138109351), seems to be just binning the return value :) So I'll aim my efforts there.

dfreeman:
TypeScript has a number of "what you've done is typesafe but probably not what you meant, so I'm flagging it" errors. The most common one is its excess property checking, which triggers when you use an object literal directly in a typed position but not when you use a variable of the same type:

```ts
declare function greet(options: { message: string }): void;

let options = { message: 'Hello', scope: 'World' };
greet(options); // ✅ typechecks

greet({ message: 'Hello', scope: 'World' }); // ❌ error TS2345
```

PR #331, opened 2022-05-22
docs: reword/update glint types for non-ts projects section

muziejus:
I think the suggestion to locate a non-TS addon's types into a `types/`
directory is a good one, so I've tried to update the documentation to
demonstrate that. Additionally, I added a note about including
`@types/ember__component` in the addon's `devDependencies`.

### DIFF
--- a/docs/ember/authoring-addons.md
+++ b/docs/ember/authoring-addons.md
@@ -39,9 +39,37 @@ A real world example of this setup can be seen in [`ember-responsive-image`][eri
 
 ## Adding Glint types to addons not written in TypeScript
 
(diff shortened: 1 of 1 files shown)


dfreeman:
Thank you, @muziejus! I have a couple of high-level observations:

**In an addon with `ember-cli-typescript`, the `types` directory is specifically where types go that _aren't_ meant to be consumed by the addon's users**[^1]. Because of that convention, it seems potentially quite confusing to a reader who's familiar with one setup to come across the other, regardless of whether they're coming from a TS addon to a non-TS one or vice versa. I wonder if an alternative like `addon-types/` might more directly call out the parallel to `addon/`?

**None of the advice here is actually specific to Glint!** All of the points you've written up here (how to use `typesVersions`, 
how to reason about needing `@types` dependencies for your declarations, and so on) apply fully to declaring types for any non-TS addon.

Given that, I'd love to see a PR adding general advice for shipping types with an addon that isn't natively written in TypeScript to the [ember-cli-typescript docs](https://docs.ember-cli-typescript.com/), maybe as a sibling to [this page](https://docs.ember-cli-typescript.com/ts/with-addons). We could then reference that page from the Glint docs instead of trying to squeeze a tutorial on library type declarations into docs that are mostly focused on higher-level topics 🙂 

[^1]: It's actually the dummy app's `types` directory, and probably should have gone in `tests/dummy/types` if we'd designed the addon layout with a bit more foresight, but at this point we're unlikely to introduce new ecosystem churn for v1 addons and instead make sure we have more sensible defaults for v2 packages going forward.

muziejus:
Thanks, @dfreeman. I was hoping to provide a one-stop tutorial, but you're right, this would work better as part of a general tutorial on adding types to non-ts addons.

PR #356, opened 2022-06-24
Migration script: `.glintrc.yml` to `glint` key in `tsconfig.json`

chriskrycho:
Introduce a new `@glint/scripts` package, with `migrate-glintrc` as the first script to be published, for migrating from `.glintrc.yml` to the new `"glint": { ... }` config in `tsconfig.json`.

<details><summary>Original notes</summary>

I think this is basically correct, but:

- [x] I have no tests, will try to add them next week.
- [x] It depends on the work @dfreeman has done to actually support this, and should therefore land *after* #354.
- [x] **Open question:** should we have a `scripts` package, or should we (probably better!) bundle this into `@glint/core` or something?

</details>


packages/scripts/src/migrate-glintrc.ts
```diff
@@ -0,0 +1,172 @@
...
+      )
+    )
+  )
+    .then(settledToResult)
+    .then(splitResults)
+    .then(dumpErrors)
+    .then(reportSuccesses);
+}
+
+main();
```
chriskrycho, line 172: We could drop this in favor of top-level `async` if we care to bump to requiring `module: Node16` in our `tsconfig.json`, but that seems… dicey to do at the top level. I'd perhaps be inclined to do it *only for this script*, since it's safe enough for the binary (we already don't support versions which don't support top-level `await`), but we definitely shouldn't for Glint itself, given the viral nature of that setting. If we do, that should probably also go with making this package be `type: "module"`.

dfreeman: Given that folks no longer import `@glint/*` packages at runtime (the closest thing is type-only imports from `@glint/template`), I would guess that migrating to ESM and assuming the availability of top-level `await` would probably be a safe thing for us to do, but I agree that doing that or not is probably something we should explore separately from this.

dfreeman:
I've landed the upstream branch, so this should be the last time you have to deal with that changing out from underneath you (sorry!)

packages/scripts/src/migrate-glintrc.ts
chriskrycho: I ended up taking a slightly different approach here, in line with your suggestion in [the comment below][comment]:

> Aside from freeing us up to adopt things like `type: 'module'` and other recent-versions-of-Node goodies earlier, it also means we can worry a little less about bringing in extra dependencies for scripting that we wouldn't necessarily want to drag in as part of `@glint/core`.

I left the root settings as they are, and provided new settings for *this package alone*. That made for a *bunch* of niceties.

packages/scripts/README.md
```diff
@@ -0,0 +1,3 @@
+# Glint scripts
+
+A handful of tools for working with Glint—e.g. for migrations.
```
chriskrycho, line 3: TODO: I need to add an explanation of the `migrate-glintrc` script, minimally how to use it.

packages/scripts/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "json5": "^2.2.1"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "workspaces": {
+    "nohoist": [
+      "yargs",
+      "@types/yargs"
+    ]
```
chriskrycho, line 50: This was how I ended up getting VS Code in particular to stop resolving to the existing (`yargs` v15) versions of `yargs` and `@types/yargs` instead of the ones here.

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
+import { readFile, writeFile } from 'fs/promises';
+import * as path from 'path';
+
+import { evaluate, patch } from 'golden-fleece';
+import yaml from 'js-yaml';
+import Result, { err, isInstance, ok, tryOrElse } from 'true-myth/result';
+import Unit from 'true-myth/unit';
+import yargs from 'yargs';
+
```
chriskrycho, line 9:
```suggestion
```

chriskrycho, line 9: Whoops, newline not needed; will push a fix.

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+const GlintRc = z.object({
+  environment: Environment,
+  checkStandaloneTemplates: z.optional(z.boolean()),
+  include: Files,
+  exclude: Files,
+});
+
+type GlintRc = z.infer<typeof GlintRc>;
+
+type GlintConfig = {
```
chriskrycho, line 33: Should probably rename this to clarify that it is `GlintTsconfig` or such.

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+}
+
+function loadOrCreateTsconfig(configPath: string): Promise<string> {
+  return readFile(configPath, { encoding: 'utf-8' }).catch((e) => {
+    console.info(
+      `Could not load tsconfig.json at ${configPath}: ${e}; attempting to create a new one`
+    );
+    return '{}';
+  });
+}
```
chriskrycho, line 63: I *think* this is the behavior we want: if there is a `.glintrc.yml` somewhere and there is *not* a corresponding `tsconfig.json`, we will create one, and the only thing it will have there is a `glint: { ... }` configuration. Does that seem right?

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+
+function saveFile(path: string, data: string): Promise<Result<Unit, string>> {
+  return writeFile(path, data)
+    .then(() => ok<Unit, string>())
```
chriskrycho, line 67: My kingdom for a language with actual inference; this really ought to be `.then(ok)`. 😑

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+  let yamlData = yaml.load(contents, {
+    onWarning: (yamlException) => errors.push(yamlException.toString()),
+  });
+  if (errors.length) {
+    return err(`Could not parse YAML:\n\t${errors.join('\n\t')}`);
+  }
+
+  return tryOrElse(
+    (e) => {
+      assert(e instanceof z.ZodError, 'error somehow not a zod error');
```
chriskrycho, line 88: Panicking here *works*, but maybe it'd be better to just turn it into an `Err` case, just with a dedicated message?

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+    return err(`Could not patch ${JSON.stringify(contents)}: not a string`);
+  }
+
+  return tryOrElse(
+    (e) => `Could not patch data:\n\t${e instanceof Error ? e.message : JSON.stringify(e)}`,
+    () => {
+      let result = evaluate(contents);
+
+      if (typeof result !== 'object' || result == null || Array.isArray(result)) {
+        throw new Error('invalid contents of tsconfig.json file');
```
chriskrycho, line 113: Should maybe think about returning the actual contents so people can see what exactly it was that failed?

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+  );
+}
+
+function patchTsconfig(contents: unknown, rc: GlintRc): Result<string, string> {
+  if (typeof contents !== 'string') {
+    return err(`Could not patch ${JSON.stringify(contents)}: not a string`);
+  }
+
+  return loadTsconfig(contents)
+    .andThen<ReadyToTransform>((loadedTsconfig) =>
```
chriskrycho, line 127: Again with the "wouldn't it be nice if inference worked consistently?" complaint.

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+            if (rc.include !== undefined) {
+              config.glint.transform.include = rc.include;
+            }
+
+            if (rc.exclude !== undefined) {
+              config.glint.transform.exclude = rc.exclude;
+            }
+          }
+
+          return patch(contents, config);
```
chriskrycho, line 163: I should probably extract the part where I build up a parent scope, and thus isolate the exception handling.

packages/scripts/src/lib/_migrate-glintrc.ts
```diff
@@ -0,0 +1,302 @@
...
+
+function settledToResult<T>(
+  settledResults: Array<PromiseSettledResult<Result<T, unknown>>>
+): Array<Result<T, string>>;
+function settledToResult<T>(
+  settledResults: Array<PromiseSettledResult<T>>
+): Array<Result<T, string>>;
+function settledToResult<T>(
+  settledResults: Array<PromiseSettledResult<T | Result<T, unknown>>>
+): Array<Result<T, string>> {
```
chriskrycho, line 177: Needs overrides because otherwise it'll happily resolve an input of `Array<PromiseSettledResult<T>>` where `T` is `Result<A, B>` to an output of `Array<Result<Result<A, B>, string>>`. 😑

tsconfig.compileroptions.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2019",
+    "target": "es2020",
```
chriskrycho, line 3: This might technically be a breaking change? If so, we can leave it as is and have the *local* compiler options bump to it.

packages/scripts/__tests__/migrate-glintrc.test.ts
```diff
@@ -0,0 +1,475 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import { fileURLToPath } from 'url';
+
+import { parse } from 'json5';
+import { afterEach, beforeEach, describe, test, expect } from 'vitest';
```
chriskrycho, line 6: Actual imports? 🤯 This is the *real* win over Jest. 😏 I can actually report, though: vitest is good.

tsconfig.compileroptions.json
dfreeman: I think we're good: https://node.green/#ES2020

It looks like ES2021 is where we start getting goodies that aren't in Node 14

packages/scripts/src/lib/_migrate-glintrc.ts
dfreeman: This seems like a panic-appropriate situation to me if we aren't expecting any other errors to be thrown here, but I'll defer to you 😄

packages/scripts/src/lib/_migrate-glintrc.ts
dfreeman: That seeeeems fine? In practice in < 0.9 I think we effectively required your `.glintrc` and `tsconfig` to be colocated with one another or things would break in weird ways (which was part of the point of this migration—acknowledging and embracing that coupling)

packages/scripts/__tests__/migrate-glintrc.test.ts
dfreeman: Neat! I'll have to look at moving the rest of our stuff over (I _suspect_ vitest won't require as many gymnastic feats of module trickery to get running in the Code testing environment)

packages/scripts/package.json
chriskrycho: Stopped being needed after correctly getting the dependency graph right.

packages/scripts/src/lib/_migrate-glintrc.ts
chriskrycho: I'll go with this and if it proves a problem we can fix it up.

packages/scripts/README.md
```diff
@@ -1,3 +1,12 @@
...
+
+## `migrate-glintrc`
+
+To migrate from `.glintrc.yml` files to using a `glint` key within `tsconfig.json` files:
+
+- install `@glint/scripts` as a dev dependency
+- run the `migrate-glintrc` script:
+    - `node ./node_modules/.bin/migrate-glintrc <paths to glintrc.yml files to migrate>`
+    - `yarn migrate-glintrc <paths to glintrc.yml files to migrate>`
```
dfreeman, line 12: An alternative that avoids needing to explicitly install the package and deal with `.bin` stuff might be:

```
npx -p @glint/scripts migrate-glintrc <path(s) to glintrc.yml files to migrate>
```

chriskrycho: Ooooh, today I learned. I went looking for that thing and could not find it anywhere. Search fail!

Issue #214, opened 2021-08-09
HTML attributes shouldn't accept element and component values


boris-petrov:
Not sure about other attributes but at least the `title` is not checked:

```hbs
<iframe src="" title={{123}}>
</iframe>
```

This should give an error for `title` but it does not.

dfreeman:
We allow setting any primitive value as an HTML attribute, since they typically have reasonable coercions (e.g. `<input value={{this.aNumber}} ...>`) or type-specific behavior (e.g. a boolean value will set or unset the attribute entirely).

If we had richer types available for the attributes that specific elements accept (e.g. `title` is always a string, `disabled` always a boolean, etc), we might consider tighter bounds, but since that type info isn't readily available and would be a huge maintenance burden to maintain by hand, we opted for the current system.

dfreeman:
> I was setting an HTMLElement (the result from a helper) and there was [Object object] in the title

That's actually something I've been meaning to fix! 🙂

Currently we have a single mechanism that encompasses emitting values as either top-level content and or in HTML attributes, but `HTMLElements` are a great example of something that should only be allowed at the top level. Block-less components are another.

Issue #128, opened 2021-04-01
Export type for invokeEmit


wagenet:
```ts
export declare function invokeEmit<
  T extends AcceptsBlocks<{}> | SafeString | string | number | boolean | null | void
>(value: T): void;
```

I'd like to be able to use this value some places:

```ts
AcceptsBlocks<{}> | SafeString | string | number | boolean | null | void
```

dfreeman:
What's the context you're looking to use this type in? Something like this?

```ts
interface MyComponentSignature {
  Args: {
    foo: SomeValueThatCanSafelyBeEmittedInAMustacheButIDontCareExactlyWhatItIs
  };
}
```

wagenet:
More or less

Issue #28, opened 2020-11-24
Document common error messages


dfreeman:
In general I think error messages should be relatively easy to map from TypeScript terms to what's happening in the template, but it's quite possible there will be situations where an error message is nonobvious if you don't know how the template is being translated to TS. JSX is _natively_ supported by the compiler and still has this problem, and even just in vanilla TypeScript authors sometimes struggle to understand what a type error is trying to tell them.

Given that we have the original template AST node available when we're emitting error messages, we may be able to eventually bake in "Hint: this may mean you forgot to do XYZ" help when certain error codes pop up on certain node types, but in the meantime we should be keeping an eye out for common situations that may not be easy to diagnose for anyone not familiar with glint internals.

chriskrycho:
@dfreeman does the one from Dec. 2020 still need docs! Since I started adding those kinds of things in #318, I figured I’d check!

dfreeman:
Here's another: this happens when you try and pass a function as an attribute value.

Technically this works today if the attribute corresponds to a prop of the same name on the DOM element, but the prop/attr confusion is something the community has talked about moving away from in the future, and the behavior of assignments like this in a world with the default helper manager RFC in place is still iffy.

```
app/components/some-template.hbs:1:17 - error TS2345: Argument of type '() => void' is not assignable to parameter of type 'EmittableValue | AcceptsBlocks<{}, any>'.
  Property '[Element]' is missing in type '() => void' but required in type 'AcceptsBlocks<{}, any>'.

1 <button onclick={{this.toggle}}>
                  ~~~~~~~~~~~~~~~

  node_modules/@glint/template/-private/integration.d.ts:41:3
    41   [Element]: El extends Element ? El : null;
         ~~~~~~~~~
    '[Element]' is declared here.

  app/components/some-template.hbs:1:17 
    1 <button onclick={{this.toggle}}>
                      ~~~~~~~~~~~~~~~
    Did you mean to call this expression?
```

dfreeman:
https://discord.com/channels/480462759797063690/484421406659182603/978754236466728960

dfreeman:
#342 brought up a key one: the disparity between the number of visible positional args and the number we report in diagnostics. Because named args are always passed as the first arg, TS sees `{{go 'foo'}}` as `go({}, 'foo')`, which means if you do something like passing a positional arg where none is expected, `{{go 'foo'}}` will say "Saw 2 args but expected 1", which is super confusing.

PR #361, opened 2022-06-28
Clearer error messages for common scenarios

dfreeman:
## Overview

Often, the type errors we expose from the translated versions of templates are confusing in the context of the template. This problem isn't unique to us—JSX type errors are notoriously obtuse, and TS _natively_ supports that syntax! That said, in many cases where we have the shape of the template syntax to guide us, we can suggest what the most likely culprit is. #28 captures some of these cases.

This PR lays out the groundwork for doing error augmentation, then attempts to tackle a few of the most egregious cases:
 - trying to invoke a non-Glint-aware entity as a component/modifier/helper (long obtuse message about overloads)
 - invoking an entity with the wrong number of positional parameters (wrong number of args reported, since named args count as one in the translation)
 - trying to assign a non-primitive (most frequently, a function) value as an HTML attribute ("not assignable to {long union of weird types}")

As we encounter other such cases in the future, we should be able to much more easily add special handling for them with this work in place.

## In Practice

Given the following `.gts` file:

```tsx
let RandomPOJO = {};

function greet(message: string, name: string) {
  return `${message}, ${name}!`;
}

<template>
  <div onclick={{RandomPOJO}}></div>

  {{RandomPOJO}}

  <RandomPOJO />

  {{greet "Hello"}}
</template>
```

Running the CLI now gives these errors (abbreviated):
```
src/Test.gts:8:8 - error TS2322: Only primitive values (see `AttrValue` in `@glint/template`) are assignable as HTML attributes. If you want to set an event listener, consider using the `{{on}}` modifier instead.
src/Test.gts:10:3 - error TS2345: Only primitive values and certain DOM objects (see `ContentValue` in `@glint/template`) are usable as top-level template content.
src/Test.gts:12:4 - error TS2769: The given value is not resolvable as an entity that can be invoked in a template, such as a component, modifier or helper.
src/Test.gts:14:3 - error TS2554: Expected 2 arguments, but got 1.
```

<details>
<summary>Full CLI output</summary>

```
$ /Users/dfreeman/code/glint/node_modules/.bin/glint
src/Test.gts:8:8 - error TS2322: Only primitive values (see `AttrValue` in `@glint/template`) are assignable as HTML attributes. If you want to set an event listener, consider using the `{{on}}` modifier instead.
  Type '{}' is not assignable to type 'AttrValue'.

8   <div onclick={{RandomPOJO}}></div>
         ~~~~~~~

src/Test.gts:10:3 - error TS2345: Only primitive values and certain DOM objects (see `ContentValue` in `@glint/template`) are usable as top-level template content.
  Argument of type '{}' is not assignable to parameter of type 'ContentValue'.

10   {{RandomPOJO}}
     ~~~~~~~~~~~~~~

src/Test.gts:12:4 - error TS2769: The given value is not resolvable as an entity that can be invoked in a template, such as a component, modifier or helper.
  No overload matches this call.
    The last overload gave the following error.
      Argument of type '{}' is not assignable to parameter of type '(...positional: unknown[]) => unknown'.
        Type '{}' provides no match for the signature '(...positional: unknown[]): unknown'.

12   <RandomPOJO />
      ~~~~~~~~~~

  ../../packages/environment-ember-template-imports/-private/dsl/index.d.ts:77:25
    77 export declare function resolve<P extends unknown[], T>(
                               ~~~~~~~
    The last overload is declared here.

src/Test.gts:14:3 - error TS2554: Expected 2 arguments, but got 1.

14   {{greet "Hello"}}
     ~~~~~~~~~~~~~~~~~
```

</details>

Or, in image form in-editor: 
An editor pane showing an error message about non-primitive values being unassignable as HTML attributes.

## Other Details

In support of this change, we now have exported `AttrValue` and `ContentValue` types in `@glint/template`. These types are used in error messages, as well as being potentially useful for general-purpose API design by consumers (see #128).

They also fix a [long-standing problem we've had](https://github.com/typed-ember/glint/issues/214) of not distinguishing between content that's valid as an HTML attribute (`AttrValue`) vs those that are valid as top-level "append content" (`ContentValue`).


packages/transform/src/diagnostics/augmentation.ts
```diff
@@ -0,0 +1,146 @@
...
+function checkResolveError(
+  diagnostic: Diagnostic,
+  mapping: MappingTree
+): ts.DiagnosticMessageChain | undefined {
+  let nodeType = mapping.sourceNode.type;
+  if (nodeType === 'ElementNode' || nodeType === 'PathExpression' || nodeType === 'Identifier') {
+    return addGlintDetails(
+      diagnostic,
+      'The given value is not resolvable as an entity that can be invoked in a template, ' +
+        'such as a component, modifier or helper.'
```
chriskrycho, line 117: I don't *yet* have a great suggestion for it, but I wonder if we can wordsmith this a bit to be a bit clearer for people who don't have our level of context. 🤔 "resolvable as an entity that can be invoked" *feels* deep in the weeds. Hmmmmm.

dfreeman: I had a similar feeling (and meant to call out in the PR description that phrasing suggestions on all these messages are welcome!) but wasn't immediately able to come up with anything clearer that wasn't somehow misleading.

dfreeman: Maybe just `The given value is not resolvable as a component, modifier or helper` is better?

PR #377, opened 2022-08-04
Add support for project references and `--build`

chriskrycho:
This PR adds basic support for using Glint with TypeScript's [Project References][ts-p-r]. This makes Glint work *much* better in large projects, by taking advantage of TypeScript's ability to do incremental builds, which take advantage of build work which has already happened.

Three qualifications about the support added in this PR:

1. It *only* introduces support for `glint --build`, *not* for any of the other options TypeScript supports in conjunction with `--build`:
    - `--dry`
    - `--force`
    - `--clean`
    - `--watch`

2. It does not introduce direct support for the `--incremental` flag. TypeScript uses this flag to support doing the same kind of local build caching in non-`composite` projects as in `composite` projects.

3. It does not include test coverage for whether *invalidation* of existing projects work. That is, it may have bugs and not rebuild when it *should* when running `glint --build`.

However, it will be much faster and easier to resolve each of those with this foundation in place. I will address them with much smaller follow-on PRs.

[ts-p-r]: http://www.typescriptlang.org/docs/handbook/project-references.html


docs/configuration/project-references.md
```diff
@@ -0,0 +1,147 @@
+Glint works out of the box with TypeScript’s [Project References][pr] feature:
+
+- Instead of running just `glint`, you run `glint --build`, and Glint will correctly do incremental builds for composite projects.
+- In the future, the Glint Language Server also takes advantage of composite projects to improve performance when working in just one part of a large codebase.
```
dfreeman, line 4: I would guess this already works as expected? To my knowledge, vanilla `tsserver` never triggers a build in a composite project itself, it just implicitly takes advantage of `.d.ts` files that have been generated by `--build` to avoid having to typecheck the other projects in the composite graph.

docs/configuration/project-references.md
```diff
@@ -0,0 +1,147 @@
...
+- extend the root options config
+- include the `types` directory as well as their own files
+
+```jsonc
+{
+  "extends": "../../tsconfig.shared.json",
+  "compilerOptions": {
+    "composite": true
+  },
+  "include": ["../../local-types/**/*.ts", "./src/**/*.ts"]
```
dfreeman, line 95: It's been a long time, but I _think_ the original reason we had to do `**/*.ts` in these was something to do with watcher issues when we were using `tsc` to do compilation in `ember-cli-typescript` (way back in 1.x). Given that, this could probably be just `["../../local-types", "src"]`.

docs/contents.md
```diff
@@ -5,7 +5,8 @@
 
 ## Using Glint
 
-- [Configuration](configuration.md)
+- [Configuration](./configuration/_index_.md)
```
dfreeman, line 8: I'm not sure this matches the file name?

packages/core/__tests__/cli/build-watch.test.ts
```diff
@@ -0,0 +1,19 @@
+// import { stripIndent } from 'common-tags';
+// import stripAnsi from 'strip-ansi';
+// import os from 'os';
+import Project from '../utils/project';
+import { describe, beforeEach, afterEach, test } from 'vitest';
+
+describe('CLI: watched build mode typechecking', () => {
+  let project!: Project;
+  beforeEach(async () => {
+    jest.setTimeout(20_000);
```
dfreeman, line 10: I think I'd expect this to 💥

packages/core/__tests__/utils/project.ts
```diff
@@ -54,9 +85,9 @@ export default class Project {
         skipLibCheck: true,
         allowJs: true,
         checkJs: false,
-        ...compilerOptions,
+        ...(config.compilerOptions ?? {}),
```
dfreeman, line 88: [nit] I think `null`/`undefined` spread fine on their own?

packages/core/src/cli/index.ts
```diff
@@ -22,6 +25,35 @@ const argv = yargs
...
+  .option('dry', {
+    implies: 'build',
+    description: `Show what would be built (or deleted, if specified with '--clean'). Same as the TS \`--dry\` flag.`,
+    type: 'boolean',
+  })
+  .option('incremental', {
+    description:
+      'Save .tsbuildinfo files to allow for incremental compilation of projects. Same as the TS `--incremental` flag.',
+    type: 'boolean',
+  })
```
dfreeman, line 56: Should we drop these flags (or maybe comment them out) until we support them?

packages/core/src/cli/perform-check.ts
```diff
@@ -1,8 +1,10 @@
-import type ts from 'typescript';
+import ts from 'typescript';
```
dfreeman, line 1:
```suggestion
import type ts from 'typescript';
```

packages/core/src/cli/utils/patch-program.ts
```diff
@@ -0,0 +1,45 @@
...
+  // a manager for the project in question, if it exists at all, to use in
+  // getting transform errors (both type errors and syntax errors).
+  let manager: TransformManager | null;
+  if (isPool(transformManagerOrPool)) {
+    let configFile = program.getCompilerOptions()['configFilePath'];
+    assert(typeof configFile === 'string', 'internal error: missing TS config file');
+    manager = transformManagerOrPool.managerFor(configFile);
+  } else {
+    manager = transformManagerOrPool;
+  }
```
dfreeman, line 27: I know these 8 lines were the result of a lot of blood, sweat and tears, but I'm pretty happy with how this all came together!

dfreeman (commented):
🎉

packages/core/__tests__/utils/project.ts
chriskrycho: Oh, neat. (In a way, it's neat. 😂)

packages/core/src/cli/index.ts
chriskrycho: Commenting them out till supported seems like the right move, will do.

packages/core/__tests__/cli/build-watch.test.ts
chriskrycho: Fixed via 1920bdd. Thank you!

docs/configuration/project-references.md
chriskrycho: Ah, sweet. Noted. I'll update accordingly!

packages/core/__tests__/utils/project.ts
dfreeman: IIRC the justification was that `Object.assign` was already specced in ES2015 to accept falsy values, and there was desire use the same semantics for `{...foo}` instead of allowing subtle differences.

(Which means _instead_ we get the less-subtle difference that `{...null}` is fine but `[...null]` blows up 🙃)

chriskrycho: de0079f

packages/core/src/cli/perform-check.ts
chriskrycho: 4bac7479c (also updated to use `TS` instead of `ts` here)

docs/contents.md
chriskrycho: You are correct; it absolutely did not! d140f11

docs/configuration/project-references.md
chriskrycho: Updated in 725e32e!

docs/configuration/project-references.md
chriskrycho: 28c7e8b, also added a note about WIP on the other flags.

packages/core/src/cli/index.ts
chriskrycho: 4d9db80

dfreeman (approved):

PR #382, opened 2022-08-08
core/cli: implement `glint --build --force`

chriskrycho:
Building on top of the work in https://github.com/typed-ember/glint/pull/377, add correct handling for the `--force` flag. This simply passes it through directly to TypeScript itself.

(Draft because it builds on #381)


dfreeman:
🎉

PR #384, opened 2022-08-11
Remove GlimmerX shims

chadhietala:
This removes the need for the GlimmerX shims and uses them directly via declaration merging. This PR also updates all the documentation to remove references to the old shims.


chadhietala:
The super weird this is the renaming the `Helper` type to something other than `Helper` the double generic issue. There has to be some ambient type or bug somewhere. The last time `Helper` from `@glimmerx/helper` was in [0.2.0](https://github.com/glimmerjs/glimmer-experimental/blame/dcd15cadbec99a3adf016bcc373a58c06ff32d71/packages/@glimmerx/helper/src/helper.ts#L8).

packages/environment-glimmerx/-private/dsl/integration-declarations.d.ts
```diff
@@ -1,9 +1,122 @@
...
+    d: D,
+    e: E,
+    g: G
+  ): (...rest: Args) => Ret;
+}>;
+
+declare module '@glimmerx/helper' {
+  export interface Helper extends InstanceType<HelperLike> {}
+  export interface FnHelper extends _FnHelper {}
+}
```
dfreeman, line 97: I think the `Helper<S, S>` issue you're seeing is due to TS losing all sense of reality when you try to do declaration merging on a `type` declaration. Rather than the `type Helper<S>`, I think we need to target the `interface HelperInstance<S>`.

Locally this gets `helper.test.ts` passing for me:
```suggestion
declare module '@glimmerx/helper' {
  export interface FnHelper extends _FnHelper {}
}

declare module '@glimmerx/helper/dist/commonjs/src/helper' {
  export interface HelperInstance<S> extends InstanceType<HelperLike<S>> {}
}
```

chadhietala: This definitely did the trick

package.json
```diff
@@ -29,5 +29,8 @@
     "release-it-yarn-workspaces": "^2.0.0",
     "typescript": "^4.7.4"
   },
+  "resolutions": {
+    "ember-cli-htmlbars": "^6.0.1"
+  },
```
chriskrycho, line 34: @dfreeman Without this, TS can end up mis-resolving it to the v5.x present transitively in the dependency graph when doing the `integration-declarations.d.ts` module merge for `ember-cli-htmlbars`. (That would seem to be a bug in TS' resolutions, given that we *explicitly* ask for v6!) I discovered this after beating my head against the wall for over an hour trying to figure out why all the *other* declaration merges in the integration declarations were working, but that one was not—and in fact was simply not being included at all. In the end, `traceResolution` told me that it *was* being merged… with the wrong version of `ember-cli-htmlbars`'s types. 🙃

I have some concerns about using `resolutions` to fix this long-term: I *think* this will ensure that all of our own packages will work correctly this way, but I want to get it cleaned up *ASAP* by updating the things we use which transitively depend on the "wrong" version. I'm also somewhat horrified that TS can mis-resolve it that way, and if I can come up with a minimal reproduction of it, I will file a bug on TS.

chriskrycho (commented):
Besides the `resolutions` issue, I also saw that the environment import wasn't working (that's what actually *exposed* the issue which I solved with `resolutions`), because the `types/<app folder>` had the wrong name. I suspect both of these issues already existed and were simply being obscured (somehow?) by the stuff we were doing with the `@glimmerx` re-exports?

package.json
dfreeman: I could also imagine this being a "Yarn is imprecise about monorepo management" issue if v5 happened to be the one to get hoisted and v6 was only present for the specific app(s) that requested it, since packages from the monorepo will always see the top-level version of a dependency unless they themselves have requested something else.

This might be a good reason to consider shifting to npm or pnpm here (both of which are stricter about getting `peerDependencies` right), though I do wish Volta and Dependabot had proper support for the latter.

dfreeman (approved):

package.json
chriskrycho: Thankfully pnpm support is in flight; I would be very happy to switch as soon as it lands.

chriskrycho: Err: Volta had pnpm support in flight. Dependabot is not getting a ton of attention since being acquired AFAICT; we should reevaluate Renovate if that ends up being a blocker.

PR #386, opened 2022-08-18
core/cli: add `glint --incremental` support

chriskrycho:
Include the flag in the CLI and incorporate it for both `--build` and regular builds.

The test here is intentionally minimal: we exercise a much broader set of features via the `--build` and especially `--build --dry` tests, which exercise these capabilities explicitly. Here, we just check that `--incremental` does the right thing when *not* in a `--build` context.

### DIFF
--- a/packages/core/src/cli/index.ts
+++ b/packages/core/src/cli/index.ts
@@ -49,14 +49,11 @@ const argv = yargs
     description: `Show what would be built (or deleted, if specified with '--clean'). Same as the TS \`--dry\` flag.`,
     type: 'boolean',
   })
-  // TODO: implement these!
-  /*
   .option('incremental', {
     description:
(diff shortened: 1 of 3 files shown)


packages/core/src/cli/index.ts
```diff
@@ -49,14 +49,11 @@ const argv = yargs
     description: `Show what would be built (or deleted, if specified with '--clean'). Same as the TS \`--dry\` flag.`,
     type: 'boolean',
   })
-  // TODO: implement these!
```
dfreeman, line 52: 🎉

dfreeman (approved):

packages/core/src/cli/index.ts
chriskrycho: Just need to do `--watch` now. And spoilers: that's just a matter of exercising the functionality thoroughly with the rest suite but it's working as expected with every test I convert, no changes required!

Issue #389, opened 2022-08-23
Generic component signatures not working with template tag / gts


gitKrystan:
```ts
import Component from '@glimmer/component';

export interface MyComponentSignature<T> {
  Args: {
    items: T[];
  };
}

export default class MyComponent<T> extends Component<MyComponentSignature<T>> {
  /*
    Argument of type '<T>(𝚪: import("@glint/environment-ember-template-imports/-private/dsl").ResolveContext<MyComponent<T>>, χ: typeof import("@glint/environment-ember-template-imports/-private/dsl")) => void' is not assignable to parameter of type '(𝚪: TemplateContext<void, EmptyObject, EmptyObject, void>, χ: never) => void'.
    Types of parameters '𝚪' and '𝚪' are incompatible.
      Type 'TemplateContext<void, EmptyObject, EmptyObject, void>' is not assignable to type 'ComponentContext<MyComponent<unknown>, MyComponentSignature<unknown>>'.
        Types of property 'this' are incompatible.
          Type 'void' is not assignable to type 'MyComponent<unknown>'.glint:ts(2345)
   */
  <template></template>
}
```

### Workaround for anyone running into this issue

I've been just doing `@ts-expect-error` above the `<template>` tag and type-checking within the tag still appears to work as expected.

dfreeman:
Oops, I forgot my closing keyword. Fixed in #390 and released in v0.9.2!

PR #390, opened 2022-08-24
Break `template()` in two

dfreeman:
## Background

Historically we've used a single `template()` function to act as the wrapper around our translated template bodies, deciding based on context whether to explicitly apply a type annotation to its `𝚪` context parameter or allow its type to be inferred.

This has generally worked for us, but it's resulted in:
 - some additional complexity during transformation to propagate type parameters and their constraints from the backing class down to the function body representing the template
 - needing to maintain a `ResolveContext` type for internal (i.e. template backing context) resolution that works completely differently from how external (i.e. invocation signature) resolution works via `resolve()`
 - special casing templates with a backing class to always emit `as unknown` (or the JSDoc equivalent) to avoid leaking complex internal type details in declarations
 - a hard requirement for users that any class associated with a template have a name
 - issues like #389.

## This Change

This change breaks `template()` into two separate functions: `templateExpression()` and `templateForBackingValue()`. The former represents the use of a tagged string or `<template>` in an general expression position, while the latter represents their use embedded in a class body (or when tied to a default export from a companion `.hbs` file).

By making this distinction, we can address each of the bullets mentioned above, since we can specialize the signature for each version of `template()` for the scenarios they matter in instead of needing to accommodate both together.

## Before and After

Concretely, given a file like this:

```tsx
const Greeting: TOC<GreetingSig> = <template>Hello</template>;

class MyComponent<T extends string> extends Component {
  <template>
    <Greeting />, World!
  </template>
}
```

Previously we would have transformed it into the equivalent of:
```ts
const Greeting: TOC<GreetingSig> = template(function(𝚪) {
  // type of 𝚪 is inferred from `TOC<GreetingSig>` in context
});

class MyComponent<T extends string> extends Component {
  static {
    template(function<T extends string>(𝚪: ResolveContext<MyComponent<T>>) {
      // type of 𝚪 is explicit based on the `ResolveContext` utility type applied
      // to the instance type of the wrapping class
    });
  }
}
```

With this change, we instead emit the equivalent of:
```ts
const Greeting: TOC<GreetingSig> = templateExpression(function(𝚪) {
  // type of 𝚪 is inferred from `TOC<GreetingSig>` in context, as above
});

class MyComponent<T extends string> extends Component {
  static {
    templateForBackingValue(this, function(𝚪) {
      // type of 𝚪 is inferred based on `this` being passed as the first arg
    });
  }
}
```

Note the code in the examples above intentionally omits things like signature definitions that aren't relevant to the specific points under discussion.

### DIFF
--- a/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
@@ -1,8 +1,7 @@
 import Component from '@ember/component';
 import {
-  template,
+  templateForBackingValue,
   resolve,
-  ResolveContext,
   yieldToBlock,
   emitComponent,
 } from '@glint/environment-ember-loose/-private/dsl';
@@ -12,9 +11,11 @@ import type { ComponentLike } from '@glint/template';
 
 {
   class NoArgsComponent extends Component {
-    static template = template(function* (𝚪: ResolveContext<NoArgsComponent>) {
-      𝚪;
-    });
+    static {
+      templateForBackingValue(this, function (𝚪) {
+        𝚪;
+      });
+    }
   }
 
   resolve(NoArgsComponent)({
@@ -44,11 +45,13 @@ import type { ComponentLike } from '@glint/template';
   class StatefulComponent extends Component {
     private foo = 'hello';
 
-    static template = template(function* (𝚪: ResolveContext<StatefulComponent>) {
-      expectTypeOf(𝚪.this.foo).toEqualTypeOf<string>();
-      expectTypeOf(𝚪.this).toEqualTypeOf<StatefulComponent>();
-      expectTypeOf(𝚪.args).toEqualTypeOf<EmptyObject>();
-    });
+    static {
+      templateForBackingValue(this, function* (𝚪) {
+        expectTypeOf(𝚪.this.foo).toEqualTypeOf<string>();
+        expectTypeOf(𝚪.this).toEqualTypeOf<StatefulComponent>();
+        expectTypeOf(𝚪.args).toEqualTypeOf<EmptyObject>();
+      });
+    }
   }
 
   emitComponent(resolve(StatefulComponent)({}));
@@ -69,18 +72,27 @@ import type { ComponentLike } from '@glint/template';
 
   interface YieldingComponent<T> extends YieldingComponentArgs<T> {}
   class YieldingComponent<T> extends Component<YieldingComponentSignature<T>> {
-    static template = template(function* <T>(𝚪: ResolveContext<YieldingComponent<T>>) {
-      expectTypeOf(𝚪.this).toEqualTypeOf<YieldingComponent<T>>();
-      expectTypeOf(𝚪.args).toEqualTypeOf<{ values: T[] }>();
+    static {
+      templateForBackingValue(this, function* (𝚪) {
+        // We can't directly assert on the type of e.g. `@values` here, as we don't
+        // have a name for it in scope. However, the yields below confirm that the
+        // only thing we can legally yield to the default block is an element of the
+        // `@values` arg, since the only information we have about that type is that
+        // the array element and the yielded value are the same.
+        yieldToBlock(
+          𝚪,
+          'default',
+          // @ts-expect-error: only a `T` is a valid yield
+          123
+        );
 
-      expectTypeOf(𝚪.this.values).toEqualTypeOf<Array<T>>();
-
-      if (𝚪.args.values.length) {
-        yieldToBlock(𝚪, 'default', 𝚪.args.values[0]);
-      } else {
-        yieldToBlock(𝚪, 'else');
-      }
-    });
+        if (𝚪.args.values.length) {
+          yieldToBlock(𝚪, 'default', 𝚪.args.values[0]);
+        } else {
+          yieldToBlock(𝚪, 'else');
+        }
+      });
+    }
   }
 
   // @ts-expect-error: missing required arg
--- a/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
@@ -75,8 +75,8 @@ import type { ComponentLike } from '@glint/template';
     static {
       templateForBackingValue(this, function* (𝚪) {
         // We can't directly assert on the type of e.g. `@values` here, as we don't
-        // have a name for it in scope. However, the yields below confirm that the
-        // only thing we can legally yield to the default block is an element of the
+        // have a name for it in scope: the type `T` is present on the class instance,
+        // but not in a `static` block. However, the yields below confirm that the
         // `@values` arg, since the only information we have about that type is that
         // the array element and the yielded value are the same.
         yieldToBlock(
--- a/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
@@ -67,8 +67,8 @@ import { ComponentLike } from '@glint/template';
     static {
       templateForBackingValue(this, function* (𝚪) {
         // We can't directly assert on the type of e.g. `@values` here, as we don't
-        // have a name for it in scope. However, the yields below confirm that the
-        // only thing we can legally yield to the default block is an element of the
+        // have a name for it in scope: the type `T` is present on the class instance,
+        // but not in a `static` block. However, the yields below confirm that the
         // `@values` arg, since the only information we have about that type is that
         // the array element and the yielded value are the same.
         yieldToBlock(
--- a/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
@@ -1,8 +1,7 @@
 import Component from '@glimmer/component';
 import {
-  template,
+  templateForBackingValue,
   resolve,
-  ResolveContext,
   yieldToBlock,
   emitComponent,
 } from '@glint/environment-ember-loose/-private/dsl';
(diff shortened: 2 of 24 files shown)


packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
```diff
@@ -63,17 +64,27 @@ import { ComponentLike } from '@glint/template';
...
-        yieldToBlock(𝚪, 'else');
-      }
-    });
+    static {
+      templateForBackingValue(this, function* (𝚪) {
+        // We can't directly assert on the type of e.g. `@values` here, as we don't
+        // have a name for it in scope. However, the yields below confirm that the
+        // only thing we can legally yield to the default block is an element of the
+        // `@values` arg, since the only information we have about that type is that
+        // the array element and the yielded value are the same.
```
chriskrycho, line 73: I may be missing something obvious, but couldn't we minimally check something like this?

```ts
expectTypeOf(Γ.args).toEqualTypeOf<{ values: Array<T> }>();
```

I can't see how `Γ.args.values` works below otherwise! Again, maybe I’m missing something, though?

packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
```diff
@@ -69,18 +72,27 @@ import type { ComponentLike } from '@glint/template';
...
-        yieldToBlock(𝚪, 'default', 𝚪.args.values[0]);
-      } else {
-        yieldToBlock(𝚪, 'else');
-      }
-    });
+    static {
+      templateForBackingValue(this, function* (𝚪) {
+        // We can't directly assert on the type of e.g. `@values` here, as we don't
+        // have a name for it in scope. However, the yields below confirm that the
+        // only thing we can legally yield to the default block is an element of the
```
chriskrycho, line 79: This confused me till I checked it out and messed around locally, so a small suggested tweak to the language in this comment:

```suggestion
        // have a name for it in scope: the type `T` is present on the class instance,
        // but not in a `static` block. However, the yields below confirm that the
```

packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
```diff
@@ -63,17 +64,27 @@ import { ComponentLike } from '@glint/template';
...
-        yieldToBlock(𝚪, 'default', 𝚪.args.values[0]);
-      } else {
-        yieldToBlock(𝚪, 'else');
-      }
-    });
+    static {
+      templateForBackingValue(this, function* (𝚪) {
+        // We can't directly assert on the type of e.g. `@values` here, as we don't
+        // have a name for it in scope. However, the yields below confirm that the
+        // only thing we can legally yield to the default block is an element of the
```
chriskrycho, line 71: And the same here:

```suggestion
        // have a name for it in scope: the type `T` is present on the class instance,
        // but not in a `static` block. However, the yields below confirm that the
```

packages/environment-glimmerx/__tests__/component.test.ts
```diff
@@ -65,9 +65,18 @@ import { AcceptsBlocks, EmptyObject } from '@glint/template/-private/integration
   }
 
   class YieldingComponent<T> extends Component<YieldingComponentSignature<T>> {
-    static template = template(function* <T>(𝚪: ResolveContext<YieldingComponent<T>>) {
-      expectTypeOf(𝚪.this).toEqualTypeOf<YieldingComponent<T>>();
-      expectTypeOf(𝚪.args).toEqualTypeOf<{ values: T[] }>();
+    static template = templateForBackingValue(this, function (𝚪) {
+      // We can't directly assert on the type of e.g. `@values` here, as we don't
+      // have a name for it in scope. However, the yields below confirm that the
+      // only thing we can legally yield to the default block is an element of the
```
chriskrycho, line 71:
```suggestion
        // have a name for it in scope: the type `T` is present on the class instance,
        // but not in a `static` block. However, the yields below confirm that the
```

packages/template/-private/dsl/types.d.ts
```diff
@@ -19,13 +19,3 @@ export type ElementForTagName<Name extends string> = Name extends keyof HTMLElem
...
-
-/**
- * Given the constructor or instance type of a component backing class, produces the appropriate
- * `TemplateContext` type for its template.
- */
-export type ResolveContext<T> = T extends HasContext<infer Context>
-  ? Context
-  : T extends Constructor<HasContext<infer Context>>
-  ? Context
-  : unknown;
```
chriskrycho, line 31: 🔥

packages/template/-private/dsl/emit.d.ts
```diff
@@ -64,16 +64,38 @@ export declare function emitComponent<T extends AcceptsBlocks<any, any>>(
...
+/*
+ * Wraps a template body that's backed by a known value (typically a class), either
+ * via a `.hbs` association to a default export or via embedding e.g. with `<template>`.
+ *
+ * The given callback accepts a template context value as well as an instance of the
+ * environment's DSL export.
+ *
+ * Note that the constructor-based signature returns an (arbitrary) abstract constructor
+ * type in order to trigger TypeScript's higher-order function type inference and propagate
+ * any type parameters on the class (see https://github.com/microsoft/TypeScript/pull/30215)
```
chriskrycho, line 88: I don't doubt this makes perfect sense, but I have a few questions such that it doesn't actually make sense *to me* right now:

1. What is the `void` return value for (vs. the constructor version)?

2. Similarly, what is the version which *doesn't* have `Args` passed into it doing? Are these `Args` a `Signature['Args']` or some other `Args`? I *think* the latter, but the prominence of `Signature['Args']` makes me doubt that; in any case I don't know how they relate!

Net… what is each overload here supplying? 😬

packages/transform/src/template/inlining/index.ts
```diff
@@ -10,49 +10,21 @@ export type CorrelatedSpansResult = {
...
-  return { contextType, typeParams, className, inClass };
-}
-
-function findContainingClass(ts: TSLib, node: ts.Node): ts.ClassLikeDeclaration | null {
+export function isEmbeddedInClass(ts: TSLib, node: ts.Node): boolean {
   let current: ts.Node | null = node;
   do {
+    // TODO: this should likely actually filter on whether the template appears in a
+    // static block or property definition, but just "am I in a class body" is the
+    // current status quo and has been ok so far.
```
chriskrycho, line 23: What would be the difference between those two from *our* POV? (Obviously it matters from a runtime and performance POV, but it's not obvious to me why we would choose to care about it.)

packages/vscode/__tests__/helpers/async.ts
```diff
@@ -4,7 +4,7 @@ export function sleep(ms: number): Promise<void> {
 
 export async function waitUntil(callback: () => unknown): Promise<void> {
   let start = Date.now();
-  while (Date.now() - start < 5_000) {
+  while (Date.now() - start < 15_000) {
```
chriskrycho, line 7: 🤣 👏🏼

chriskrycho (changes requested):
This is lovely. I had some questions, but everything I'm looking for here is filed under *clarification of the docs and comments* rather than anything about the implementation! 🎉

packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
dfreeman: There's no `T` we can reference, because type parameters on classes are only in scope within their instance properties and methods, not static ones. That's what I was aiming to get at with `we don't have a name for it in scope`, but I'm happy to adjust if you can suggest better phrasing.

chriskrycho: I *was* missing something, see comment below. DOH. (I meant to remove this.)

packages/template/-private/dsl/emit.d.ts
dfreeman: > What is the void return value for (vs. the constructor version)?
>
> Similarly, what is the version which doesn't have Args passed into it doing?

Similar to the constructor return type, the `Args` aren't supplying anything; it's just necessary to have a type of that shape/position in order to trigger higher-order function type inference. Without that, TS won't infer a function type with type parameters for `body`, which is the root cause of the error showing up in #389.

This playground demonstrates the issue: https://tsplay.dev/w1pAyW

> Net… what is each overload here supplying? 😬

The overloads here are a 1:1 mapping from the original `ResolveContext` type:
```ts
export type ResolveContext<T> = T extends HasContext<infer Context>
  ? Context
  : T extends Constructor<HasContext<infer Context>>
  ? Context
  : unknown;
```

One case handles receiving a constructor (which is the common case), and the other handles non-constructor values (which `TemplateOnlyComponent` was originally an example of, though we might have made that a constructor today; I don't remember off the top of my head).

packages/transform/src/template/inlining/index.ts
dfreeman: If I wrote something like this today:

```tsx
class MyComponent {
  InternalThing = <template>Hello!</template>;

  <template>
    <this.InternalThing />
  </template>
}
```

Then the template-only component we're assigning to `InternalThing` would be treated as a `templateForBackingClass`, which is incorrect.

dfreeman: It's the status quo and no one has run into it yet to my knowledge, which is why I didn't want to go down the rabbithole of actually fixing it as part of this and further delay releasing, but I _did_ want to leave a note that the behavior's incorrect

chriskrycho: Ahhhh, that example is clarifying. Thank you.

packages/template/-private/dsl/emit.d.ts
dfreeman: > I wonder if we still need to keep non-constructor value support here, though, given that TemplateOnlyComponents now by definition have to go through the other path?

That's a good callout. I left it because I was uncertain and just directly translating things over, but it's easy enough to drop the second overload and see if anything breaks. I'll try that and report back.

packages/environment-glimmerx/__tests__/component.test.ts
dfreeman: The extra indentation here infuriated CI 😂😭

packages/template/-private/dsl/emit.d.ts
dfreeman: Looks like it works fine for us. Thanks for pointing it out!

chriskrycho: Sweet!

chriskrycho (approved):

PR #397, opened 2022-08-31
feat: expose component helper for glimmerx

chadhietala:
We actually use this to create rehydratable regions of a page


dfreeman:
Oh good to know `component` works now! In early GlimmerX releases it was completely broken.

Can you add tests to ensure this works as expected?

dfreeman:
Thank you! I'll aim to get a release out with this later today

PR #399, opened 2022-09-01
feat: Expose Globals registry to glimmerx

chadhietala:
While GlimmerX is a module-centric world, applications can still have
macros that are defined by the surrounding build system. This was
partially there with `config.additionalGlobals` but we actually
need to expose a registry that can be merged into so that the
globals aren't `any` types


chriskrycho (approved):
Seems reasonable to me but I'll defer to @dfreeman before merging it!

dfreeman (approved):
Looks good to me.

My internet has gone out, so I won't be able to do a release. @chriskrycho will you have the chance to do that today?

chriskrycho:
Should be able to! Merging it now, will try to release in a bit.

PR #402, opened 2022-09-01
core/cli: fix empty `.buildinfo` files with `--incremental`

chriskrycho:
@dfreeman encountered a scenario where the `.buildinfo` file was generated but empty. The first commit here is a first pass to verify that it does get generated at least sometimes. Unfortunately, it does not reproduce the bug; but it *does* tell us that this at least *generally* works.

### DIFF
--- a/packages/core/__tests__/cli/incremental.test.ts
+++ b/packages/core/__tests__/cli/incremental.test.ts
@@ -1,4 +1,4 @@
-import { existsSync, statSync } from 'fs';
+import { existsSync, statSync, readFileSync } from 'fs';
 
 import { stripIndent } from 'common-tags';
 import { beforeEach, describe, expect, test } from 'vitest';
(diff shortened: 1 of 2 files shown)


dfreeman:
Closing the loop on our conversation: the file wasn't literally empty, but it only contained a version marker without any actual build info:

```json
{"version":"4.8.2"}
```

I'll poke at it some next week if I get the opportunity

chriskrycho:
Ah, yes. We need a slightly more robust test than this; I originally used a snapshot and that had the full contents, so there is something else different between the scenario our tests hit and the real-world failure!

chriskrycho:
Update here: I identified that the problem is we were *testing* `--build` but what was broken was a normal 'check' pass. The new tests exercise it correctly. Unfortunately, I haven't been able to figure out yet what we *do* need to do to actually make it work. I’m pushing a WIP commit that shows one path which *doesn’t* work, and because it doesn’t I am mildly stumped.

chriskrycho:
My notes from the commit message in the latest:

> I would expect this to work: it simply wires up the compiler the *exact* same way we do for `--build`, but with `--incremental`, but... no. It's possible (perhaps even likely) that we need to do some extra wiring; knowing for sure will likely require reading the TS source code. As it does for these things.

package.json
```diff
@@ -15,7 +15,7 @@
     "release-it": "echo \"Running release-it via yarn breaks publishing! Use npx or a Volta global installation.\""
   },
   "volta": {
-    "node": "14.19.1",
+    "node": "16.17.1",
```
chriskrycho, line 18: Ignore this; I'll pull it out of this and check it in on a separate PR and just make sure CI still tests—I spent another 15 minutes yesterday trying to figure out whyyyyyy I couldn't get anything to work with Vitest before remembering that it was the "bump to Node 16 so ESBuild works natively" dance.

PR #403, opened 2022-09-02
bugfix: Actually include globals.d.ts

chadhietala:
@glint/environment-glimmerx@0.9.3 is broken because we forgot to add
`globals/index.d.ts` to `files` in package.json.


chriskrycho (approved):
🤣

PR #412, opened 2022-09-15
add helper and modifier helpers

### LINKED ISSUE #410: support helper and modifier helpers
using `{{#let (helper ...)}...`  or `{{#let (modifier ...)}}` does not work

patricklx:
fixes #410 

the documentation for them is not published yet. It should be there when 4.7 is published:
https://github.com/emberjs/ember.js/commit/d75a0cb13cb323165e38c79d1de177a1f747d88d

### DIFF
--- a/packages/environment-ember-loose/-private/dsl/globals.d.ts
+++ b/packages/environment-ember-loose/-private/dsl/globals.d.ts
@@ -7,10 +7,12 @@ import { EachKeyword } from '../intrinsics/each';
 import { EachInKeyword } from '../intrinsics/each-in';
 import { FnHelper } from '../intrinsics/fn';
 import { GetHelper } from '../intrinsics/get';
+import { HelperHelper } from '../intrinsics/helper';
 import { InputComponent } from '../intrinsics/input';
 import { LinkToKeyword, LinkToComponent } from '../intrinsics/link-to';
 import { LogHelper } from '../intrinsics/log';
 import { MountKeyword } from '../intrinsics/mount';
+import { ModifierHelper } from '../intrinsics/modifier';
 import { MutKeyword } from '../intrinsics/mut';
 import { OnModifier } from '../intrinsics/on';
 import { OutletKeyword } from '../intrinsics/outlet';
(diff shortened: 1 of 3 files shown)


chriskrycho:
Thank you! One of us will review this week!

chriskrycho:
@patricklx thanks again for opening this; mind fixing up the CI failures? I'll try to review tomorrow or Friday!

dfreeman:
Thank you for opening this, @patricklx!

In addition to the things flagged in CI, I think these are going to need to be a bit more complex to capture what `{{helper}}` and `{{modifier}}` can do, which isn't just looking up global helpers/modifiers, but also [pre-binding arguments](https://api.emberjs.com/ember/4.7/classes/Ember.Templates.helpers/methods/helper?anchor=helper) to either a global `{{helper "foo" 123 arg="hi"}}` or to actual helper values `{{helper this.someHelper 123 arg="hi"}}`.

You can see an example of how this works with the `{{component}}` helper, which does the same thing: https://github.com/typed-ember/glint/blob/main/packages/environment-ember-loose/-private/intrinsics/component.d.ts

Is that something you've got the time/interest to take on? If so, great, and let us know if you need any additional guidance. If not, no worries—Chris or I should hopefully be able to get to it in the near future.

boris-petrov:
I just noticed that these two:

```hbs
<div {{(if true (modifier "on" "click" this.func))}}></div>
<div {{(if true (modifier "on")) "click" this.func}}></div>
```

Actually mean the same thing so Glint should somehow support them both. No idea how. :smile:

dfreeman:
Yep! Those two examples should fall out naturally and "just work" assuming we implement the currying semantics of those two helpers I mentioned above 🙂

SergeAstapov:
Hi @patricklx! Would you be able to pick up the suggestion from @dfreeman how to overcome missing parts here?

chriskrycho:
Actually, I believe @dfreeman is working on getting these done. [From Discord](https://discord.com/channels/480462759797063690/786312479620726804/1094015678022959254):

> I’m working on adding those (and hopefully slightly improving inference characteristics for  `{{component}}`) for what should be the final 1.0 beta release

🎉

PR #416, opened 2022-09-29
Remove unhelpful `glint-language-server` docs page

dfreeman:
Git history clearly disagrees with me, but I feel like this is the third time I've deleted this page only to discover that it's still there. It's a remnant from when the docs were organized differently, and this remaining scrap of content about the language server isn't particularly useful to anyone.


dfreeman:
Fixing nightly is next on my list

PR #419, opened 2022-10-04
Add docs for JS usage, including an example

chriskrycho:
This integrates the work done by @erinsinger93 in #208 into the updated docs site we landed later, and adds some extra content.

Closes #208

### DIFF
--- a/docs/with-js.md
+++ b/docs/with-js.md
@@ -1,0 +1,173 @@
+If you have not yet migrated your project to TypeScript or you are in the process of migrating your project, the Glint CLI and language server will still function properly.
+
+## Setup
+
+All the Glint setup steps are identical.
+
+{% hint style="info" %}
+
+We do mean *identical*: unlike most existing JavaScript projects, you will need to include TypeScript in your dev dependencies.
+
+{% endhint %}
+
+Thus, installation for a JS setup looks like this:
+
+{% tabs %}
+{% tab title="Yarn" %}
+
+```sh
+yarn add --dev typescript @glint/core @glint/environment-ember-loose
+```
+
+{% endtab %}
+{% tab title="npm" %}
+
+```sh
+npm install -D typescript @glint/core @glint/environment-ember-loose
+```
+
+{% endtab %}
+{% endtabs %}
+
+{% code title="tsconfig.json" %}
+
+```javascript
+{
+  "compilerOptions": { /* ... */ },
+  "glint": {
+    "environment": "ember-loose"
+  }
+}
+```
+
+{% endcode %}
+
+## Providing JS "types"
+
+To receive equivalent rich editor support for component JS files in your project, you will need to document your components with valid JSDoc. For example:
+
+```js
+// SimpleComponent.js
+
+import Component from '@glint/environment-glimmerx/component';
+import { hbs } from '@glimmerx/component';
+import { helper } from '@glint/environment-glimmerx/helper';
+
+const or = helper(
+  /** @param {[a: *, b: *]} param */
+  ([a, b]) => a || b
+);
+
+/**
+ * @typedef SimpleComponentSignature
+ * @property {object} Args
+ * @property {string} Args.message
+ */
+
+/** @extends {Component<SimpleComponentSignature>} */
+export default class SimpleComponent extends Component {
+  static template = hbs`
+    <h1>This is my simple message: {{or @message 'hello'}}</h1>
+  `;
+}
+```
+
+### Signatures
+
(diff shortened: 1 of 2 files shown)


wagenet (approved):

docs/with-js.md
```diff
@@ -0,0 +1,173 @@
...
+{% tab title="npm" %}
+
+```sh
+npm install -D typescript @glint/core @glint/environment-ember-loose
+```
+
+{% endtab %}
+{% endtabs %}
+
+{% code title="tsconfig.json" %}
```
dfreeman, line 32: One note: I think this either needs to be `jsconfig.json` _or_ you need to explicitly set `allowJs: true` in your `compilerOptions`.

docs/with-js.md
```diff
@@ -0,0 +1,173 @@
...
+
+```js
+// SimpleComponent.js
+
+import Component from '@glint/environment-glimmerx/component';
+import { hbs } from '@glimmerx/component';
+import { helper } from '@glint/environment-glimmerx/helper';
+
+const or = helper(
+  /** @param {[a: *, b: *]} param */
```
dfreeman, line 57: If we want to go all-out with JSDoc...

```suggestion
  /**
    * @template T
    * @template U
    * @param {[a: T, b: U]} param
    * @returns T | U
    */
```

docs/with-js.md
```diff
@@ -0,0 +1,173 @@
...
+The component also provides a named block, `label`, which does not yield any value, allowing customization of the `<label>` for the `<input>`:
+
+```js
+/**
+ * @typedef FancyInputBlocks
+ * @property {[]} label
+ */
+```
+
+Finally, the component signature itself assembles those and also specifies the `Element`:
```
dfreeman, line 150: Are we missing an actual `@typedef` for `FancyComponentSignature` or am I just glossing over it?

docs/with-js.md
```diff
@@ -0,0 +1,173 @@
...
+export default class FancyInput extends Component {
+  // implementation...
+}
+```
+
+Now this will provide useful completion, go-to-definition, and refactoring.
+
+### Caveats
+
+Undocumented component JS files will behave exactly as regular undocumented JS behaves: no information will be provided for editor support apart from what can be derived from the context.
```
dfreeman, line 168: You definitely miss out on a lot by not writing doc comments, but I don't think we should undersell what value does exist here.

Type info, basic refactorings, and go-to-definition will work for anything hanging off of `this`, and inference based on things that _do_ have types (builtins, third-party dependencies with Glint types, etc) will work as a baseline without any extra typing.

dfreeman (approved):
🎉

docs/with-js.md
chriskrycho: Doh, thank you.

```suggestion
{% code title="jsconfig.json" %}
```

docs/with-js.md
chriskrycho: Good point. I'll add some text to that effect (this was just what was there in the original PR).

docs/with-js.md
chriskrycho: Nope, you’re correct: I forgot to copy it down here!

docs/with-js.md
chriskrycho:
```suggestion
Undocumented component JS files will behave exactly as regular undocumented JS behaves: no information will be provided for editor support apart from what can be derived from the context.

That caveat notwithstanding, Glint still offers quite a bit even without doc comments. Type info, basic refactorings, and go-to-definition will work for anything referring to `this`, and inference based on things that do have types (built-in items Glint knows about, third-party dependencies with Glint types, etc) will work as a baseline without any extra typing.
```

docs/with-js.md
```diff
@@ -0,0 +1,173 @@
...
+{
+  "compilerOptions": { /* ... */ },
+  "glint": {
+    "environment": "ember-loose"
+  }
+}
+```
+
+{% endcode %}
+
```
chriskrycho, line 44:
```suggestion

{% hint style="info" %}

You can also use a `tsconfig.json` file with `compilerOptions.allowJs: true`.

{% endhint %}

```

docs/with-js.md
```diff
@@ -0,0 +1,173 @@
...
+```js
+/**
+ * @typedef FancyInputBlocks
+ * @property {[]} label
+ */
+```
+
+Finally, the component signature itself assembles those and also specifies the `Element`:
+
+```js
```
chriskrycho, line 152:
```suggestion
```js
/**
  * @typedef FancyInputSignature
  * @property {FancyInputArgs} Args
  * @property {FancyInputElement} Element
  * @property {FancyInputBlocks} Blocks
  */
```

PR #427, opened 2022-10-12
Add `auto-glint-nocheck` script

### LINKED ISSUE #269: Make clear the path for incremental adoption

dfreeman:
## Background

Today, the primary strategy folks tend to use when adopting Glint in an existing codebase is to set up either an `include` or `exclude` list under `glint.transform` in their `tsconfig.json`. While this does allow for incremental adoption, it has a few downsides:
 - it's not obvious to someone reading the code whether a given template is typechecked or not
 - it's easy for the list (especially `include`) to fall out of sync as the codebase evolves
 - opting out of transformation disables _all_ Glint support for the files in question, including things like hover info and go-to-definition that can function in many cases even with type errors

The introduction of the `@glint-nocheck` directive made it possible to opt templates out of typechecking on an individual basis while still allowing Glint to analyze them, but adding these directives by hand is tedious and not something most groups have anecdotally bothered with.

## This Change

This change adds a new executable to the `@glint/scripts` package: `auto-glint-nocheck`.

Similar in spirit to `lint-to-the-future`, this script launches the Glint language server in your project, locates any templates that have type errors in them, and adds a `@glint-nocheck` directive to them.

For multi-line templates, this directive goes on the first line and includes an explanatory comment (default: `{{! @glint-nocheck: not typesafe yet }}`), while in single-line embedded templates, it's simply inserted with no comment directly before the template content.

## Notes

A few other related changes are landing as part of this:
 - switching off `esModuleInterop` for the whole codebase for easier consumption from true ESM packages (which there's a good chance we'll migrate to more completely someday)
 - pulling out an unpublished `@glint/test-utils` package to host our testing `Project` abstraction for scaffolding test projects to test Glint against
 - adding a basic explicitly-unstable programmatic interface to `@glint/core` (this is currently completely undocumented and may stay that way, but at a minimum it's now _there_)

I'm also provisionally adding `@glint/scripts` to the tests we run under Windows since, like `@glint/core` and the Code extension, it's directly interacting with the outside world and subject to OS quirks.

Closes #269.


packages/config/src/config.ts
```diff
@@ -1,4 +1,4 @@
-import path from 'path';
+import * as path from 'node:path';
```
chriskrycho, line 1: SO FANCY

packages/config/src/environment.ts
```diff
@@ -1,8 +1,8 @@
-import resolve from 'resolve';
-import path from 'path';
-import escapeStringRegexp from 'escape-string-regexp';
-import type ts from 'typescript';
-import SilentError from 'silent-error';
+import * as path from 'node:path';
+import resolve = require('resolve');
+import escapeStringRegexp = require('escape-string-regexp');
```
chriskrycho, line 3: …so *not* fancy. 😭 I hate `= require()` so much. It’s definitely the right move here, of course, but 😭 😭 😭

packages/core/src/common/transform-manager.ts
```diff
@@ -249,6 +249,15 @@ export default class TransformManager {
     return fileStat.mtime > companionStat.mtime ? fileStat.mtime : companionStat.mtime;
   };
 
+  /** @internal `TransformInfo` is an unstable internal type */
+  public findTransformInfoForOriginalFile(originalFileName: string): TransformInfo | null {
```
chriskrycho, line 253: I believe we actually want this:

```ts
friend findTransformInfoForOriginalFile(originalFileName: string): TransformInfo | null
```

:trollface:

packages/core/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  glintConfig: GlintConfig;
+  transformManager: TransformManager;
+  languageServer: GlintLanguageServer;
+  shutdown: () => void;
+}
+
+/**
+ * This function is available to consumers as an unstable API. We will not go
+ * out of our way to change or break it, but there may be breaking changes
+ * to its behavior or type signature outside of major version bumps.
```
chriskrycho, line 17: This is the (TypeScript) way.

packages/core/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * See the `auto-glint-nocheck` implementation in `@glint/scripts` for a
+ * sample use of this API.
+ *
+ * @internal
+ */
+export function analyzeProject(projectDirectory: string = process.cwd()): ProjectAnalysis {
+  let glintConfig = loadConfig(projectDirectory);
+  let documents = new DocumentCache(glintConfig);
+  let transformManager = new TransformManager(glintConfig, documents);
```
chriskrycho, line 27: A question, mostly for my own edification: is there any reason we would need this to use the `TransformManagerPool` approach, or is the idea that the things this would be used for would always have only a single manager corresponding to the single language server, regardless of the project structure?

If we *would* want to support a flow that needed the pool, it's also totally fine to add later, too; but I don't have a strong intuition about whether that would even make sense—I think *no*?

packages/scripts/src/lib/_auto-nocheck.ts
```diff
@@ -0,0 +1,160 @@
...
+          describe: 'The explanation to be included in @glint-nocheck comments',
+        })
+    )
+    .wrap(100)
+    .strict()
+    .parseSync();
+}
+
+type GlintCore = typeof import('@glint/core') &
+  typeof import('@glint/core/lib/language-server/util/index.js');
```
chriskrycho, line 70: Is this to avoid a direct dependency on it, or what? 🤔 Mind adding a comment so future-us remembers?

packages/scripts/src/lib/_auto-nocheck.ts
```diff
@@ -0,0 +1,160 @@
...
+    let [core, util] = await Promise.all([
+      import(findImport('@glint/core', cwd)),
+      import(findImport('@glint/core/lib/language-server/util/index.js', cwd)),
+    ]);
+
+    return { ...core, ...util };
+  } catch {
+    throw new Error(
+      `This script requires @glint/core >= 0.9.6 to be available in your project in order to run.`
+    );
```
chriskrycho, line 89: Does putting it in `peerDependencies` make sense, or *no* because it is apt to be run as a standalone script rather than actually installed in a project?

packages/scripts/src/migrate-glintrc.ts
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
```
chriskrycho, line 1: Huzzah. No idea what I hit here before. 🤷🏼‍♂️

packages/core/src/index.ts
dfreeman: For the sole use case we currently have, I think saying "run `auto-glint-nocheck` for each individual project as you turn on Glint for it" makes sense, but I could definitely imagine other programmatic use cases that might want something like `analyzeCompositeProject` that would provide a more holistic view of a composite setup.

My inclination is to hold off until we have a concrete use case for that in mind, as I'm _guessing_ the shape we'd want to return from that might be slightly different.

packages/scripts/src/lib/_auto-nocheck.ts
dfreeman: Big-picture, this whole chunk is to make sure we're using the project-local version of `@glint/core`.

These two lines specifically are because I find `type FooMod = typeof import('foo')` a more intuitive way of capturing "the type of this entire module" than `import * as foo from 'foo'; type FooMod = typeof foo'`.

I'll add a note.

packages/scripts/src/lib/_auto-nocheck.ts
dfreeman: I think no, for exactly that reason. I've had to open enough PRs over the years to remove codemods and the like from package dependencies that if I had my way it would be illegal to install these things as a persistent dependency in a project 😂

dfreeman:
...did Prettier break some of the `build`/`build-watch` snapshots? 🤨

chriskrycho:
It looks like 😬

packages/core/src/index.ts
chriskrycho: Makes sense to me!

packages/scripts/src/lib/_auto-nocheck.ts
chriskrycho: Ah, that's fair and makes sense 👍🏼

PR #430, opened 2022-10-19
Fix JS-backed types

wagenet:
Previously, there were some cases where `as` could be used in a generated .js file. TS isn't happy about this.

### DIFF
--- a/packages/transform/src/template/inlining/companion-file.ts
+++ b/packages/transform/src/template/inlining/companion-file.ts
@@ -49,7 +49,9 @@ export function calculateCompanionTemplateSpans(
     let backingValue: string | undefined;
     if (targetNode) {
       let moduleName = path.basename(script.filename, path.extname(script.filename));
-      backingValue = `({} as unknown as typeof import('./${moduleName}').default)`;
+      backingValue = useJsDoc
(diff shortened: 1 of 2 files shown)


packages/transform/src/template/inlining/companion-file.ts
```diff
@@ -49,7 +49,9 @@ export function calculateCompanionTemplateSpans(
     let backingValue: string | undefined;
     if (targetNode) {
       let moduleName = path.basename(script.filename, path.extname(script.filename));
-      backingValue = `({} as unknown as typeof import('./${moduleName}').default)`;
+      backingValue = useJsDoc
+        ? `(/** @type {typeof import('./${moduleName}').default} */ {})`
```
dfreeman, line 53: I think we need extra parens here (a quirk of TS's rules for applying JSDoc `@type` declarations to expressions):

```suggestion
        ? `(/** @type {typeof import('./${moduleName}').default} */ ({}))`
```

dfreeman (changes requested):
Great find—thank you for tracking this down!

packages/transform/src/template/inlining/companion-file.ts
wagenet: Fixed!

dfreeman (approved):

PR #432, opened 2022-10-20
Improve custom diagnostic messages

### LINKED ISSUE #373: Invalid `{{component}}` first arg leads to a confusing message
In `ember-loose`, where `{{component}}` accepts a string as its first value, typoing the name of your component gives an error like:

```
TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'abstract new (...args: unknown[]) => Invokable<(args: unknown) => AcceptsBlocks<any, any>>'
```

While this is technically true, it's super confusing. At a bare minimum if we see a string literal as the first arg, we should probably suggest that you check for typos and that you haven't forgotten to add a registry entry.

### LINKED ISSUE #375: Incorrect diagnostic augmentation for `{{yield}}` type errors
We see a TS2345 on a top-level mustache and automatically assume its an attempt to emit invalid content, but `{{yield}}` is a special form that we should special-case.

```
Only primitive values and certain DOM objects (see `ContentValue` in `@glint/template`) are usable as top-level template content.
  Argument of type '[]' is not assignable to parameter of type 'NonNullable<FlattenBlockParams<ExpandSignature<T>["Blocks"]>["default"]>'.glint:ts(2345)
```

### LINKED ISSUE #398: Improve error message when a helper/modifier/component is missing from the loose registry
The current error is:

```
app/components/workflow-tracker/index.hbs:3:17 - error TS7053: Element implicitly has an 'any' type because expression of type '"toggle"' can't be used to index type 'Globals'.
  Property 'toggle' does not exist on type 'Globals'.

3   {{on 'click' (toggle 'showing' this)}}
                  ~~~~~~
```

@chriskrycho tells me "that error actually means it’s not in your registry. We should note it as an error to rewrite."

dfreeman:
This PR updates some of our custom diagnostic augmentations:
 - taking care of a false-positive detection for no-arg `{{yield}}` (fixes #375)
 - suggesting a missing registry entry for a failed global lookup (fixes #398)
 - clarifying the message when the first arg to `{{component}}` is bad (fixes #373)
 - explicitly calling out that directly invoking a component via the `{{component}}` helper isn't possible with Glint and suggesting a workaround (also captured in #373)

### DIFF
--- a/packages/core/__tests__/language-server/diagnostic-augmentation.test.ts
+++ b/packages/core/__tests__/language-server/diagnostic-augmentation.test.ts
@@ -635,4 +635,95 @@ describe('Language Server: Diagnostic Augmentation', () => {
       ]
     `);
   });
+
+  test('failed `component` name lookup', () => {
+    project.setGlintConfig({ environment: ['ember-loose'] });
+    project.write({
+      'index.ts': stripIndent`
+        import Component from '@glimmer/component';
+
+        export default class MyComponent extends Component {
+          componentName = 'bar' as const';
+        }
+      `,
+      'index.hbs': stripIndent`
+        {{#let 'baz' as |baz|}}
+          {{#let
+            (component 'foo') 
+            (component this.componentName)
+            (component baz)
+            as |Foo Bar|
+          }}
+            {{! @glint-ignore: we don't care about errors here}}
+            <Foo /><Bar /><Baz />
+          {{/let}}
+        {{/let}}
+      `,
+    });
+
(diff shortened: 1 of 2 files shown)


dfreeman:
CI is running but... seems to have dissociated itself from the PR? 🤔 

https://github.com/typed-ember/glint/actions/runs/3291643073

(Also we seemingly need to upgrade most or all of our actions, based on those notes. Another day, another fiddly infra change!)

PR #437, opened 2022-10-24
Note that generic template-only-component signatures are not supported

boris-petrov:
cc @dfreeman


dfreeman (approved):
Thank you, @boris-petrov!

PR #440, opened 2022-10-26
Handle out-of-scope files more cleanly in the language server

### LINKED ISSUE #380: Don't crash on unrecognized file types

dfreeman:
Currently if you open a file we _might_ be able to provide diagnostics for but in fact isn't something the language server is aware of, you can get an error like this:

```
Glint encountered an error computing diagnostics for this file. This is likely a bug in Glint; please file an issue, including any code and/or steps to follow to reproduce the error.

Error: Could not find source file: 'app/components/my-component.js'.
```

This might happen if you open a `.js` file with `allowJs: false`, or a `.gts` file in a project without the template-imports environment. Depending on the order files are opened in, it can also happen if we accidentally synthesize a `.ts` backing file for a `.hbs` one that actually has a `.js` module somewhere.

This change ensures we're a bit more careful in those situations and do a better job determining:
 - whether we should be trying to analyze a file at all
 - whether a `.hbs` file has a backing module other than the default one we would create for it

Fixes #380

### DIFF
--- a/packages/core/__tests__/language-server/hover.test.ts
+++ b/packages/core/__tests__/language-server/hover.test.ts
@@ -172,4 +172,70 @@ describe('Language Server: Hover', () => {
       },
     });
   });
+
+  describe('JS in a TS project', () => {
+    test('with allowJs: true', () => {
+      let tsconfig = JSON.parse(project.read('tsconfig.json'));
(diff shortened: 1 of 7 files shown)


chriskrycho (commented):
Just caught up on this via GH notifications backlog. 👍🏼

Issue #436, opened 2022-10-21
Vague autocomplete weirdness


dfreeman:
I haven't dived into this carefully enough to even give a precise title, but in poking at autocomplete behavior, there are still some inconsistencies that weren't totally fixed by https://github.com/typed-ember/glint/pull/393

For example, if I have in scope a variable `trigger: { buttonId: string }` and write something like this in my template:

```hbs
<div id={{trigger.butto}}>...</div>
{{!       cursor here /\         }}
```

Then as I'm typing, the only suggestion I get is `GamepadButton'

<img width="450" alt="Code suggesting `GamepadButton`" src="https://user-images.githubusercontent.com/108688/197197327-c6b4ee8d-13a3-445c-8363-db7aafd743c2.png">

If I hit escape to clear that suggestion and manually trigger again with <kbd>Ctrl</kbd>-<kbd>Space</kbd>, I get a suggestion for the expected `buttonId` property.

<img width="445" alt="Code suggesting `buttonId`" src="https://user-images.githubusercontent.com/108688/197197565-5f0439c0-69aa-4739-a576-04d11047ab53.png">

I'm not sure if this is still somehow an issue with trying to service the completion request before the document contents have updated, or a difference in implicit vs explicit completion behavior, or what, but it seems clear that we know what the correct answer is in a vacuum but sometimes fail to give it when things are on the move.

PR #442, opened 2022-10-26
Further autocomplete improvements

dfreeman:
As with #393, this still doesn't solve our underlying "only syntactically-valid templates can produce completions" problem, but it _does_ resolve the main issue called out in #436.

At a high level, several different related issues were coming together to produce the behavior described in #436:
 - completions triggered as the result of typing a partial identifier are always considered `CompletionTriggerKind.Invoked` (this is per the LSP, not just a weird choice on Code's part)
 - we weren't forwarding TS's sorting information through the LSP, so items didn't necessarily display in order of relevance
 - standalone templates with a syntax error end up mapping to a 0-length span at the top level of the corresponding backing module, meaning that when we tried to produce completions for e.g. `{{foo.}}`, we'd return the full list of hundreds of available globals

Fixing those three errors, completion now looks like this:

Glint providing mostly-reasonable completion for a let-bound local value and one of its properties in a template

(Note that in a loose-mode template we can't provide completions for `myHashObject`, as we have to assume during emit that e.g. `myH` is referring to some global and not an incomplete reference to a local identifier)

### DIFF
--- a/packages/core/__tests__/language-server/completions.test.ts
+++ b/packages/core/__tests__/language-server/completions.test.ts
@@ -57,6 +57,26 @@ describe('Language Server: Completions', () => {
     expect(completions).toBeUndefined();
   });
 
+  test('in a template with syntax errors', () => {
+    project.setGlintConfig({ environment: 'ember-loose' });
+
+    let code = stripIndent`
+      Hello, {{this.target.}}!
+    `;
+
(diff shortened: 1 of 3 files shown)


chriskrycho:
This is a big improvement. Thank you, as ever.

PR #443, opened 2022-10-29
Fix unused hash

### LINKED ISSUE #374: Strict mode: hash is required to be imported, but glint reports it as not used.
Sample gjs:
```js
import { hash } from '@ember/helper';

<template>{{yield (hash x=1)}}</template>
```

Sample Error:
```
app/components/limber/output/compiler/index.gts:3:1 - error TS6133: 'hash' is declared but its value is never read.

3 import { hash } from '@ember/helper';
```

Since removing the import creates a runtime error, "hash is not defined", I think Glint should not report the hash import as never read?

I'm using 
 - ember-source 4.5
 - ember-template-imports 3.0.1
 - @glint* 0.9.1

wagenet:
Fixes #374 

I've not actually tested the behavior of this in a real app yet.

### DIFF
--- a/packages/environment-ember-template-imports/-private/dsl/index.d.ts
+++ b/packages/environment-ember-template-imports/-private/dsl/index.d.ts
@@ -93,3 +93,5 @@ export declare function templateExpression<
 >(
   f: (𝚪: Context, χ: never) => void
 ): TemplateOnlyComponent<never> & (new () => Invokable<Signature> & HasContext<Context>);
+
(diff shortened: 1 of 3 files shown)


wagenet:
Turns out this isn't so straight-forward since, while this works for ember-template-imports it does not work for ember-loose.

dfreeman:
Thanks @wagenet! This was taken care of as part of a larger reworking in #453

PR #445, opened 2022-11-02
Update CI config

dfreeman:
We're getting a number of deprecations from GitHub in our Actions runs. This should take care of all of them except one, pending https://github.com/GabrielBB/xvfb-action/pull/31


dfreeman:
I'm also removing the dependency from the `Floating Dependencies` and `TypeScript Nightly` jobs on the base `Tests` one, as it's no fun to wait for those after even _Windows_ has finished running.

Issue #411, opened 2022-09-13
Discussion/pre-RFC: what is the right long-term home for environment types?


chriskrycho:
Today, Glint supplies all of the type definitions it requires to work via a combination of the [@glint/template](https://github.com/typed-ember/glint/tree/main/packages/template) package (the baseline for the rest) and the various environments which extend that appropriately for their own context. This is serving us well so far, but it has the potential to cause some issues down the line, which I want to note down as I'm thinking about them today.

One of the frustrations with DefinitelyTyped is that there is no good way to match the versioning between the types on DT with the actual version of any given package, particularly around *new features added in minor releases*. For example, we recently had a very annoyed user complaining about the Ember 3.x types including all APIs released up through 3.28 on types which *worked* for any Ember 3.x consumer, and thereby providing some misleading suggestions for APIs which do not exist on the version he was on. (I do not want to over-index on that particular complaint, of course, but it is illustrative of the challenge.) *We have the same potential issue for every item which appears in the globals list for any given environment.*

Consider the introduction—hopefully soon!—of the equality and comparison keywords: these will need to be added to the globals/built-ins list in `environment-ember-loose`, but they will only actually work correctly in Ember on a version which actually includes that functionality.[^polyfill] We do not have any particularly *good* way to signal which versions of Glint environment types are compatible with a given Ember version. We can (and, as long as we are taking this tack, definitely *should*) document it, but we also then have the problem of figuring out how to handle bug fix releases—do we end up having `@glint/template` with `1.1.x` and `1.2.x` maintenance streams ongoing, where `1.2.x` supports Ember 4.10+, while `1.1.x` does not? Etc.

Net, I propose that medium-to-long-term, the Glint environments should move *out* of Glint to the packages which they represent. Now, that has its own potential problems in that we could end up with Glint depending on types from those packages and those packages depending on types from Glint. One way to cut that knot is to make sure that `@ember/glint` (hypothetical future package) is a real package which is versioned in lockstep with and published out of the same repo as `ember-source`, and to then guarantee that the direction of dependencies only ever running these ways:

- `ember-source` ← `@glint/template`
- `ember-source` ← `@ember/glint`
- `@glint/template` ← `@ember/glint`

[...]

Issue #396, opened 2022-08-29
Meta issue: function helper resolution bugs


dfreeman:
I'm opening this to track issues with function helpers that we expect will be fixed with our [planned overhaul](https://tsplay.dev/mM1zQm) of the internal representation of invokable template entities within Glint.

dfreeman:
### `unknown` is treated as named-args-like

Given this function,
```ts
declare function isEqual(a: unknown, b: unknown): boolean;
```

This will resolve as a helper with one positional arg and shapeless named args.

dfreeman:
### Named-args-like args _have_ to be passed as named args

Given this function:
```ts
declare function constructLanguageTag(options: { country: string; language: string }): string;
```

This should be legal to invoke as either of the following:
```hbs
{{constructLanguageTag country="US" language="en"}}
{{constructLanguageTag (hash country="US" language="en")}}
```

However, currently only the former is accepted.

boris-petrov:
### Class args don't work in any way

Given:

```ts
class Foo {}
const someFooInstance: Foo;
declare function handleFoo(foo: Foo): string;
```

This should be legal:

```
{{handleFoo someFooInstance}}
```

However, currently there is an error - `Expected 0 arguments, but got 1`.

dfreeman:
Not a helper resolution bug, but something we should verify as part of overhauling `[Invoke]`: we should make sure assignability works as expected with `ComponentLike`, `WithBoundArgs`, etc.

As a concrete example, this component:

```ts
class MyComponent extends Component<{
  Args: { foo?: string };
  Element: HTMLImageElement;
  Blocks: { default: [] };
}> {}
```

Should be assignable to (among other things) `ComponentLike` with no parameters, since it's legal to invoke with no args, `...attributes` or blocks.

PR #447, opened 2022-11-03
Overhaul internal `[Invoke]` signature representation

dfreeman:
## Background

Internally, we represent the signature of template-invokable entities along these lines:

```ts
[Invoke]: (namedArgs: Named, ...positionalArgs: Positional) => Result
```

Where `Result` may be a bare return type for a helper, or `BoundModifier<ElementType>` or a modifier, or `AcceptsBlocks<BlockParams, ElementType>` for a component.

This worked reasonably, allowing us to invoke a call like `{{something 1 2 three=3}}` as `resolve(something)({ three: 3 }, 1 , 2)`. Since named args always came first, there was no ambiguity about named vs positional arguments in our signatures.

Then the default helper manager RFC came, and it introduced ambiguity between named vs positional arguments. For example, a function helper `join(items, { separator })` could be correctly be invoked as either `{{join items separator=', '}}` or {{join items (hash separator=', ')}}`. Because we have to normalize that function to our "named args first" format, we can't handle that ambiguity: we have to choose to only support one or the other.

#396 contains a collection of such issues that result from our having to force function helpers to match the shape of our internal representation.

## This Change

Following this change, our internal representation of the signature of template-invokable entities puts named args at the end, matching how native function helpers work.

```ts
[Invoke]: (...positionalArgs: Positional, named: Named) => Result
```

Where `named` may be optional if `Named` could be empty, and `Result` might be either a bare return type for a helper, `ModifierReturn` for a modifier, or `ComponentReturn<Blocks, Element>` for a component.

### Function Resolution

Aligning our internal representation with the one used by function helpers means we can now just let functions resolve as themselves (concretely, `resolve<T extends AnyFunction>(item: T): T`). This in turn:
 - massively reduces the number of `resolve()` signatures we need
 - avoids having to deconstruct and reconstruct function types during resolution
 - allows us to support arbitrary type guard and type assertion helper types
 - avoids the ambiguity issues captured in #396

### Component and Modifier Changes

The completion type for components changed from `AcceptsBlocks<Blocks, Element>` to `ComponentReturn<Blocks, Element`, but this was just a rename for better clarity and consistency.

The completion type for modifiers changed from `BoundModifier<Element>` to an unparametrized `ModifierReturn` type. To account for this, modifiers now receive the element itself as their first positional argument (which aligns with function-based modifiers in GlimmerX today), and the `applyModifier` DSL function simply validates the return type.

This may change in the future if or when the default modifier manager RFC advances, but in the meantime this change should have no external impact other than fixing a variance discrepancy.

### Named Args

In the example above, our `join(items, { separator })` function could be invoked by either passing two positional parameters or by passing one positional param and a named `separator` param.

However, for an equivalent classic-style `helper(([items], { separator }) => ...)`, it's _not_ legal to pass the named args as a second positional argument, since the helper manager distinguishes between the two. To account for this, the resulting `[Invoke]` signature for such a helper ends up looking like;

```ts
(items: Array<unknown>, named: NamedArgs<{ separator?: string }>) => string
```

The `NamedArgs` wrapper type requires an additional symbol key that the transformation layer provides when an invocation uses named arg syntax rather than a positional `(hash)` expression.

Trying to invoke this helper as `{{join items (hash separator=', ')}}` would produce an error like:
```
Argument of type '{ separator: string; }' is not assignable to parameter of type 'NamedArgs<{ separator?: string | undefined; }>'.
  Property '[NamedArgs]' is missing in type '{ separator: string; }' but required in type 'NamedArgsMarker'. glint:ts(2345)
```

### Tradeoffs

While this change greatly simplifies our resolution logic and improves expressiveness for authors using function helpers, it's not a free win.

#### Positional/Named Conflation

By moving named args to the end of the function signature, we lose the ability to validate entities that accept both an unbounded number of unknown positional arguments _and_ named args.

This comes down to limits in TypeScript function types, which in turn are influenced by ECMAScript's own restrictions: namely that you can't define a function like this:

```js
function myHelper(...positional, named) {}
```

The above function produces `SyntaxError: Rest parameter must be last formal parameter` when parsed.

In TypeScript we can work around this in a limited way by typing the function as:

```ts
function myHelper(...args: [...positional: Positional, named?: Named]) {}
```

However, if `Positional` is something like `unknown[]`, then TS can no longer validate the final parameter containing named args, since that's a perfectly valid `unknown` value. This scenario is relatively rare, but not unheard of: the `{{link-to}}` builtin is an example of one entity that has a signature like this.

#### Arity Error Reporting

This change also makes the existence of named args as functionally a positional parameter more visible to end users. When a signature always had named args, and thus invocation always passed _some_ value there, we could always simply subtract 1 when reporting arity errors.

Now, because a signature may or may not have named args in it, we can't always account for that. Instead, when reporting an arity error, if the invocation includes named args, we append a note to the diagnostic that named args together "count" as 1 argument.

Ultimately I think these tradeoffs are reasonable for what we gain, but they certainly bear mention.

## Notes

I'm going to leave #396 open for now, as the assignability issues raised in the last comment there aren't yet resolved (though this change helps unblock that!)

I'm planning a follow-up PR that I expect will introduce our own `ExpandComponentSignature` separate from `@glimmer/component`'s (for reasons outlined in https://github.com/typed-ember/glint/issues/411#issuecomment-1267038048), and that should enable us to ensure assignability works as we want.


packages/template/__tests__/emit-component.test.ts
```diff
@@ -40,31 +41,27 @@ class MyComponent<T> extends TestComponent<MyComponentSignature<T>> {
...
     templateForBackingValue(this, function (𝚪) {
-      const component = emitComponent(resolve(globals.let)({}, 𝚪.this.state.ready));
+      const component = emitComponent(resolve(globals.let)(𝚪.this.state.ready));
       const [isReady] = component.blockParams.default;
 
       {
         const 𝛄 = emitElement('div');
         expectTypeOf(𝛄).toEqualTypeOf<{ element: HTMLDivElement }>();
-        applyModifier(𝛄.element, resolve(globals.on)({}, 'click', 𝚪.this.wrapperClicked));
+        applyModifier(resolve(globals.on)(𝛄.element, 'click', 𝚪.this.wrapperClicked));
```
chriskrycho, line 50: Thinking about the general implementation, and looking at this specific example—to your point about the representation changing if the default modifier manager is approved and ships something *different* than GlimmerX ships, do we have any risk from that? Particularly, do we have any risk of stuff that will surface as a breaking change for our consumers?

packages/template/-private/keywords/component.d.ts
```diff
@@ -1,19 +1,58 @@
...
+} from '../integration';
+
+type PartiallyAppliedComponent<Component, Args> = Component extends Invokable<AnyFunction>
+  ? WithBoundArgs<
+      Component,
+      Exclude<Args extends NamedArgNames<Component> ? Args : never, typeof NamedArgs>
+    >
+  : never;
 
 export type ComponentKeyword = DirectInvokable<{
```
chriskrycho, line 20: This definition form hurts my head. 😂 

What would you think of extracting the type as an `interface` or `type` alias here? It took me actually opening it locally and following the definition of the `DirectInvokable` type to see that it was a function to actually be able to *parse* this properly as a type with two overloads of a function definition. Extracting it to a (non-exported) `type` alias should preserve the exact same semantics we have as is, but make it a bit clearer what's actually going on.

chriskrycho: (Also, I recognize that this definition form isn't new, but while it was a bit confusing before the overload *really* hurt my head when reading it now.)

packages/template/__tests__/keywords/debugger.test.ts
```diff
@@ -4,7 +4,7 @@ import { DebuggerKeyword } from '../../-private/keywords';
 const debuggerKeyword = resolve({} as DebuggerKeyword);
 
 // Can be invoked as {{debugger}}
-emitContent(debuggerKeyword({}));
+emitContent(debuggerKeyword());
```
chriskrycho, line 7: The lack of extra empty objects in so many places really is nice. (Noticed repeatedly, just thought to comment on it on seeing this one!)

packages/template/__tests__/keywords/in-element.test.ts
```diff
@@ -7,24 +7,24 @@ declare const element: HTMLElement;
...
 
 // Can be invoked with a ShadowRoot
-emitComponent(inElementKeyword({}, shadow));
+emitComponent(inElementKeyword(shadow));
 
 // Accepts an `insertBefore` argument
-emitComponent(inElementKeyword({ insertBefore: null }, element));
-emitComponent(inElementKeyword({ insertBefore: undefined }, element));
+emitComponent(inElementKeyword(element, { insertBefore: null, ...NamedArgsMarker }));
+emitComponent(inElementKeyword(element, { insertBefore: undefined, ...NamedArgsMarker }));
```
chriskrycho, line 17: Not related to this PR, but is `undefined` allowed? [The docs][docs] say only `undefined` is.

> It also supports an optional named argument:
> 
> - `insertBefore` -- by default the DOM element's content is replaced when used as `destinationElement`. Passing `null` changes the behaviour to appended at the end of any existing content. Any other value than `null` is currently not supported.

[docs]: https://api.emberjs.com/ember/4.7/classes/Ember.Templates.helpers

packages/environment-glimmerx/-private/dsl/index.d.ts
```diff
@@ -26,32 +26,23 @@ import {
...
-): (named: EmptyObject, arg: unknown, ...args: Args) => T;
-export declare function resolve<El extends Element, Args extends unknown[]>(
-  item: (element: El, ...args: Args) => void | (() => void)
-): (named: EmptyObject, ...args: Args) => ModifierReturn<El>;
-export declare function resolve<Args extends unknown[], T>(
-  item: (...args: Args) => T
-): (named: EmptyObject, ...args: Args) => T;
+export declare function resolve<T extends ((...params: any) => any) | null | undefined>(
+  item: T
+): NonNullable<T>;
```
chriskrycho, line 45: Oh thank goodness. So many fewer overloads. 😅

packages/environment-ember-loose/-private/intrinsics/action.d.ts
```diff
@@ -12,36 +12,34 @@ export type ActionResult<T, Args extends ActionNamedArgs<T>> = undefined extends
 
 export type ActionKeyword = DirectInvokable<{
   <Ret, Args extends ActionNamedArgs<Ret>, Params extends unknown[]>(
-    args: Args,
-    f: (...rest: Params) => Ret
+    f: (...rest: Params) => Ret,
```
chriskrycho, line 15: The pattern does begin to feel less weird once you've internalized it, and it would be a bit annoying to extract everywhere, so maybe that's an argument *not* to extract like I suggested above.

packages/template/__tests__/emit-component.test.ts
dfreeman: In principle this should be invisible to consumers (other than incidentally fixing the variance issue I mentioned, but we would've done that regardless). If Ember ships something different than GlimmerX, what we'll likely do is update the internal representation to match that and add a `resolve()` signature to the GlimmerX environment that converts its function signature to whatever Ember lands on.

While this _could_ have observable effects for GlimmerX users, the chances are slimmer there since modifiers tend not to do a lot of the type gymnastics that helpers do (no assertions/guards, no parametric return types, etc). I also imagine it's reasonably likely that GlimmerX would eventually align with Ember, allowing us to drop that extra resolution, but if not it's still not a huge deal.

packages/template/__tests__/keywords/in-element.test.ts
dfreeman: I wasn't the original author here, but I think at runtime passing `undefined` is treated the same as not specifying `insertBefore` at all. Whether we should allow explicit `undefined` or not is a conversation I'm happy to have (my leaning is to say no), but I'd rather not change the semantics as part of this PR either way.

packages/template/-private/keywords/component.d.ts
dfreeman: Pulling out an `interface` or `type` here would only remove the surrounding `DirectInvokable<...>` wrapper, I think. Beyond that the syntax would still be the same.

The only way to specify a multi-signature function type otherwise, to my knowledge, would be to declare the function as a value and then `typeof` it, which I'd lean away from—I don't love declaring nonexistent constants when it's avoidable.

packages/template/__tests__/keywords/in-element.test.ts
chriskrycho: Totally agreed about not changing it in this PR.

packages/template/-private/keywords/component.d.ts
chriskrycho: Yeah, I [eventually concluded](https://github.com/typed-ember/glint/pull/447#discussion_r1014509850) the same.

chriskrycho (approved):

PR #453, opened 2022-11-10
Generalize handling of special forms

### LINKED ISSUE #374: Strict mode: hash is required to be imported, but glint reports it as not used.

dfreeman:
## Background

Currently, Glint has a hard-coded list of 5 global names ("keywords") that, when they appear as the head of a `CallNode`, will be translated in a special way to TypeScript instead of being treated as a generic `Invokable`:
 - `if`
 - `unless`
 - `yield`
 - `hash`
 - `array`

This (mostly) works for the traditional `ember-loose` environment, but it has some shortcomings that become more apparent in GlimmerX and strict-mode Ember.

GlimmerX actually doesn't have baked-in `hash` and `array` helpers, but we incorrectly treat them as supported.

In strict-mode Ember, `hash` and `array` aren't globals, but rather imports from `@ember/helper`. In the best case scenario, we end up flagging those imported helpers as unused (#374), but we also fail to handle them correctly if they're imported under another name, and we fail to flag a problem at all if the developer forgets the import.

Relatedly, as we consider #411 and think about what it means for environments to evolve and introduce new values that we want to specially translate, this current approach won't scale well at all.

## Change Overview

This PR generalizes the globals listed above to a notion of "special forms" (terminology used by [several lisp dialects](https://courses.cs.northwestern.edu/325/readings/special-forms.php)).

The set of available special form translations is fixed by `@glint/transform`, but what global and/or imported identifiers are subject to that special treatment is specified by the active environment.

As an example, the configuration that `@glint/environment-ember-template-imports` provides as of this PR looks like this:

```ts
specialForms: {
  globals: {
    if: 'if',
    unless: 'if-not',
    yield: 'yield',
  },
  imports: {
    '@ember/helper': {
      array: 'array-literal',
      hash: 'object-literal',
    },
  },
},
```

The end result is a model similar to that of `@babel/parser`: the core library is aware of all possible special syntax, but enabling support for it is managed by other packages that effectively flip the switch on.

Adopting this approach will allow us to add support for things like logical and comparison operators in `@glint/template` at any time, but only enable them as appropriate when releasing a new version of (the currently hypothetical) `@ember/glint-environment` that corresponds to the version of Ember where they land.

Fixes #374


chriskrycho (approved):
This is really, really smart. 👍🏼

PR #454, opened 2022-11-11
Merge the `transform` and `config` packages into `core`

dfreeman:
Early in Glint's lifetime, there were reasons a given consumer might want to use `@glint/config` or `@glint/transform` without actually running the language server or CLI from `@glint/core`. The VS Code extension was one example of this.

Today, though, `@glint/core` is the only real consumer of those two packages, and neither of them has a user-facing programmatic API that we intend to support.

Accordingly, this PR merges `@glint/config` and `@glint/transform` into `@glint/core` to help simplify our overall package topology.


dfreeman:
@bwbuchanan I'm sorry, I'm about to pull the rug out from under you in `packages/transform` again—this is just moving files around, so hopefully shouldn't be too terrible to resolve!

chriskrycho (commented):
👍

PR #456, opened 2022-11-11
Add support for equality operators (Ember RFC 560)

### LINKED ISSUE #221: Support type-narrowing via discriminated unions
[Type-narrowing via discriminated unions](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) is a useful TypeScript feature. I believe it's not supported via Glint:

```ts
interface Circle {
  kind: "circle";
  radius: number;
}

interface Square {
  kind: "square";
  sideLength: number;
}

type Shape = Circle | Square;

export default class Foo extends Component<...> {
  public declare shape: Shape;

  public get foo(): number {
    return this.shape.kind === 'circle' ? this.shape.radius : this.shape.sideLength;
  }
}
```

```hbs
{{#if (eq this.shape.kind "circle")}}
  {{log this.shape.radius}}
{{/if}}
```

The `foo` getter compiles just fine as the TypeScript compiler does it's job, however the `log` in the HBS gives an error:

```
error TS2339: Property 'radius' does not exist on type 'Shape'.
  Property 'radius' does not exist on type 'Square'.

6   {{log this.shape.radius}}
                     ~~~~~~
```

It would be nice to be able to do this. Not sure it's possible though. At the very least a "built-in" `eq` helper must be defined and Glint should be hard-coded to know about it I guess.

bwbuchanan:
Resolves #221

As these helpers have not yet been merged into Ember, you can opt in by adding to your glint config:

```
"glint": {
  "environment": {
    "ember-loose": {
      "additionalSpecialForms": {
        "globals": {
          "eq": "===",
          "not-eq": "!=="
        }
      }
    },
    "ember-template-imports": {
      "additionalSpecialForms": {
        "imports": {
          "ember-truth-helpers/helpers/eq": { "default": "===" },
          "ember-truth-helpers/helpers/not-eq": { "default": "!==" }
        }
      }
    }
  }
}
```

### DIFF
--- a/packages/core/src/transform/template/template-to-typescript.ts
+++ b/packages/core/src/transform/template/template-to-typescript.ts
@@ -195,6 +195,11 @@ export function templateToTypescript(
           emitArrayExpression(formInfo, node);
           break;
 
+        case '===':
+        case '!==':
+          emitBinaryOperatorExpression(formInfo, node);
+          break;
+
         default:
           record.error(`${formInfo.name} is not valid in inline form`, rangeForNode(node));
           emit.text('undefined');
@@ -311,6 +316,27 @@ export function templateToTypescript(
           emit.text('undefined');
         }
 
+        emit.text(')');
+      });
+    }
+
+    function emitBinaryOperatorExpression(
+      formInfo: SpecialFormInfo,
+      node: AST.MustacheStatement | AST.SubExpression
+    ): void {
+      emit.forNode(node, () => {
+        assert(
+          node.hash.pairs.length === 0,
+          `{{${formInfo.name}}} only accepts positional parameters`
+        );
+        assert(node.params.length === 2, `{{${formInfo.name}}} requires exactly two parameters`);
+
+        const [left, right] = node.params;
+
+        emit.text('(');
+        emitExpression(left);
+        emit.text(` ${formInfo.form} `);
+        emitExpression(right);
         emit.text(')');
       });
     }
--- a/packages/core/src/transform/template/template-to-typescript.ts
+++ b/packages/core/src/transform/template/template-to-typescript.ts
@@ -327,9 +327,12 @@ export function templateToTypescript(
       emit.forNode(node, () => {
         assert(
           node.hash.pairs.length === 0,
-          `{{${formInfo.name}}} only accepts positional parameters`
+          () => `{{${formInfo.name}}} only accepts positional parameters`
         );
-        assert(node.params.length === 2, `{{${formInfo.name}}} requires exactly two parameters`);
+        assert(
+          node.params.length === 2,
+          () => `{{${formInfo.name}}} requires exactly two parameters`
+        );
 
         const [left, right] = node.params;
 
--- a/packages/environment-ember-template-imports/-private/environment/index.ts
+++ b/packages/environment-ember-template-imports/-private/environment/index.ts
@@ -1,8 +1,17 @@
-import { GlintEnvironmentConfig } from '@glint/core/config-types';
+import { GlintEnvironmentConfig, GlintSpecialFormConfig } from '@glint/core/config-types';
 import { preprocess } from './preprocess';
 import { transform } from './transform';
 
-export default function emberTemplateImportsEnvironment(): GlintEnvironmentConfig {
+export default function emberTemplateImportsEnvironment(
+  options: Record<string, unknown>
+): GlintEnvironmentConfig {
+  let additionalSpecialForms =
+    typeof options['additionalSpecialForms'] === 'object'
+      ? (options['additionalSpecialForms'] as GlintSpecialFormConfig)
+      : {};
+
+  const additionalGlobalSpecialForms = additionalSpecialForms.globals ?? {};
+
   return {
     tags: {
       'ember-template-imports': {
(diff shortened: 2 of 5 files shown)


packages/environment-ember-template-imports/-private/environment/index.ts
```diff
@@ -13,12 +22,15 @@ export default function emberTemplateImportsEnvironment(): GlintEnvironmentConfi
...
               yield: 'yield',
+              ...additionalGlobalSpecialForms,
             },
             imports: {
               '@ember/helper': {
                 array: 'array-literal',
                 hash: 'object-literal',
+                ...(additionalSpecialForms.imports?.['@ember/helper'] ?? {}),
               },
+              ...(additionalSpecialForms.imports ?? {}),
```
dfreeman, line 33: Spreading `null` or `undefined` is just a no-op, so anywhere in these you have `...(something ?? {})` can just be `...something` instead

packages/core/src/transform/template/template-to-typescript.ts
```diff
@@ -315,6 +320,27 @@ export function templateToTypescript(
...
+    function emitBinaryOperatorExpression(
+      formInfo: SpecialFormInfo,
+      node: AST.MustacheStatement | AST.SubExpression
+    ): void {
+      emit.forNode(node, () => {
+        assert(
+          node.hash.pairs.length === 0,
+          `{{${formInfo.name}}} only accepts positional parameters`
+        );
+        assert(node.params.length === 2, `{{${formInfo.name}}} requires exactly two parameters`);
```
dfreeman, line 332:
```suggestion
        assert(
          node.hash.pairs.length === 0,
          () => `{{${formInfo.name}}} only accepts positional parameters`
        );
        assert(
          node.params.length === 2,
          () => `{{${formInfo.name}}} requires exactly two parameters`
        );
```

This is a tiny optimization that _probably_ matters very little in the end, but by making these assertion messages thunks we can avoid doing string concatenations every time we emit a binop expression.

dfreeman (commented):
A couple tiny nits, but the overall change looks great here. Thank you for driving this forward, @bwbuchanan!

dfreeman (approved):

dfreeman:
Thank you!

boris-petrov:
I just saw that I don't handle `lt`, `gt`, etc in my `tsconfig.json`. When I tried, it complained with `lt is not valid in inline form`. Does that mean that these operators have to also be implemented separately?

dfreeman:
@boris-petrov see https://github.com/typed-ember/glint/issues/221#issuecomment-1368933322

Comparison operators don't have any special semantics for TS, so they don't need to be implemented as special forms. You can just add them to the template registry as usual.

bwbuchanan:
> I just saw that I don't handle `lt`, `gt`, etc in my `tsconfig.json`. When I tried, it complained with `lt is not valid in inline form`. Does that mean that these operators have to also be implemented separately?

@boris-petrov 

No special-case handling is needed in glint for `lt`, `gt`, etc. You can import them directly in your templates but I don't think that `ember-truth-helpers` is yet shipping typedefs, so you'll need to define them yourself in your project, e.g.

types/ember-truth-helpers/gt.d.ts
```typescript
import { HelperLike } from '@glint/template';

const gt: HelperLike<{
  Args: { Positional: [value1: number, value2: number] };
  Return: boolean;
}>;

export default gt;
```

boris-petrov:
Thanks both for the answer and sorry for the late response on my side.

I already do what @bwbuchanan suggests - i.e. I already have type-definitions for them. I thought that TypeScript does some type-narrowing (or `null`/`undefined`-removal) when it sees `<` for example but I guess I was mistaken and that's not true. So yes, my request here doesn't make sense. Thanks for the time!

bwbuchanan:
I did a couple of TypeScript experiments and could not find a case in which numeric comparison operators resulted in type-narrowing of their operands.

bwbuchanan:
[ember 7.1.0](https://github.com/emberjs/ember.js/blob/main/CHANGELOG.md#v710-beta1-may-12-2026) lands `and`, `or`, `not`, `eq`, `neq` (not `not-eq` / `notEq`), as built-ins and also removes the need to explicitly import `hash`, `array`, `on`, `fn`.

I'm not sure if glint can detect which ember version is in use (i.e. which builtins exist) or whether we need to add these to the `additionalSpecialForms.globals` config.

NullVoxPopuli:
This is already done, make sure you use latest glint with ember 7.1

boris-petrov:
Is this going to work correctly on Glint 1 and Ember 7.1?

NullVoxPopuli:
Nay, only glint 2

dfreeman:
You can still opt-in with `additionalSpecialForms` in your Glint 1 config

bwbuchanan:
> This is already done, make sure you use latest glint with ember 7.1

Are you sure about this? I'm using @glint/ember-tsc 1.8.0 with ember 7.1.0 and `eq` doesn't seem to be usable for type-narrowing discriminated unions without explicitly defining `additionalSpecialForms.globals`. I also couldn't find any current code that was declaring these to be special forms for ember 7.1.0, only code declaring them as globals.

PR #458, opened 2022-11-14
Add an export statement to */-private/index.ts

### LINKED ISSUE #455: Spurious error when using --isolatedModules in development
https://github.com/typed-ember/glint/blob/e35714c5fbccf279eaf476dd6e0b2161397085d2/packages/environment-ember-loose/-private/index.ts

This file needs `export {};` to avoid a spurious error when typechecking a project that uses `--isolatedModules`.

```
../../../repos/glint/packages/environment-ember-loose/-private/index.ts:1:1 -
error TS1208: 'index.ts' cannot be compiled under '--isolatedModules' because it is considered a global
script file. Add an import, export, or an empty 'export {}' statement to make it a module.
```

This issue only affects use of a directly-linked glint repo (e.g. during development / testing of glint itself using a different project as the subject). The generated `index.d.ts` does contain the `export {};`, so users of the npm package are unaffected.

bwbuchanan:
This avoids compile error when building with --isolatedModules, which affects projects that directly link the sources for glint instead of importing it from npm.

There is no change to the compiled output.

Fixes #455

### DIFF
--- a/packages/environment-ember-loose/-private/index.ts
+++ b/packages/environment-ember-loose/-private/index.ts
@@ -4,3 +4,4 @@
 
 /// <reference path="../registry/index.d.ts" />
 /// <reference path="./dsl/integration-declarations.d.ts" />
+export {};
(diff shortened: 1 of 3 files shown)


dfreeman:
👍  Makes sense, thanks!

PR #461, opened 2022-11-15
Fix Get Helper

chadhietala:
Fixes Get

In the preview types `UnwrapComputedPropertyGetter` does not exist. This removes the import in favor of merging the declarations and branding to correctly infer the type.

@chriskrycho mentioned something about creating a test that isolates the types but seems tricky since `ember-source` is at 4.4 in this repo.


dfreeman:
Thank you!

> @chriskrycho mentioned something about creating a test that isolates the types but seems tricky since ember-source is at 4.4 in this repo

I think it's important that we validate that Glint works with both the DT types and the native ones, and that we guard against regressions on that front in the future. There's no reason we need to be locked to a single Ember version in our test apps—and in fact, I don't see a way to test both sources of truth for the types in a single app anyway.

Given that, I think what we'd want for this would be an additional Ember application under `test-packages` that's set up to consume the preview types. It wouldn't need to be a full-fledged app, or even necessarily be buildable, just a barebones setup we can validate typechecking against.

If that's something you have the time to do as part of this PR that would be great, but if not no worries—I'll make sure it happens at some point as part of the stuff queued up for our next release.

test-packages/ts-ember-preview-types/app/components/ember-component.hbs
```diff
@@ -0,0 +1,21 @@
+<div ...attributes id={{unique-id}}>
+  {{this.required}}
+  {{this.optional}}
+  {{this.hasDefault}}
+
+  {{@required}}
+  {{@optional}}
+  {{@hasDefault}}
+
+  {{get this "required"}}
```
chadhietala, line 10: I think this the correct way to test this? /cc @chriskrycho

chriskrycho: I believe so! And: I will try to help push this the rest of the way across the line today.

chriskrycho:
I've updated this in a couple ways:

- I updated the new test app to *only* have type deps, since we don't run it and only need to make sure the types are present. We should be able to update a bunch of the other test packages the same way.
- I added a check that this test package is *not* accidentally (read: via incorrect dependency hoisting) getting the DT types. Thanks to @dfreeman for that suggestion!
- I made sure that we didn't make any unrelated changes to the package.

Thanks for the initial work here, @chadhietala!

test-packages/ts-ember-preview-types/types/global.d.ts
```diff
@@ -0,0 +1,6 @@
+// Types for compiled templates
+declare module 'ts-ember-app/templates/*' {
+  import { TemplateFactory } from 'htmlbars-inline-precompile';
```
dfreeman, line 3: Unrelated to this PR, we should probably update the blueprint to either pull this from `ember-cli-htmlbars` or remove the declaration entirely 🤔

packages/environment-ember-loose/-private/intrinsics/get.d.ts
```diff
@@ -1,6 +1,15 @@
 import { DirectInvokable } from '@glint/template/-private/integration';
 import ObjectProxy from '@ember/object/proxy';
-import { UnwrapComputedPropertyGetter } from '@ember/object/-private/types';
+import '@ember/object/-private/types';
+
+declare const GetSetMarker: unique symbol;
```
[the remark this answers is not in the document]
chriskrycho: Absolutely.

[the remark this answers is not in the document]
chriskrycho:
```suggestion
// This hack lets us support both the stable/preview types from `ember-source`
// and the classic types still live on DefinitelyTyped. If using the DT types,
// this acts as a declaration merge for a module which is then integrated via
// the rest of the DT types. If using the stable/preview types, this ends up
// being a no-op.
declare const GetSetMarker: unique symbol;
```

dfreeman (approved):
Thank you, @chadhietala and @chriskrycho!

bendemboski:
Is there going to be a release that includes this anytime soon? As I understand it, this blocks the usage of the Ember preview types with GLint although if there's a workaround other than `"skipLibCheck": true` I'd love to hear about it!

chriskrycho:
@bendemboski it’s in all three of the 1.0.0 beta releases so far, including the one we cut last week. 😄 Note that the 1.0 beta releases are *more* stable, polished, and feature-ful than any of the 0.x releases were, so you should feel comfortable upgrading to them!

bendemboski:
Ooooooh, okay, thanks!

PR #463, opened 2022-11-18
Clean up package dependencies

dfreeman:
## Background

This change addresses some long-standing issues with the dependency declarations in some of our packages:
 1. Our environments should be declaring peerDeps on _every_ package they attempt to augment in order to ensure the target package is actually visible
 2. Environments currently declare a direct dependency on `@glint/template`, which means they can end up getting their own copy, but `@glint/template` really needs to be a highlander package because all consumers need to be on the same page about the symbols it uses for branding
 3. `@glint/template` forces a dependency on `@glimmer/component` for its `ExpandSignature` utility type, even though consumers of `@glint/template` may not need or want `@glimmer/component` themselves

## Change Overview

For (1), this primarily means we add a bunch of optional `peerDependencies` to the environment packages, which ultimately has minimal impact on consumers other than ensuring strict package managers arrange things so that those packages are visible to the environments for type augmentation.

For (2), we instead make `@glint/template` a _peer_ dependency of the environment packages. This is the key breaking change in this PR, as it requires end consumers to depend on `@glint/template` themselves. This is reasonable, though, as in practice those consumers end up frequently importing from `@glint/template` to add declarations to their own registries, and because as noted there _must_ only be one copy of the package in play.

For (3), we inline `ExpandSignature` from `@glimmer/component` into `@glint/template` and then actually deconstruct it slightly, since we never care about the combined expanded signature as a whole, but only ever its individual pieces. I also pulled in and adapted the types tests for `ExpandSignature` to make sure we're still honoring everything that worked there before.

## Notes

The change in (3) ultimately means we can probably simplify and stop exporting `ExpandSignature` from `@glimmer/component`, since Glint was meant to be the only external consumer of that, but I don't think there's any particular urgency there.

Also: with these changes we should be able to finally address the assignability issues raised in #396, and I _think_ at the end of everything we can probably drop `EmptyObject` entirely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ TypeScript-powered tooling for Glimmer templates.
 
 ## Getting Started
 
-Glint is broken into several different packages that you may use, depending on the details of your project. Typically you'll add `@glint/core` and a Glint environment package that reflects the type of project you're working on, then add a `"glint"` key to your `tsconfig.json` that tells Glint what it should look at.
+Glint is broken into several different packages that you may use, depending on the details of your project. Typically you'll add `@glint/core`, `@glint/template` and a Glint environment package that reflects the type of project you're working on, then add a `"glint"` key to your `tsconfig.json` that tells Glint what it should look at.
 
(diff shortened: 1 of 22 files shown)


chriskrycho (commented):
👍🏼

PR #465, opened 2022-11-21
Drop `hbs` tag support in `ember-template-imports`

dfreeman:
Aligning with the discussion in https://github.com/ember-template-imports/ember-template-imports/pull/18, this PR drops support for the `hbs` string tag from `ember-template-imports`. The expectation is that `ember-template-imports` itself is going to deprecate `hbs` soon, and since we have a good opportunity to make breaking changes at the moment, the `@glint/environment-ember-template-imports` package now _only_ operates on `<template>` literals.

### DIFF
--- a/packages/core/__tests__/cli/build-watch.test.ts
+++ b/packages/core/__tests__/cli/build-watch.test.ts
@@ -4,7 +4,7 @@ import { stripIndent } from 'common-tags';
 import stripAnsi = require('strip-ansi');
 import { afterEach, beforeEach, describe, expect, test } from 'vitest';
 
-import { Project, BASE_TS_CONFIG, INPUT_SCRIPT, setupCompositeProject } from '@glint/test-utils';
+import { Project, BASE_TS_CONFIG, INPUT_SFC, setupCompositeProject } from '@glint/test-utils';
 
 const BUILD_WATCH_TSCONFIG = {
   ...BASE_TS_CONFIG,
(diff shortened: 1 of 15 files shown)


chriskrycho:
BRUTAL.

Thanks for doing this!

dfreeman:
Swapping tests over to `.gts` uncovered some scenarios where custom extensions weren't playing nicely with `--build` and `--incremental` modes, and it also motivated me to finally fix the "we leave `<template>` in your file and generate a million errors if you have one syntax error in your template" papercut, so overall this should now be a nice little bundle of DX improvements for ETI users.

Hopefully CI agrees 🤞

chriskrycho:
Oh, interesting catch and good fix on the build and incremental modes.

flashios09:
@dfreeman I don't understand why `hbs` was removed !
The `<template>` syntax isn't compatible with typescript(`.ts`), force the usage of the `.gts`/`.gjs` extension
Why we can't just keep the 2 solutions for `.ts` and `.gts`/`.gjs` ?

With typescript(`.ts`):
```ts
import Component from "@glimmer/component";
import { action } from "@ember/object";
import { on } from "@ember/modifier";
import { hbs } from "ember-template-imports";

export default class MyButton extends Component {
  @action
  onClick() {
    // do something
  }
  static template = hbs(`
    <button type="button" {{on "click" this.onClick}}>{{yield}}</button>
  `, { scope: { on } });
}
```

chriskrycho:
Because that’s what we decided to do as a community with emberjs/rfcs#779!

flashios09:
@chriskrycho Does this mean that we will not be able to write SFC in typescript(`.ts`) anymore ?

chriskrycho:
Yes, the plan is to align everyone on `.gts`. 😄

flashios09:
Ah ok, thanks, it does make sense now :)

PR #467, opened 2022-11-22
Use a `./glint-environment-definition` entrypoint for env packages

dfreeman:
Glint environments need to expose two public entrypoints: one that triggers any type augmentations the environment provides, and one that exposes its actual definition to `@glint/core`.

Today, environments use their `main` entrypoint for the former, allowing consumers to write `import '@glint/environment-x'` to pull in type augmentations.

For the latter, we've used an ad-hoc `"glint-environment"` key in `package.json`. Now that we've switched environment packages over to conventionally use `exports`, though, this means that packages need to explicitly expose their `package.json` to be importable. Moreover, `exports` makes the `"glint-environment"` key itself an unnecessary layer of indirection.

This change moves to have environment packages expose a `./glint-environment-definition` entrypoint in their `exports` and does away with the old top-level `glint-environment` key. This simplifies our loading logic by just "using the platform".

This is a breaking change if anyone has reverse-engineered our environment interface and built their own, though to my knowledge that isn't super commonplace.

### DIFF
--- a/packages/core/__tests__/config/environment.test.ts
+++ b/packages/core/__tests__/config/environment.test.ts
@@ -163,7 +163,9 @@ describe('Config: Environments', () => {
         `${envDir}/package.json`,
         JSON.stringify({
           name: '@glint/environment-test-env',
-          'glint-environment': 'env',
+          exports: {
+            './glint-environment-definition': './env.js',
+          },
         })
       );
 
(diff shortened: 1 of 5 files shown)


chriskrycho (commented):
Retroactive 👍

PR #469, opened 2022-11-22
honor `compilerOptions.incremental` in build mode

### LINKED ISSUE #468: undefined buildOptions taking precedence over tsconfig compilerOptions entries
## System Info
glint --version: 0.9.7
node --version: v16.14.0

## Problem
when running `glint --build`, I am seeing the following error when building a project containing project references:
```
Error: Debug Failure. project /Users/hmajoros/Job/voyager-web/tsconfig.json expected to have at least one output
    at Object.getFirstProjectOutput (/Users/hmajoros/Job/voyager-web/node_modules/typescript/lib/typescript.js:110333:25)
    at updateOutputTimestamps (/Users/hmajoros/Job/voyager-web/node_modules/typescript/lib/typescript.js:125327:38)
    at Object.done (/Users/hmajoros/Job/voyager-web/node_modules/typescript/lib/typescript.js:124629:21)
    at build (/Users/hmajoros/Job/voyager-web/node_modules/typescript/lib/typescript.js:125406:32)
    at Object.build (/Users/hmajoros/Job/voyager-web/node_modules/typescript/lib/typescript.js:125599:101)
    at performBuild (/Users/hmajoros/Job/voyager-web/node_modules/@glint/core/lib/cli/perform-build.js:17:69)
    at Object.<anonymous> (/Users/hmajoros/Job/voyager-web/node_modules/@glint/core/lib/cli/index.js:97:42)
```

after a bit of debugging, it appears the issue is that glint is setting `buildOptions.incremental` to `undefined`, since `argv.incremental` is undefined.
https://github.com/typed-ember/glint/blob/main/packages/core/src/cli/index.ts#L84

the downstream effects are that in the typescript compiler, `hasProperty(buildOptions, 'incremental')` is true, which overrides the `compilerOptions.incremental` field with `undefined`.
https://github.com/microsoft/TypeScript/blob/00dc0b6674eef3fbb3abb86f9d71705b11134446/src/compiler/tsbuildPublic.ts#L335

this then causes our program to error out, because the top level project is not `composite` or `incremental`:
https://github.com/microsoft/TypeScript/blob/00dc0b6674eef3fbb3abb86f9d71705b11134446/src/compiler/emitter.ts#L502

and therefore we cannot generate any output `buildInfo`:
https://github.com/microsoft/TypeScript/blob/00dc0b6674eef3fbb3abb86f9d71705b11134446/src/compiler/emitter.ts#L722

## Possible Solution
modify the following code to only set `incremental` in buildOptions if it was passed as a command-line arg.
https://github.com/typed-ember/glint/blob/main/packages/core/src/cli/index.ts#L84

```typescript
  let buildOptions: TS.BuildOptions = {
    clean: argv.clean,
    force: argv.force,
    dry: argv.dry,
  };
  if ('incremental' in argv) {
    buildOptions.incremental = argv.incremental;
  }
```

hmajoros:
Fixes #468

could use help writing tests 🙈

### DIFF
--- a/packages/core/src/cli/index.ts
+++ b/packages/core/src/cli/index.ts
@@ -81,8 +81,10 @@ if (argv.build) {
     clean: argv.clean,
     force: argv.force,
     dry: argv.dry,
-    incremental: argv.incremental,
   };
+  if ('incremental' in argv) {
+    buildOptions.incremental = argv.incremental;
+  }
 
   // Get the closest TS to us, since we have to assume that we may be in the
   // root of a project which has no `Glint` config *at all* in its root, but


dfreeman:
@chriskrycho knows the `--build` test suite a bit better than I do, but I _think_ you should be able to add a test similar to [the one I wrote for the non-`build` case](https://github.com/typed-ember/glint/pull/428/files) to the as part of this ["single-pass build mode typechecking"](https://github.com/typed-ember/glint/blob/main/packages/core/__tests__/cli/build.test.ts#L18) block.

hmajoros:
@dfreeman debugged a bit with Chris this afternoon. I made a few changes to the `setupCompositeProject()` test helper, verified that any test running `projects.root.build()` failed, applied my changes, verified those tests are now passing.

NOW, the _new_ issue / topic: why does having an `include: ["./types-or-whatever/**/*"]` in the root tsconfig.json, paired with non-incremental builds cause TS to throw the error:
```
Error: Debug Failure. project /Users/hmajoros/Job/glint-test/tsconfig.json expected to have at least one output
```

cc @chriskrycho because this was confusing the both of us _a lot_ :joy:

dfreeman:
> why does having an include: `["./types-or-whatever/**/*"]` in the root tsconfig.json, paired with non-incremental builds cause TS to throw the error:

I... have no idea 😬 

Is it legal to combine `files: []` with a non-empty `include` array? (I'm just spitballing; maybe that's fine)

chriskrycho:
A bit of extra color for this and #468: the key here is not whether this is itself a recommended pattern (we’ve actually already planned to switch how we’re solving the thing this exposed in our repo), but that the `tsconfig.json` that triggered the odd failure in #468 works without issue when running `tsc`; the issue *only* shows up with when running `glint` (prior to this PR of course).

dfreeman (approved):
Thank you!

PR #470, opened 2022-11-22
Validate assignability rules for `ComponentLike` and friends

dfreeman:
## Background

Now that we've [overhauled `[Invoke]`](https://github.com/typed-ember/glint/pull/447) to have a simpler internal representation, we're better positioned to ensure that various `ComponentLike` values property reflects the real-world subtyping relationship between their types.

As a concrete example, a value with the bare `ComponentLike` type represents a component that can be invoked with no args, no blocks, and no modifiers. Given a definition like this:

```ts
class MyComponent extends Component<{
  Args: { name?: string };
  Element: HTMLImageElement;
  Blocks: { default: [] };
}> {}
```

The value `MyComponent` should be assignable to `ComponentLike`, because it _is_ legal to invoke without passing any args, blocks or modifiers.

If, however, the `name` arg were _not_ optional, then `MyComponent` should no longer be assignable to `ComponentLike`, as it would no longer be legal to invoke without any arguments.

The goal of this PR is to capture this and similar intuitions about the subtyping relationship between `ComponentLike`, `HelperLike` and `ModifierLike` values in the type system, and to put tests in place to ensure we maintain those relationships moving forward.

## Details

The steps taken for this change also had a couple of side effects that were nice in their own right.

First, landing #447 meant we no longer need the `EmptyObject` type, because now when an entity accepts no named args, we just don't include a parameter for them in `[Invoke]` at all, so we no longer need to worry about triggering EPC for `{}` cases.

Removing `EmptyObject` also, in an ouroboric way, enabled us to clean up the last place where we _did_ create a parameter unnecessarily for named args in `InvokableArgs`. This site was already EPC-enabled even without `EmptyObject` because of the `NamedArgs` wrapper, but it's nice not to have the parameter at all.

These two changes together ensure that the pairwise subtyping relationship between `Args` and `Blocks` works as expected, leaving only `Element`.

Internally we've used the same `null` marker as the public signature type does to represent the element of a component that never spreads its `...attributes`. This doesn't work for assignability purposes, as we want a component that _does_ spread `...attributes` on a particular element type to be usable in a context where we aren't going to bother passing attributes or modifiers. Accordingly, we now use `unknown` internally for this scenario.

### DIFF
--- a/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
@@ -7,7 +7,6 @@ import {
   NamedArgsMarker,
 } from '@glint/environment-ember-loose/-private/dsl';
 import { expectTypeOf } from 'expect-type';
-import { EmptyObject } from '@glint/template/-private/integration';
 import type { ComponentLike } from '@glint/template';
 
 {
@@ -50,7 +49,7 @@ import type { ComponentLike } from '@glint/template';
       templateForBackingValue(this, function* (𝚪) {
         expectTypeOf(𝚪.this.foo).toEqualTypeOf<string>();
         expectTypeOf(𝚪.this).toEqualTypeOf<StatefulComponent>();
-        expectTypeOf(𝚪.args).toEqualTypeOf<EmptyObject>();
+        expectTypeOf(𝚪.args).toEqualTypeOf<{}>();
       });
     }
   }
--- a/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
@@ -18,12 +18,6 @@ import type { ComponentLike } from '@glint/template';
     }
   }
 
-  resolve(NoArgsComponent)({
-    // @ts-expect-error: extra named arg
-    foo: 'bar',
-    ...NamedArgsMarker,
-  });
-
   resolve(NoArgsComponent)(
     // @ts-expect-error: extra positional arg
     'oops'
--- a/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
@@ -6,7 +6,6 @@ import {
   emitComponent,
   NamedArgsMarker,
 } from '@glint/environment-ember-loose/-private/dsl';
-import { EmptyObject } from '@glint/template/-private/integration';
 import { expectTypeOf } from 'expect-type';
 import { ComponentLike } from '@glint/template';
 
@@ -44,7 +43,7 @@ import { ComponentLike } from '@glint/template';
       templateForBackingValue(this, function* (𝚪) {
         expectTypeOf(𝚪.this.foo).toEqualTypeOf<string>();
         expectTypeOf(𝚪.this).toEqualTypeOf<StatefulComponent>();
-        expectTypeOf(𝚪.args).toEqualTypeOf<EmptyObject>();
+        expectTypeOf(𝚪.args).toEqualTypeOf<{}>();
       });
     }
   }
--- a/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
@@ -12,12 +12,6 @@ import { ComponentLike } from '@glint/template';
 {
   class NoArgsComponent extends Component {}
 
-  resolve(NoArgsComponent)({
-    // @ts-expect-error: extra named arg
-    foo: 'bar',
-    ...NamedArgsMarker,
-  });
-
   resolve(NoArgsComponent)(
     // @ts-expect-error: extra positional arg
     'oops'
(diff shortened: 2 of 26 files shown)


packages/environment-ember-loose/__tests__/type-tests/ember-component.test.ts
```diff
@@ -19,12 +18,6 @@ import type { ComponentLike } from '@glint/template';
     }
   }
 
-  resolve(NoArgsComponent)({
-    // @ts-expect-error: extra named arg
```
chriskrycho, line 23: Do we have coverage for this (just without the marker, I guess?) elsewhere?

packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
```diff
@@ -6,19 +6,12 @@ import {
...
 } from '@glint/environment-ember-loose/-private/dsl';
-import { EmptyObject } from '@glint/template/-private/integration';
 import { expectTypeOf } from 'expect-type';
 import { ComponentLike } from '@glint/template';
 
 {
   class NoArgsComponent extends Component {}
 
-  resolve(NoArgsComponent)({
-    // @ts-expect-error: extra named arg
```
chriskrycho, line 17: Same question here.

packages/environment-glimmerx/__tests__/component.test.ts
```diff
@@ -17,12 +17,6 @@ import { ComponentReturn, EmptyObject } from '@glint/template/-private/integrati
     });
   }
 
-  resolve(NoArgsComponent)({
```
chriskrycho, line 20: I feel like I'm missing something because it definitely seems like these should still be errors, if shaped differently. What am I not grokking?

packages/template/-private/signature.d.ts
```diff
@@ -41,21 +41,24 @@ export type ComponentSignatureBlocks<S> = S extends { Blocks: infer Blocks }
...
     }
-  : EmptyObject;
+  : {};
 
 /** Given a component signature `S`, get back the `Element` type. */
-export type ComponentSignatureElement<S> = S extends { Element: infer Element } ? Element : null;
-
-// These shenanigans are necessary to get TS to report when named args
-// are passed to a signature that doesn't expect any, because `{}` is
-// special-cased in the type system not to trigger EPC.
```
chriskrycho, line 51: This goes away now by way of handling it through arity rather than through parameter type, right?

packages/template/__tests__/component-like.test.ts
```diff
@@ -2,21 +2,16 @@ import { ComponentLike, WithBoundArgs } from '@glint/template';
 import { resolve, emitComponent, NamedArgsMarker } from '@glint/template/-private/dsl';
 import { expectTypeOf } from 'expect-type';
 import { ComponentReturn, NamedArgs } from '../-private/integration';
+import TestComponent from './test-component';
 
 {
   const NoArgsComponent = {} as ComponentLike<{}>;
 
+  // @ts-expect-error: extra arg
```
chriskrycho, line 10: 👍

chriskrycho (commented):
Retroactive review: this looks great overall and I think the changes all broadly make sense. I left some questions inline about removing rather than updating certain tests, but my one slightly bigger concern here is with the `null` → `unknown` transition for our internal representation on `Element`. The motivation made sense; is our error reporting going to be confusing, though?

packages/environment-ember-loose/__tests__/type-tests/glimmer-component.test.ts
dfreeman: Same as above

packages/environment-ember-loose/__tests__/type-tests/helper.test.ts
```diff
@@ -56,7 +56,9 @@ import { EmptyObject as GlintEmptyObject, NamedArgs } from '@glint/template/-pri
   let definition = helper(<T, U>([a, b]: [T, U]) => a || b);
   let or = resolve(definition);
 
-  expectTypeOf(or).toEqualTypeOf<{ <T, U>(t: T, u: U, named?: NamedArgs<EmptyObject>): T | U }>();
+  // Using `toMatch` rather than `toEqual` because helper resolution (currently)
```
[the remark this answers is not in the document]
dfreeman: It's something we can update in DT/Ember at some point, but there's no pressing urgency—it should no longer impact Glint one way or the other, and I don't expect it to be visible to other consumers at all.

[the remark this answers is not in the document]
chriskrycho: Excellent.

packages/environment-glimmerx/__tests__/component.test.ts
chriskrycho: Covered above.

packages/template/-private/signature.d.ts
dfreeman: Yep, exactly. (And actually, even if that weren't the case, the `NamedArgs` wrapper type we employ now would end up triggering EPC anyway)

dfreeman:
> my one slightly bigger concern here is with the `null` → `unknown` transition for our internal representation on `Element`. The motivation made sense; is our error reporting going to be confusing, though?

Our error reporting is already pretty confusing in the "you didn't define an `Element` in your signature" scenario, both when trying to pass `...attributes` to such a component and when trying to use `...attributes` in such a component's template.

I'm inclined to say that subbing `unknown` for `null` in those error messages is at worst a lateral move on the "how nonsensical is this error message" scale, and that ultimately the right answer is that we should consider adding some special handling for these in our `diagnostics/augmentation` module.

chriskrycho:
Makes sense to me!

PR #472, opened 2022-11-23
Honor lowercase in-scope variables when used as components

### LINKED ISSUE #360: Lowercase in-scope identifiers aren't treated as components
If I write something like this in a template:

```hbs
{{#let (component ...) as |div|}}
  <div>hi</div>
{{/let}}
```

Currently we treat that as being a regular HTML `<div>`, but at runtime it will actually invoke the given component instead.

dfreeman:
Fixes #360


chriskrycho (approved):
😭 but it's what it needs to be for now

PR #479, opened 2022-11-28
Make the Code extension publishable

dfreeman:
The Visual Studio Marketplace and I have had a _day_. While Code is happy to load the unpacked extension in development, I hit a couple of snags when trying to actually publish our 1.0.0-beta.1:
 - Right off the bat, `1.0.0-beta.1` isn't actually a valid version according to the marketplace. Instead, we can publish `1.0.0` packaged with the `--pre-release` flag.
 - `vsce` doesn't understand `.cjs` entrypoints (even though Code itself will [happily use them](https://github.com/microsoft/vscode-vsce/pull/740))
 - After upgrading `vsce` to get a version new enough to understand `--pre-release`, I suddenly began getting a number of "these `node_modules` files have the same case-insensitive paths" errors when packaging, all pointing to identical pairs of paths. Particularly exasperating given I'm on a case-insensitive file system myself.

To address the first two issues, I've dropped the `-beta.1` trailer from the Code package version and converted its contents back to `type: "commonjs"`. For the third, we're now bundling the extension into a single file to avoid packaging any dependencies at all, which is already a best practice but one I'd been putting off.


chriskrycho (commented):
👍🏼

PR #483, opened 2022-11-29
Fix up a couple ESM oddities in the CLI

dfreeman:
This fixes a couple oddities relating to our conversion to ESM in the CLI, namely that `--version` and `--debug-intermediate-representation` were broken.

The former is fixed by reading our own `package.json` and manually specifying the version for `yargs`. The latter is incidentally also fixed by that, as we were using `require` to pull in a few additional modules for setting up the IR emit, and that isn't defined by default in a `type: 'module'` file in Node.


dfreeman:
Oh good, 500 from the npm registry in CI 💀

chriskrycho (approved):

PR #487, opened 2022-11-30
Ensure unions of arg types distribute correctly

### LINKED ISSUE #486: Union types for named args in signatures don't distribute correctly
If I have a helper signature like this:

```ts
interface MyHelperSignature {
  Args: {
    Named: { req: string } | { optional?: number };
  };
  Return: string;
}
```

Then the `MaybeNamed` utility type we use when converting that signature to an `[Invoke]` type ends up dropping the named args entirely, resulting in `() => string`

dfreeman:
Fixes #486, which is the same class of bug we had to spend a bit of time squashing when we first landed native signature support in `@glimmer/component` and friends. This one was a bit more subtle in that it required all named args to be optional in at least one union member.


packages/template/__tests__/component-like.test.ts
```diff
@@ -117,6 +117,29 @@ import TestComponent from './test-component';
...
+
+  expectTypeOf(info).toEqualTypeOf<
+    (
+      ...args:
+        | [full: string, named: NamedArgs<{ force: boolean }>]
+        | [full: string, named?: NamedArgs<Partial<{ foo: string; bar: string }>>]
+        | [first: string, last: string, named: NamedArgs<{ force: boolean }>]
+        | [first: string, last: string, named?: NamedArgs<Partial<{ foo: string; bar: string }>>]
+    ) => ComponentReturn<{}>
+  >();
```
chriskrycho, line 140: 😭

chriskrycho (approved):

PR #490, opened 2022-12-07
Introduce `@glint/type-test` package

dfreeman:
This PR introduces a new `@glint/type-test` package that provides a set of `expect-type`-like expectations for use in templates.

Today, writing type-level tests for Glint types tends to involve using private Glint APIs, both in [our own declaration tests](https://github.com/typed-ember/glint/blob/type-test/packages/environment-ember-loose/__tests__/type-tests/intrinsics/on.test.ts) and in third-party libraries like [`@gavant/glint-template-types`](https://github.com/Gavant/glint-template-types/blob/main/tests/ember-svg-jar/svg-jar-test.ts). These tests are tightly coupled to Glint's internal implementation details, and they also aren't necessarily easy to grok for someone trying to understand what they're testing.

I'll plan on backporting a version of this package to release as 0.9.x as well, so that folks on 0.9 who are using private APIs for testing can migrate to `@glint/type-test` and _then_ move to 1.0, rather than having to update all of their tests to account for internal changes as part of upgrading.


packages/core/__tests__/cli/check.test.ts
```diff
@@ -1,7 +1,7 @@
 import { stripIndent } from 'common-tags';
 import stripAnsi = require('strip-ansi');
 import { describe, beforeEach, afterEach, test, expect } from 'vitest';
-import { Project } from '@glint/test-utils';
+import { Project } from 'glint-monorepo-test-utils';
```
chriskrycho, line 4: 👍🏼 to this.

chriskrycho (commented):
🤯 Nothing like some HKTs-via-dictionaries. 😅

PR #495, opened 2022-12-13
Ensure we handle `@glint-nocheck` in composite projects correctly

dfreeman:
Because TypeScript's `--build` mode is primarily driven off of `emit()` calls to the underlying `Program` instances that make up the project, it's not always guaranteed to pull its diagnostics via the outer `BuilderProgram` diagnostic methods that we override in `patchProgram`.

If we silence diagnostics in those methods and TS subsequently sees the untransformed version of those diagnostics later, it will print those diagnostics for the user. Also troublingly, the presence of those diagnostics in combination with `noEmitOnError` will cause corresponding `.d.ts` files _not_ to be emitted for files that are suppressing template type errors with `@glint-nocheck`.

This change ensures that when we patch the outer program's diagnostic methods, we also patch the _inner_ program's corresponding methods, which is where `emit()` calls draw their diagnostics from. Note that it's not sufficient to _only_ patch the inner program, as the outer one doesn't simply call through, but (for reasons that are beyond me) has its own similar-but-not-the-same implementation using other TS internals.

I also added a call to `ts.sortAndDeduplicateDiagnostics` in locations where we inject our transformation diagnostics to avoid scenarios where they wind up injected twice—this also has the side effect of ensuring that all diagnostics for a given source file are grouped together, which they previously might not have been.

### DIFF
--- a/packages/core/src/cli/perform-build-watch.ts
+++ b/packages/core/src/cli/perform-build-watch.ts
@@ -2,7 +2,7 @@ import type * as TS from 'typescript';
 
 import { buildDiagnosticFormatter } from './diagnostics.js';
 import { sysForCompilerHost } from './utils/sys-for-compiler-host.js';
-import { patchProgram } from './utils/patch-program.js';
+import { patchProgramBuilder } from './utils/patch-program.js';
 import TransformManagerPool from './utils/transform-manager-pool.js';
 
 export function performBuildWatch(
@@ -12,14 +12,11 @@ export function performBuildWatch(
 ): void {
   let transformManagerPool = new TransformManagerPool(ts.sys);
   let formatDiagnostic = buildDiagnosticFormatter(ts);
+  let buildProgram = ts.createEmitAndSemanticDiagnosticsBuilderProgram;
 
   let host = ts.createSolutionBuilderWithWatchHost(
     sysForCompilerHost(ts, transformManagerPool),
-    (...args) => {
-      const program = ts.createEmitAndSemanticDiagnosticsBuilderProgram(...args);
-      patchProgram(program, transformManagerPool);
-      return program;
-    },
+    patchProgramBuilder(ts, transformManagerPool, buildProgram),
     (diagnostic) => console.error(formatDiagnostic(diagnostic))
   );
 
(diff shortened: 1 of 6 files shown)


packages/core/src/cli/perform-build-watch.ts
```diff
@@ -12,14 +12,11 @@ export function performBuildWatch(
...
+  let buildProgram = ts.createEmitAndSemanticDiagnosticsBuilderProgram;
 
   let host = ts.createSolutionBuilderWithWatchHost(
     sysForCompilerHost(ts, transformManagerPool),
-    (...args) => {
-      const program = ts.createEmitAndSemanticDiagnosticsBuilderProgram(...args);
-      patchProgram(program, transformManagerPool);
-      return program;
-    },
+    patchProgramBuilder(ts, transformManagerPool, buildProgram),
```
chriskrycho, line 19: Ah, this is good; I kept meaning to do something of this shape, but never figured out the full thing it needed. Glad it's done here!

packages/core/src/cli/perform-build.ts
```diff
@@ -15,14 +15,11 @@ interface BuildOptions extends TS.BuildOptions {
 export function performBuild(ts: TypeScript, projects: string[], buildOptions: BuildOptions): void {
   let transformManagerPool = new TransformManagerPool(ts.sys);
   let formatDiagnostic = buildDiagnosticFormatter(ts);
+  let buildProgram = ts.createEmitAndSemanticDiagnosticsBuilderProgram;
```
chriskrycho, line 18: In `perform-watch` below, you just inlined this; any reason not to do the same here?

chriskrycho (approved):
😰 heck of a weird issue. Thank you for fixing it, and props to @hmajoros for the 🕵🏼‍♂️ work that not only showed up the original issue but also helped get close enough for you to fix it!

packages/core/src/cli/perform-build.ts
dfreeman: It made Prettier wrap the `patchProgramBuilder` call from 1 line to 5, and I... didn't like that 😅

chriskrycho: Hahahaha I figured it might be that. 👍🏼

PR #498, opened 2022-12-15
Widen supported range for ember-modifier to include v4

SergeAstapov:
with https://github.com/ember-modifier/ember-modifier/releases/tag/v4.0.0, we can't bump `ember-modifier` to v4.0.0 in the project due to `peerDependencies` range including only v3

### DIFF
--- a/packages/environment-ember-loose/-private/dsl/integration-declarations.d.ts
+++ b/packages/environment-ember-loose/-private/dsl/integration-declarations.d.ts
@@ -64,9 +64,9 @@ declare module '@ember/component/helper' {
 //////////////////////////////////////////////////////////////////////
 // Modifiers
 
-import 'ember-modifier/-private/class/modifier';
+import 'ember-modifier';
 
-declare module 'ember-modifier/-private/class/modifier' {
+declare module 'ember-modifier' {
   export default interface ClassBasedModifier<S> extends InstanceType<ModifierLike<S>> {}
 }
 
--- a/packages/environment-ember-loose/__tests__/type-tests/modifier.test.ts
+++ b/packages/environment-ember-loose/__tests__/type-tests/modifier.test.ts
@@ -56,9 +56,6 @@ import type Owner from '@ember/owner';
     'hello'
   );
 
-  type InferSignature<T> = T extends Modifier<infer Signature> ? Signature : never;
-  expectTypeOf<InferSignature<NeatModifier>>().toEqualTypeOf<NeatModifierSignature>();
-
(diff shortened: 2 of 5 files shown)


chriskrycho (approved):
Good catch; thanks!

dfreeman:
Looking at the test output, there may be some compatibility issues with the v4 `ember-modifier` types?

SergeAstapov:
@dfreeman I’ll look into these, there seem to be tests that needs to be updated.

dfreeman:
I didn't think the types had changed either, but _something_ definitely seems to be going wrong in the tests. 

@SergeAstapov as a starting point you may try merging `main` into your branch—we recently fixed some other CI issues there related to the DT migration.

SergeAstapov:
@dfreeman @chriskrycho *I think* figured out where issue comes from, however can't quite think of the way to fix it.

the `typecheck` was failing in `packages/environment-ember-loose/__tests__/type-tests/modifier.test.ts` file.
first issue was that `NeatModifier` was not converted to latest ember-modifier API. Once this was done, another issue popped up.

the following line causes trouble

```ts
let neat = resolve(NeatModifier);
```

because `resolve` function does not match `Modifier` class signature anymore...
As far as I can understand, `ember-modifier@3.2.7` has `/index.d.ts` file with following contents

```
export { default } from './-private/class/modifier';
export { default as modifier } from './-private/function-based/modifier';
// ...
```

and `./-private/class/modifier` has notable lines

```
// ...
export default interface ClassBasedModifier<S = DefaultSignature> extends Opaque<S> {}
export default class ClassBasedModifier<S = DefaultSignature> {
  // ...
}
// ...
```

*However*, in `ember-modifier@4.0.0` we use `rollup` + `rollup-plugin-ts` for bundling, which results in having `index.d.ts` like so

```ts
// ...
interface ClassBasedModifier<S = DefaultSignature> extends Opaque<S> {}
declare class ClassBasedModifier<S = DefaultSignature> {
  // ...
}
export { ClassBasedModifier as default, modifier };
// ...
```

now I'm little puzzled how to make this back on track...

dfreeman:
Thank you for your work on this @SergeAstapov! I probably won't have a chance to dig in further until after the holidays, but I'm happy to pick it up then and see if we can work out a path forward

SergeAstapov:
sounds like a good plan, thank you @dfreeman!
I'll see if I can find the solution myself. if no updates posted in this thread - consider all my attempts failed.

dfreeman:
Thank you, @SergeAstapov!

PR #500, opened 2022-12-15
Ensure adding/removing directives busts any cached typecheck results

hmajoros:
**Fix info from @dfreeman:**

Whether a file contains e.g. a `@glint-nocheck` directive doesn't impact what TypeScript itself sees in that module, only whether we emit errors from it or not. This means that in incremental builds, removing that directive won't produce a visible change from TypeScript's perspective, so we won't emit the now-revealed errors until the next time the `tsbuildinfo` cache is cleared.

This PR just has us emit a comment anywhere we see a `@glint` directive so that the contents of the transformed file will change if a directive comes or goes and TS will reconsider it for typechecking.

**Original writeup of the bug from @hmajoros:**

I believe I have found a bug in `glint --build --watch`, and generally, the same thing happens when running `glint --build` twice in a row.

STR: setup your project with a template ignored via the `{{! @glint-nocheck }}` flag, that would throw a type error in the absence of the nocheck directive. run a `glint --build --watch`, validate that it passes because of the nocheck directive. remove the directive, and notice that the `--build --watch` process does not pick up the fact that there is now a type error from the template.

Same issue can be reproduced running consecutive `glint --build`'s removing the nocheck directive in between runs.

Only way to fix this currently is to run `glint --build --clean` after removing the directive, then re-running `--build`

This PR adds a failing test to reproduce the behavior: after removing the nocheck directive, we would expect the output to show 1 error, instead it shows none.

### DIFF
--- a/packages/core/__tests__/cli/build-watch.test.ts
+++ b/packages/core/__tests__/cli/build-watch.test.ts
@@ -208,6 +208,7 @@ describe('CLI: watched build mode typechecking', () => {
 
       await watch.terminate();
     });
+
     test('reports on errors introduced after removing a glint-nocheck directive', async () => {
       let code = stripIndent`
         import '@glint/environment-ember-template-imports';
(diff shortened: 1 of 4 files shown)


dfreeman:
Thank you for the failing test, @hmajoros! That was super helpful in getting this fixed.

PR #504, opened 2022-12-19
Error when TS is insufficient version

NullVoxPopuli:
I thought that it'd be good to error when TS is insufficient.
Currently, when running `glint` _and_ typescript is < 4.8.0, gts files are passed, untransformed, directly to the type checker -- which felt like a weird behavior.


dfreeman:
Thank you! We should also look into doing something similar in the language server, but that can happen separately.

I'm going to hold on merging this for the moment while we sort through fallout from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63669 (probably need to bump a few dependencies locally), but I _super_ appreciate this fix!

PR #510, opened 2023-01-02
Add support for logical operators (RFC 562)

### LINKED ISSUE #221: Support type-narrowing via discriminated unions

bwbuchanan:
Resolves #221. With this PR merged, you should be able to fully make use of type narrowing for discriminated unions.

Note that the type effects of these operators do not strictly match the semantics specified by RFC 562, in that they express **JavaScript truthiness** instead of **Handlebars truthiness**. It's not clear to me that it is possible to express Handlebars truthiness in TypeScript in such a way that you can usefully narrow types with it in combination with logical operators.

As these helpers have not yet been merged into Ember, you can opt in by adding to your glint config:

```json
{
  "glint": {
    "environment": {
      "ember-loose": {
        "additionalSpecialForms": {
          "globals": {
            "and": "&&",
            "or": "||",
            "not": "!"
          }
        }
      },
      "ember-template-imports": {
        "additionalSpecialForms": {
          "imports": {
            "ember-truth-helpers/helpers/and": { "default": "&&" },
            "ember-truth-helpers/helpers/or": { "default": "||" },
            "ember-truth-helpers/helpers/not": { "default": "!" }
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/packages/core/__tests__/transform/template-to-typescript.test.ts
+++ b/packages/core/__tests__/transform/template-to-typescript.test.ts
@@ -611,6 +611,67 @@ describe('Transform: rewriteTemplate', () => {
         `);
       });
     });
+
+    describe('{{and}}', () => {
+      test('with two arguments', () => {
+        let template = stripIndent`
+        {{log (testAnd 1 2)}}
+        `;
+
+        expect(templateBody(template, { globals: ['testAnd'], specialForms: { testAnd: '&&' } }))
+          .toMatchInlineSnapshot(`
+        "χ.emitContent(χ.resolve(log)((1 && 2)));"
+        `);
+      });
+
+      test('with three arguments', () => {
+        let template = stripIndent`
+        {{log (testAnd 1 2 3)}}
+        `;
+
+        expect(templateBody(template, { globals: ['testAnd'], specialForms: { testAnd: '&&' } }))
+          .toMatchInlineSnapshot(`
+        "χ.emitContent(χ.resolve(log)((1 && 2 && 3)));"
+        `);
+      });
+    });
+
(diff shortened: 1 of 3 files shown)


dfreeman (approved):
Looks great, thank you for your hard work here @bwbuchanan!

boris-petrov:
Do we get a new release really soon with that change, please, @dfreeman? :smile:

dfreeman:
@boris-petrov I'm working my way through a big backlog of both OSS and work stuff, but aiming to do a release early next week

PR #520, opened 2023-01-17
Honor user formatting preferences in completions

chadhietala:
This requires #499

In this PR we thread through the `ConfigManager` through the vscode binding so that completions can utilize the formatting and preferences of the workspace if they are defined. For calling the `GlintLanguageServer` programmatically formatting and preferences can be passed but default to `{}` instead of being hardcoded at the call site.


dfreeman:
Thank you!

PR #544, opened 2023-03-14
Expose TS language service and original document contents in `ProjectAnalysis`

cafreeman:
This PR updates `@glint/core` to expose a few more pieces of the language server implementation that should make it easier for Glint to be used programmatically (e.g. in https://github.com/rehearsal-js/rehearsal-js).

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -10,7 +10,6 @@ export interface ProjectAnalysis {
   glintConfig: GlintConfig;
   transformManager: TransformManager;
   languageServer: GlintLanguageServer;
-  documents: DocumentCache;
   service: ts.LanguageService;
   shutdown: () => void;
 }
@@ -39,10 +38,9 @@ export function analyzeProject(projectDirectory: string = process.cwd()): Projec
     glintConfig,
     transformManager,
     languageServer,
-    documents,
     service: languageServer.service,
     shutdown,
   };
 }
 
-export type { DocumentCache, TransformManager, GlintConfig, GlintLanguageServer };
+export type { TransformManager, GlintConfig, GlintLanguageServer };
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -45,4 +45,4 @@ export function analyzeProject(projectDirectory: string = process.cwd()): Projec
   };
 }
 
-export { DocumentCache, TransformManager, GlintConfig, loadConfig, GlintLanguageServer };
+export type { DocumentCache, TransformManager, GlintConfig, GlintLanguageServer };
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -3,12 +3,15 @@ import DocumentCache from './common/document-cache.js';
 import TransformManager from './common/transform-manager.js';
 import GlintLanguageServer from './language-server/glint-language-server.js';
 import * as utils from './language-server/util/index.js';
+import * as ts from 'typescript';
 
 /** @internal */
 export interface ProjectAnalysis {
   glintConfig: GlintConfig;
   transformManager: TransformManager;
   languageServer: GlintLanguageServer;
+  documents: DocumentCache;
+  service: ts.LanguageService;
   shutdown: () => void;
 }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/language-server/glint-language-server.ts
```diff
@@ -48,7 +48,7 @@ interface TransformedOffsets {
 }
 
 export default class GlintLanguageServer {
-  private service: ts.LanguageService;
+  readonly service: ts.LanguageService;
```
cafreeman, line 51: We make this readonly instead of private so that we can pull it off the language server and return it as part of the `ProjectAnalysis`

packages/core/src/index.ts
```diff
@@ -32,5 +35,14 @@ export function analyzeProject(projectDirectory: string = process.cwd()): Projec
...
+    glintConfig,
+    transformManager,
+    languageServer,
+    documents,
+    service: languageServer.service,
+    shutdown,
+  };
 }
+
+export { DocumentCache, TransformManager, GlintConfig, loadConfig, GlintLanguageServer };
```
chriskrycho, line 48: I will definitely defer to @dfreeman here on whether we want the `loadConfig` exported (I could see it being fine, but I also have an intuitive HMMMMM response); but we *definitely* want to go out of our way here to make sure the others are `export type` rather than direct `export`; they should absolutely never be constructed directly by end users.

packages/core/src/index.ts
```diff
@@ -32,5 +35,14 @@ export function analyzeProject(projectDirectory: string = process.cwd()): Projec
   let languageServer = new GlintLanguageServer(glintConfig, documents, transformManager);
   let shutdown = (): void => languageServer.dispose();
 
-  return { glintConfig, transformManager, languageServer, shutdown };
+  return {
+    glintConfig,
+    transformManager,
+    languageServer,
+    documents,
```
chriskrycho, line 42: The `documents` export here gives me the heebie-jeebies. It's very much an internal detail of how the system works, and not (I think?) a thing we want to expose as something to generally support. Maybe a better approach here would be to define what it is a consumer *needs* that the `DocumentCache` provides a way of getting, and provide a function here which can answer that question (but in a way we can evolve over time)? 🤔

cafreeman: That's a fair concern, for sure.

IME we really only need `getDocumentVersion`, `getDocumentContents`, and `updateDocument`, which could easily just be three methods on a POJO that gets returned from `analyzeProject`. Would that work for you?

chriskrycho: That definitely seems *more* reasonable! Can you expand a bit on what each of those is used for? 🙏🏼

cafreeman: Of course!

All three of these methods are used to satisfy parts of the TypeScript `LanguageServerHost` interface (e.g. `getScriptSnapshot`).

`getDocumentContents` and `updateDocument` are both used to (as you might guess) interact with documents that Glint has stored. This comes in handy since we are essentially using Glint to analyze and (sometimes) fix issues found throughout a codebase while converting it to TypeScript, so the ability to, for example, pull a document's contents out of Glint,  apply a codefix to it, and then write it back to the document cache comes in very handy.

We don't actually use `getDocumentVersion` directly, but it may get called by other parts of the TS  Compiler API as part of the language host.

chriskrycho: Makes sense to me; I'll let @dfreeman confirm, though!

dfreeman: I'm a little confused about the usage pattern here—from the fact that you're needing to construct a `LanguageServiceHost`, it sounds like you're spinning up a second `LanguageService` in _addition_ to the one Glint instantiates when you run `analyzeProject`?

packages/core/src/index.ts
```diff
@@ -3,12 +3,14 @@ import DocumentCache from './common/document-cache.js';
 import TransformManager from './common/transform-manager.js';
 import GlintLanguageServer from './language-server/glint-language-server.js';
 import * as utils from './language-server/util/index.js';
+import * as ts from 'typescript';
```
dfreeman, line 6: Can we either:
 - make this a type-only import to avoid accidentally pulling in TS at runtime with some seemingly-innocuous change later?
 - just let consumers access `languageServer.service` now that it's non-`private` and leave `ProjectAnalysis` as-is?

dfreeman (commented):
One quibble about making an import type-only (or dropping it entirely) — otherwise looks good!

packages/core/src/index.ts
cafreeman: Good call, I just removed service from `ProjectAnalysis` and dropped the ts import.

dfreeman (approved):

PR #558, opened 2023-04-22
Allow `each-in` to accept `key` as a named arg

### LINKED ISSUE #541: Support `key` option in `each-in` helper
The `key` option in the `{{#each}}` and `{{#each-in}}` helpers are useful for improved rendering of elements generated by arrays and objects. Currently, glint supports `key` the `{{#each}}` helper, but it is not supported for `{{#each-in}}`. For example

```js
get object() {
  return {
    first: [1, 2, 3],
    second: [4, 5, 6],
    third: [7, 8, 9]
  }
}
```
```handlebars
{{#each-in this.object key='@key' as |key value|}}
  <div>{{key}}: {{value}}</div>
{{/each}}
```
produces this typescript error:
```
Argument of type '{ key: string; }' is not assignable to parameter of type 'EmptyObject'. Object literal may only specify known properties, and 'key' does not exist in type 'EmptyObject'.
```

The `key='@key'` syntax is supported by [glimmer](https://github.com/glimmerjs/glimmer-vm/blob/4f1bef0d9a8a3c3ebd934c5b6e09de4c5f6e4468/packages/%40glimmer/reference/lib/iterable.ts#L57-L58) and is mentioned by this [bugfix](https://github.com/emberjs/ember.js/pull/16399) (admittedly, there's not a whole lot of documentation of the `key` option for the `{{#each-in}}` helper.)

LanceStasinski:
fixes #541

### DIFF
--- a/packages/environment-ember-loose/-private/intrinsics/each-in.d.ts
+++ b/packages/environment-ember-loose/-private/intrinsics/each-in.d.ts
@@ -2,7 +2,10 @@ import { ComponentLike } from '@glint/template';
 
 export type EachInKeyword = abstract new <T>() => InstanceType<
   ComponentLike<{
-    Args: { Positional: [object: T] };
+    Args: {
+      Positional: [object: T]
+      Named: { key?: string }
+    };
     Blocks: {
       default: [key: EachInKey<T>, value: Exclude<T, null | undefined>[EachInKey<T>]];
       else?: [];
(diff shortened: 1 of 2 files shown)


dfreeman:
Thanks @LanceStasinski! The change itself looks good, just a couple of complaints from Prettier to knock out and then we can get this merged

dfreeman:
Thank you!

LanceStasinski:
> Thank you!

@dfreeman, you're welcome!

PR #583, opened 2023-05-26
Add missing argument to LinkTo component type

bitwolfe:
The `replace` argument was missing from the Glint types.
  https://github.com/emberjs/ember.js/blob/v4.12.0/packages/@ember/-internals/glimmer/lib/components/link-to.ts#L468


chriskrycho (approved):
Thanks! And even more importantly than that implementation detail, [it is documented here](https://guides.emberjs.com/v4.12.0/routing/linking-between-routes/#toc_replacing-history-entries).

dfreeman:
Thanks for tracking down the documentation, @chriskrycho!

PR #586, opened 2023-06-05
support template processing customization

### LINKED ISSUE #585: allow customizing template processing in custom environments for hbs files
i'm trying to create a custom env for [ember-hbs-imports](https://github.com/patricklx/ember-hbs-imports)
which was working in a hacky way until 0.9.7. see https://github.com/patricklx/ember-hbs-imports/blob/main/glint/glint-hbs-imports-env.js
I just need to be able to preprocess the template contents, or even better, transform the ast. afterwards I can probably add a transform for the ts source part to add setEmitMetadata
and I would need to be able to pass globals...

not sure if there is a better way?

patricklx:
resolves #585

### DIFF
--- a/packages/core/src/config/environment.ts
+++ b/packages/core/src/config/environment.ts
@@ -12,8 +12,6 @@ import {
   PathCandidateWithDeferral,
   SourceKind,
 } from '@glint/core/config-types';
-import TransformManager from '../common/transform-manager.js';
-import * as util from '../language-server/util/index.js';
 
 const require = createRequire(import.meta.url);
 
--- a/packages/core/src/config/environment.ts
+++ b/packages/core/src/config/environment.ts
@@ -12,6 +12,8 @@ import {
   PathCandidateWithDeferral,
   SourceKind,
 } from '@glint/core/config-types';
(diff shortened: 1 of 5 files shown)


chriskrycho:
Hi @patricklx – thanks for the work here, but I am closing this in line with my comment on #585:

> Glint does not have the maintainer time commitment to support multiple syntaxes. Given [Ember RFC 779](https://github.com/emberjs/rfcs/pull/779), we are committed to supporting loose mode templates for `.hbs` and strict mode templates in `.gjs`/`.gts` with `<template>`, and we will absolutely not support any other modes. I understand that might be disappointing, but anything else would be dishonest about what we can or cannot support!
> 
> Additionally, given Ember’s own direction, I would strongly recommend that you deprecate `ember-hbs-imports` and move to the form the community will be standardizing on.

Going forward, we do not intend to support environments adding custom behavior/processing as we did in the past: we had some minimal support for that capability because it was *necessary* while the community worked out the path to a shared design for strict mode, but now that we've landed on `<template>` for Glimmer/Ember as a whole, we do not expect to support any other authoring modes.

Issue #579, opened 2023-05-18
Expose loadConfig to consumers


egorio:
I'm working on a tool to migrate JS to TS, including Glint project. The tool needs access to glint config values from `tsconfig.json` file and team [is implementing a function](https://github.com/rehearsal-js/rehearsal-js/blob/master/packages/utils/src/cli.ts#L831) to do this, but there are few challenges there to make it work right.

As you know, the standard TS functions to parse config ignore `glint` option, so you can't use them.
It's required to implement a function to search for this value recursively across all tsconfig hierarchy and merge all found glint values.

Glint has its [own function to load the config](https://github.com/typed-ember/glint/blob/main/packages/core/src/config/loader.ts#L64-L83) to do this as well.

The basic implementation of the function to load config is not complex, but if we want to have it typed and validated then we need to "reimplement" the [glint/packages/core/src/config/](https://github.com/typed-ember/glint/blob/main/packages/core/src/config/index.ts#L14-L21) which sounds counter-productive while glint has it already.

It's also a good practice to provide such functionality for open-source community, for example [prettier's resolvaConfig](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options) or [eslint's calculateConfigForFile](https://eslint.org/docs/latest/integrate/nodejs-api#-eslintcalculateconfigforfilefilepath).

So the question is could Glint has loadCondig and it's types (already [exposed](https://github.com/typed-ember/glint/blob/main/packages/core/src/index.ts#L43)) to be exported to end user?

egorio:
I can come with PR if Glint team will agree to export it.

dfreeman:
> It's also a good practice to provide such functionality for open-source community, for example [prettier's resolvaConfig](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options) or [eslint's calculateConfigForFile](https://eslint.org/docs/latest/integrate/nodejs-api#-eslintcalculateconfigforfilefilepath).

As a reminder, none of the programmatic interfaces exposed from `@glint/core` are officially public, and while we'll make a good-faith effort not to break those interfaces unnecessarily, they're intentionally undocumented and not a part of the semver contract for the package. Given that, comparing to ESLint and Prettier's public APIs isn't necessarily the best point of reference for "good practice" 😉 

[...]

PR #606, opened 2023-07-31
Expose loadConfig

egorio:
Expose loadConfig for consumers

> Since we already provide access to the GlintConfig instance to consumers who run analyzeProject, it seems reasonable to offer a way to load that config without forcing a full project analysis.

Closes https://github.com/typed-ember/glint/issues/579


egorio:
@dfreeman

chriskrycho (approved):

chriskrycho:
Thanks!

Issue #610, opened 2023-08-08
Problem with generics in Signatures


gossi:
I'm providing glint support for `ember-element-helper` tildeio/ember-element-helper#107

Here are the types based on the ones from Dan (in tildeio/ember-element-helper#102):

```ts
export type ElementFromTagName<T extends string> = T extends keyof HTMLElementTagNameMap
  ? HTMLElementTagNameMap[T]
  : Element;

type Positional<T extends string> = [name: T];
type Return<T extends string> = typeof EmberComponent<{
  Element: ElementFromTagName<T>;
  Blocks: { default: [] };
}>;

export interface ElementSignature<T extends string> {
  Args: {
    Positional: Positional<T>;
  };
  Return: Return<T> | undefined;
}

export default class ElementHelper<T extends string> extends Helper<ElementSignature<T>> {}
```

Using the helper works straight away:

```
<template>
  {{#let (element 'div') as |Tag|}}
    <Tag id="lalala" ...attributes>Hello there!</Tag>
  {{/let}}
</template>
```

Going a bit more dynamic and allowing a `@tag` to be passed in, works _when_ the type is made explicit:

```ts
import { type ElementSignature } from 'ember-element-helper';

interface ElementReceiverSignature{
  Element: HTMLDivElement; // 1: explicit element here
  Args: {
    tag: ElementSignature<'div'>['Return']; // 2: explicit tag name here
  };
  Blocks: {
    default: [];
  };
}

export default class ElementReceiver extends Component<ElementReceiverSignature> {
  <template>
    {{#let @tag as |Tag|}}
      <Tag id="content" ...attributes>{{yield}}</Tag>
    {{/let}}
  </template>
}
```

... of course those explicit types do not make sense, when we want to have any element being passed in. Making it generic makes the problem visible:

```ts
import {
  type ElementFromTagName,
  type ElementSignature
} from 'ember-element-helper';

interface ElementReceiverSignature<T extends string> {
  Element: ElementFromTagName<T>;
  Args: {
    tag: ElementSignature<T>['Return'];
  };
  Blocks: {
    default: [];
  };
}

export default class ElementReceiver<T extends string> extends Component<
  ElementReceiverSignature<T>
> {
  <template>
    {{#let @tag as |Tag|}}
      <Tag id="content" ...attributes>{{yield}}</Tag>
    {{/let}}
  </template>
}
```

This reveals two locations where glint throws both times the same error:

1. `<Tag` and
2. `...attributes`

[...]

PR #613, opened 2023-08-17
Allow generics in Signatures

gossi:
Potential fix to #610

I was applying the fix as part of this PR locally on my PR tildeio/ember-element-helper#107 and see if in my type-test-component:

```ts
import type { ElementSignature, ElementFromTagName } from 'ember-element-helper';

interface ElementReceiverSignature<T extends string = 'article'> {
  Element: ElementFromTagName<T>;
  Args: {
    tag: ElementSignature<T>['Return'];
  };
  Blocks: {
    default: [];
  }
}

export default class ElementReceiver<T extends string> extends Component<ElementReceiverSignature<T>> {}
```

```hbs
{{#let @tag as |Tag|}}
  {{!@glint-ignore}}
  <Tag id="content" ...attributes>{{yield}}</Tag>
{{/let}}
```

I was able to remove the line `{{!@glint-ignore}}` and if the type linting still passes, which is the case.

So, yes this is a potential fix, but I share the same concerns as @dfreeman this might come with side-effects.

So I'll make it a draft PR.

### DIFF
--- a/packages/template/-private/signature.d.ts
+++ b/packages/template/-private/signature.d.ts
@@ -45,7 +45,7 @@ export type ComponentSignatureBlocks<S> = S extends { Blocks: infer Blocks }
 
 /** Given a component signature `S`, get back the `Element` type. */
 export type ComponentSignatureElement<S> = S extends { Element: infer Element }
-  ? NonNullable<Element> extends never
+  ? Element extends null
     ? unknown
     : Element
   : unknown;


chriskrycho:
Thanks for opening this! The fact that CI is ✅ is a great initial sign – one of us will try to review in detail shortly.

dfreeman:
This should include, at a minimum, a new test case for the scenario this is intended to fix. Otherwise there's nothing to prevent us from flipping this back in the future and causing the same problem again

(I'd also recommend coming up with a more specific title for this PR, since that's what will go in the release notes—we clearly already "allow generics in signatures" in the general case, since many built-ins like `let` and `each` require that 😉)

packages/template/-private/signature.d.ts
```diff
@@ -45,7 +45,8 @@ export type ComponentSignatureBlocks<S> = S extends { Blocks: infer Blocks }
 
 /** Given a component signature `S`, get back the `Element` type. */
 export type ComponentSignatureElement<S> = S extends { Element: infer Element }
-  ? NonNullable<Element> extends never
+  ? NonNullable<Element> extends never 
+// ? Element extends null
```
gossi, line 49: this is a temporary revert. That is to write failing tests, which when switching back to the fix should be resolved.

packages/template/__tests__/signature.test.ts
```diff
@@ -153,3 +154,56 @@ interface FullLongSig {
...
+}
+
+expectTypeOf<ComponentSignatureArgs<ElementReceiverSignature<'div'>>>().toEqualTypeOf<{
+  Named: {
+    element: ElementSignature<'div'>['Return'];
+  };
+  Positional: []
+}>();
+
+expectTypeOf<ComponentSignatureArgs<ElementReceiverSignature<null>>>().toEqualTypeOf<{
```
gossi, line 194: Actually, this line is the problematic part here, I'd say. For what I think needs to be tested, is how the components are invoked, rather than the static definition?

That is, in invocation, this should be `null`. But this is obviously a failure:

> `Type 'null' does not satisfy the constraint 'string'.`

To my knowledge, to properly test this, is to provide _broken_ types. How to do that?

gossi:
I added some tests. I copied over all the relevant types from `(element)` helper as well as a component receiving it. Feels a lot, but makes it accurate. If this can be less types, please let me know.

I also left comments where I got stuck. I think, I need to test broken/invalid types, in a environment with correct types?

wozny1989:
Hi @gossi! Just wondering if you’re planning to work on this at some point.

gossi:
I stopped. Iirc back then too much red showed up - and I didn't wanted to go through all this as it didn't seem much related to what I changed. However since then a lot changed on the glint repo itself. Feel free to pick it up.

Issue #609, opened 2023-08-03
Use `content-tag` in `@glint/environment-ember-template-imports`


NullVoxPopuli:
ember-template-imports is an implementation detail of a feature, where those implementation details are being phased out.
For example, `ember-template-imports` is not used within the v2-addon blueprint (with rollup -- there is a rust plugin that handles the transform, per RFC: https://github.com/emberjs/rfcs/pull/933), and `@embroider/vite` uses that same plugin in apps, and actually _breaks_ with `ember-template-imports` present (one too many transforms are happening).

So, I think Glint should _also_ use https://www.npmjs.com/package/content-tag 

~~However, because how Glint handles `<template>` is an implementation details, there is another approach as well. There is a javascript implementation of content-tag used in this PR: 
https://github.com/ember-cli/eslint-plugin-ember/pull/1920/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519~~

NullVoxPopuli:
the JS one, it seems is more focused on having consistent line outputs and other meta, whereas the rust-based one is purely concerned with the function of the transform. Idk if the rust-based one could/should also emit that meta, but I asked about it in the eslint-plugin-ember PR, and it seems different needs were... _needed_. :upside_down_face:

patricklx:
The js one (ember-template-tag) has 3 main functions.
transformForLint, transform, parse.
Parse just parses the code and returns meta information about the detected templates.
I guess that should be enough?

NullVoxPopuli:
@patricklx can it do error-tolerant parsing? one thing Glint is lacking right now is the ability to suggest args / blocks / etc _while_ typing, when at a given mid-type the overall template is invalid syntax

patricklx:
@NullVoxPopuli it's based on babel and it enables error recovery by default. But if you mean template parts, then that has nothing to do with the template detection, but an issue in glimmer parsing, as glimmer does not allow mustache paths ending with a `.`

NullVoxPopuli:
Unblocks gts in Polaris: https://github.com/NullVoxPopuli/polaris-starter/pull/11

PR #615, opened 2023-08-29
Use content tag

NullVoxPopuli:
Resolves: https://github.com/typed-ember/glint/issues/609
~~Blocked by: https://github.com/embroider-build/content-tag/issues/19~~
~~Blocked by: https://github.com/embroider-build/content-tag/pull/31~~

Unblocks: https://github.com/NullVoxPopuli/polaris-starter/pull/11

Problem:
- content-tag requires syntax to be correct
- LSPs need fault-tolerant parsing to aid in intellisense, suggestions and all that

This isn't a new problem for Glint, but it does take things in the wrong direction a bit.

### DIFF
--- a/packages/environment-ember-template-imports/-private/environment/preprocess.ts
+++ b/packages/environment-ember-template-imports/-private/environment/preprocess.ts
@@ -8,8 +8,6 @@ import { GLOBAL_TAG, PreprocessData, TemplateLocation } from './common';
 const TEMPLATE_START = `[${GLOBAL_TAG}\``;
 const TEMPLATE_END = '`]';
 
-// NOTE: This import is a lie -- We are not real ESM, and this is compiled to CJS.
-//       ESM usage would be different here, and this is the CJS version of content-tag.
 import { Preprocessor } from 'content-tag';
 const p = new Preprocessor();
 
--- a/packages/environment-ember-template-imports/-private/environment/preprocess.ts
+++ b/packages/environment-ember-template-imports/-private/environment/preprocess.ts
@@ -1,14 +1,42 @@
 import { GlintExtensionPreprocess } from '@glint/core/config-types';
-import { parseTemplates } from 'ember-template-imports';
 import { GLOBAL_TAG, PreprocessData, TemplateLocation } from './common';
-
+/**
+ * We don't need to actually generate the code that would be emitted for
+ * a real app to build, we can use placeholders, so we have less offsets
+ * to worry about (precompile, setComponentTemplate, their imports, etc)
+ */
 const TEMPLATE_START = `[${GLOBAL_TAG}\``;
 const TEMPLATE_END = '`]';
 
+// NOTE: This import is a lie -- We are not real ESM, and this is compiled to CJS.
+//       ESM usage would be different here, and this is the CJS version of content-tag.
+import { Preprocessor } from 'content-tag';
+const p = new Preprocessor();
+
+interface Parsed {
+  type: 'expression' | 'class-member';
+  tagName: 'template';
+  contents: string;
+  range: {
+    start: number;
+    end: number;
+  };
+  contentRange: {
+    start: number;
+    end: number;
+  };
+  startRange: {
+    end: number;
+    start: number;
+  };
+  endRange: {
+    start: number;
+    end: number;
+  };
+}
+
 export const preprocess: GlintExtensionPreprocess<PreprocessData> = (source, path) => {
-  let templates = parseTemplates(source, path, { templateTag: 'template' }).filter(
-    (match) => match.type === 'template-tag'
-  );
+  let templates = p.parse(source, path) as Parsed[];
 
   let templateLocations: Array<TemplateLocation> = [];
   let segments: Array<string> = [];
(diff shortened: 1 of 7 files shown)


packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,12 +1,19 @@
 import { GlintExtensionPreprocess } from '@glint/core/config-types';
```
NullVoxPopuli, line 1: ignore everything in here until content-tag has `parseTemplates`

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,26 +1,86 @@
...
+/**
+ * We don't need to actually generate the code that would be emitted for
+ * a real app to build, we can use placeholders, so we have less offsets
+ * to worry about (precompile, setComponentTemplate, their imports, etc)
+ */
 const TEMPLATE_START = `[${GLOBAL_TAG}\``;
 const TEMPLATE_END = '`]';
 
+// NOTE: This import is a lie -- We are not real ESM, and this is compiled to CJS.
+//       ESM usage would be different here, and this is the CJS version of content-tag.
```
dfreeman, line 12: Where's the lie here? Wouldn't this import operate the same under ESM?

In the node repl, for instance, I see this:
```
> (await import('content-tag')).Preprocessor
[class Preprocessor]
```

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,26 +1,86 @@
...
+import { Preprocessor } from 'content-tag';
+const p = new Preprocessor();
+
+interface Parsed {
```
dfreeman, line 16: Can we have `content-tag` expose this type itself, either by generating it from the underlying Rust struct(s) or something like [`#[wasm_bindgen(typescript_custom_section)]`](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-rust-exports/typescript_custom_section.html)?

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,26 +1,86 @@
...
+
 export const preprocess: GlintExtensionPreprocess<PreprocessData> = (source, path) => {
-  let templates = parseTemplates(source, path, { templateTag: 'template' }).filter(
-    (match) => match.type === 'template-tag'
-  );
+  // NOTE: the preprocessor requires fully syntactically correct code
+  //       which isn't appropriate for an LSP, so
+  //       1. we need to catch parse errors
+  //       2. we need to move to something else, so we can have
+  //          intellisense on in-progress-typed code (such as `@`)
```
dfreeman, line 43: It sounds like this comment isn't entirely accurate (though catching the error this way seems like the right thing to do regardless)

dfreeman (commented):
> - content-tag requires syntax to be correct
> - LSPs need fault-tolerant parsing to aid in intellisense, suggestions and all that
>
> This isn't a new problem for Glint, but it does take things in the wrong direction a bit.

Based on Ed's research yesterday, it sounds like this isn't quite true, which is great news!

packages/environment-ember-template-imports/-private/environment/preprocess.ts
NullVoxPopuli: imprecise language use on my part, but the content-tag code is cjs
- https://www.npmjs.com/package/content-tag?activeTab=code

and more precisely, _node_ cjs.

Which is an important distinction for this effort: https://github.com/embroider-build/content-tag/pull/18
which has more to do with balancing the tradeoff between standalone ESM package with no external node dependencies than the concept of ESM vs CJS in general.

packages/environment-ember-template-imports/-private/environment/preprocess.ts
NullVoxPopuli: I think so -- tho I don't know how specific we'd get. @ef4 said an update to one of the build tools would give us better types when updated.

It's also totally reasonable for us to ship our own d.ts, and ignore the generated one.

packages/environment-ember-template-imports/-private/environment/preprocess.ts
dfreeman: The import as it's currently written would work regardless of whether this module were transpiled to CJS or ESM, right? And if `content-tag` were to change that, I'd expect it to be as part of a major version bump, so I'm not sure as a reader what this comment is meant to be informing me of. What value is it adding?

packages/environment-ember-template-imports/-private/environment/preprocess.ts
dfreeman: I'm not too hung up on any specific approach on the `content-tag` side; it just makes me nervous to be encoding assumptions here that could break at any time 😅

packages/environment-ember-template-imports/-private/environment/preprocess.ts
NullVoxPopuli: idk, my assumption has been that content-tag only supported cjs, so adding an esm (and more specifically, non-node) entrypoint would be non-breaking.

will discuss with others

packages/environment-ember-template-imports/-private/environment/preprocess.ts
NullVoxPopuli: removed :upside_down_face:

NullVoxPopuli:
Observations:
## Deleting the closing `</template>`
### Example 1
kinda goofy
image

### Example 2
this actually makes sense and provides meaningful feedback about the missing closing `</template>`, which I believe is an improvement?
image

## Deleting the opening `<template>`

### Example 1

sqigglies everywhere.
maybe this is something we can get our swc parser to help with though?

image

## Deliberate syntax error

### "importing a number"

image

### extra quote

image

### number * string

image

### forgetting to give a variable a value

image

### accidentally discovered function noop shorthand

this is actually valid, but looks goofy

image

NullVoxPopuli:
I broke something here, so I'm starting over here: https://github.com/typed-ember/glint/pull/655

PR #616, opened 2023-08-31
Fix "Helper and Modifier Signatures" code example

SergeAstapov:
just a typo


chriskrycho (changes requested):
Ah, thank you! Let's actually go the other way, and use `values` in the body instead.

chriskrycho (approved):

chriskrycho:
Thank you!

PR #620, opened 2023-09-19
Support import paths with .gts extensions

### LINKED ISSUE #618: Support verbatimModuleSyntax, allowing importing of gts files
I also opened this here: https://github.com/embroider-build/addon-blueprint/issues/194

But I have a repro here: https://github.com/universal-ember/ember-primitives/pull/114

error looks like this:
```
[js] created dist in 738ms
[js] npm run build:js exited with code 0
[types] src/components/portal.gts:4:49 - error TS2307: Cannot find module './portal-targets.gts' or its corresponding type declarations.
[types] 
[types] 4 import { findNearestTarget, type TARGETS } from './portal-targets.gts';
[types]                                                   ~~~~~~~~~~~~~~~~~~~~~~
[types] 
```

Note that `gts` extensions were used in the import path to avoid needing `@rollup/plugin-node-resolve`

lukemelia:
Note that using imports with extensions requires TypeScript 5 and enabling the setting to support importing files with typescript extensions ("allowImportingTsExtensions").

Fixes #618


package.json
```diff
@@ -28,7 +28,7 @@
     "eslint": "^8.27.0",
     "prettier": "^2.1.1",
     "release-it": "^15.5.0",
-    "typescript": "~4.8.0"
+    "typescript": "~5.2.2"
```
NullVoxPopuli, line 31: why upgrade TS? 
it looks like it broke the lints :sweat_smile:

lukemelia: The setting to support importing files with typescript extensions ("allowImportingTsExtensions") was added in Typescript 5. Perhaps it would be better to leave the TS version at 4.8 anyway? This is the feedback I'm hoping for on this PR.

NullVoxPopuli: is it possible to support `allowImportingTsExtensions` _and_ TS 4.8?
Like, have some version check somewhere?

I think the only way Glint could have a breaking change in `@glint/core` is if it broke lockstep with all the other packages -- introducing a breaking change in the other packages would have an obnoxious impact on the ecosystem, imo

lukemelia: The other changes only kick in if you have an import with a .gts extension, so I guess it is probably OK to not require 4.8.0 here but document that you need TS 5 if you want to import with extensions?

lukemelia: With Typescript 4.8.0. there is a failing test:

```
src/index.gts:1:1 - error TS2691: An import path cannot end with a '.ts' extension. Consider importing './Greeting.js' instead.
```

I think we need TS5 and `allowImportingTsExtensions: true` in order for this to pass. I'll look into whether I can use TS5 for just that test.

NullVoxPopuli: we can also conditionally skip the test, and be sure to add TS 4.8 to the test matrix, as well as TS 5.x

lukemelia: I think I figured it out. Let's see how this CI run does.

lukemelia: 🤣 failed comically quickly! I'll pick this up again later.

lukemelia: I do believe this is resolved now

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: 16
```
NullVoxPopuli, line 31: why did node change? was node 14 supported?

packages/core/src/common/document-cache.ts
```diff
@@ -58,8 +58,15 @@ export default class DocumentCache {
     );
   }
 
+  private gtsAwareExtName(filename: string): string {
+    if (filename.endsWith('.gts.ts')) {
```
NullVoxPopuli, line 62: two extensions? 
:upside_down_face:

packages/core/src/transform/template/import-declaration-span.ts
```diff
@@ -0,0 +1,53 @@
...
+    let updatedNode = ts.factory.updateImportDeclaration(
+      node,
+      node.modifiers,
+      node.importClause,
+      newModuleSpecifier,
+      node.assertClause
+    );
+    let originalStart = node.getStart();
+    let originalLength = node.getEnd() - originalStart;
+    const printer = ts.createPrinter();
```
NullVoxPopuli, line 31: do we need to create a new printer each time this function is called?

test-packages/ts-template-imports-app/app/components/Greeting.gts
```diff
@@ -13,10 +14,12 @@ export default class Greeting extends Component<GreetingSignature> {
...
 
-function repeat(value: string, times: number): string {
-  return Array(times).fill(value).join('');
-}
-
 const Bang: TOC<{ Args: { times: number } }> = <template>
   {{repeat '!' @times}}
 </template>
+
+declare module '@glint/environment-ember-loose/registry' {
```
NullVoxPopuli, line 21: why was this added? it's only needed for loose mode

test-packages/ts-template-imports-app/app/helpers/repeat.ts
```diff
@@ -0,0 +1,9 @@
+import { helper } from '@ember/component/helper';
+
+function repeat(params: [string, number] /*, hash*/): string {
+  return params[0].repeat(params[1]);
+}
+
+const repeatHelper = helper(repeat);
```
NullVoxPopuli, line 7: why wrap in `helper`?

As of ember-source@4.5, we don't need helper, and can "just" export the functions (and change the signature to be a normal function as it once before it was moved in to this file)

test-packages/ts-template-imports-app/app/components/Greeting.gts
NullVoxPopuli: https://github.com/typed-ember/glint/pull/620/files#r1331742697

test-packages/ts-template-imports-app/package.json
```diff
@@ -4,7 +4,17 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "test": "glint"
+    "test": "npm-run-all test:*",
+    "test:typecheck": "glint",
```
NullVoxPopuli, line 8: why run both?

NullVoxPopuli (commented):
left some questions. thanks for working on this! this is a massive help!!

packages/core/src/common/document-cache.ts
lukemelia: This is the extension that TypeScript tries if you import a .gts module since it has no intrinsic understanding of the `.gts` extension. We're intercepting it to map it to the actual file.

test-packages/ts-template-imports-app/app/components/Greeting.gts
lukemelia: I thought it would be a good idea to incorporate loose mode to make sure we have decent coverage for this scenario since in today's apps using .gts files, they will still have route templates.

test-packages/ts-template-imports-app/app/helpers/repeat.ts
lukemelia: 👍  Will fix.

test-packages/ts-template-imports-app/package.json
lukemelia: 👍  Will fix.

lukemelia: Actually, I did this because the other test apps are running both of these.

lukemelia:
Updates made! Thanks for the feedback @NullVoxPopuli

packages/core/src/common/document-cache.ts
ef4: This is worth a comment in the source because it will definitely look weird to future readers.

ef4 (approved):
Thanks, this looks good to me.

dfreeman (changes requested):
Sorry, I've been on vacation and disconnected this week, and I'm catching up on my phone on the train back now.

I'll dive deeper when I have computer access again, but please hold on merging this. I don't think the transform layer is likely the right place to be solving this, which would explain why it's been so painful.

dfreeman:
I've opened a PR with an alternative approach to this problem in #621

@lukemelia I want to apologize for the miscommunication on this and the time/effort it's cost you—it's 100% my fault. I'd had a separate conversation in Discord about what would need to happen to fix this, and then on #618 I got sidetracked by some of the meta-questions being asked and didn't focus properly on the fix you were suggesting there to realize that you hadn't been a part of or seen that original conversation.

Thank you very much for your work on this, and I'm sorry for pointing you in the wrong direction!

lukemelia:
No worries, @dfreeman. Thanks for resetting us onto the right path.

PR #621, opened 2023-09-24
Support custom extensions with `allowImportingTsExtensions`

### LINKED ISSUE #618: Support verbatimModuleSyntax, allowing importing of gts files

dfreeman:
This is an alternative approach to the one taken in #620, and fixes #618.

#620 attempts to support import paths with an explicit `.gts` extension by rewriting those imports during transformation, but module resolution really shouldn't be a concern of the transform layer. For example:
 - It's important that we fix #618 in a way that works correctly even in files that we aren't transforming at all (consider that vanilla `.ts` files might import from `.gts` files)
 - Glint is already required to intercept TypeScript's view of the project files in ways that the transform layer can't impact (consider that we already support imports of `.gts` files _without_ the extension)

TS has a hard-coded internal set of the extensions it considers to "be" TypeScript, and we work around this fact by presenting custom extensions like `.gjs` and `.gts` as `.js` and `.ts` files respectively. This is how such files get included during whole-project typechecking, and is also how extensionless imports of `.gts` and `.gjs` are resolved today.

This PR addresses #618 in the same manner as we handle custom extensions in general: by providing a custom hook for the compiler API to implement the behavior we want. In this instance, the hook is `resolveModuleNameLiterals` rather than the system-level `readFile` and friends, but the gist is the same: when a `.gts` file is imported, we intercept and rewrite the request to as though it had been for the corresponding `.ts` file.

### DIFF
--- a/packages/core/__tests__/cli/custom-extensions.test.ts
+++ b/packages/core/__tests__/cli/custom-extensions.test.ts
@@ -3,6 +3,8 @@ import { stripIndent } from 'common-tags';
 import stripAnsi = require('strip-ansi');
 import { describe, beforeEach, afterEach, test, expect } from 'vitest';
 import { Project } from 'glint-monorepo-test-utils';
+import typescript from 'typescript';
+import semver from 'semver';
 
 describe('CLI: custom extensions', () => {
   let project!: Project;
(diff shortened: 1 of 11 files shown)


packages/core/__tests__/cli/custom-extensions.test.ts
```diff
@@ -121,4 +123,47 @@ describe('CLI: custom extensions', () => {
...
+      expect(stripAnsi(result.stderr)).toMatchInlineSnapshot(`
+        "index.gts:1:22 - error TS2307: Cannot find module './Greeting.gts' or its corresponding type declarations.
+
+        1 import Greeting from './Greeting.gts';
+                               ~~~~~~~~~~~~~~~~
+        "
+      `);
+    });
+
+    test.runIf(semver.gte(typescript.version, '5.0.0'))(
```
NullVoxPopuli, line 154: this is a cool util! test.runIf!

NullVoxPopuli (approved):

lukemelia (approved):
👍  Much better in every way than #620 !

lukemelia:
Also, I tried this branch of glint against the project we are trying to convert to a v2 addon, and it appears to build and typecheck successfully.

dfreeman:
Should be available in `@glint/core@1.2.0`!

NullVoxPopuli:
thanks! Already PR'd an update here: https://github.com/embroider-build/addon-blueprint/pull/206

bartocc:
Things are moving fast, it's awesome 👏

PR #625, opened 2023-10-03
Call out `ember-template-imports` config more clearly

BoussonKarel:
Tell the user which key they should add to tsconfig.json. The previous docs text stated that you "should configure it in `tsconfig.json` under `glint.environment`", but it did not say what key should be added.

image


dfreeman:
Thank you!

PR #629, opened 2023-10-06
Update `vscode-uri` and match their docs

NullVoxPopuli:
On projects using 1.2.0 of Glint, they'll run in to this:
```
file:///home/nvp/Development/NullVoxPopuli/polaris-starter-2/node_modules/.pnpm/@glint+core@1.2.0_typescript@5.2.2/node_modules/@glint/core/lib/language-server/util/index.js:3
import VSCodeURI from 'vscode-uri';
       ^^^^^^^^^
SyntaxError: The requested module 'vscode-uri' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)
```

Which makes sense since `@glint/core` didn't declare a dep on `vscode-uri`.

Locally, when testing, I was using the 1.2.0 version of the language server, / vscode extension, so I don't know if that'll need to be published :shrug:

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -37,6 +37,7 @@
     "semver": "^7.5.2",
     "silent-error": "^1.1.1",
     "uuid": "^8.3.2",
+    "vscode-uri": "^3.0.8",
     "vscode-languageserver": "^8.0.1",
     "vscode-languageserver-textdocument": "^1.0.5",
     "vscode-uri": "^3.0.2",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -37,10 +37,9 @@
     "semver": "^7.5.2",
     "silent-error": "^1.1.1",
     "uuid": "^8.3.2",
-    "vscode-uri": "^3.0.8",
     "vscode-languageserver": "^8.0.1",
     "vscode-languageserver-textdocument": "^1.0.5",
-    "vscode-uri": "^3.0.2",
+    "vscode-uri": "^3.0.8",
     "yargs": "^17.5.1"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


packages/vscode/package.json
```diff
@@ -96,7 +96,7 @@
     "@types/mocha": "^10.0.1",
     "@types/vscode": "^1.68.1",
     "@vscode/test-electron": "2.2.0",
-    "@vscode/vsce": "^2.19.0",
+    "@vscode/vsce": "^2.21.1",
```
NullVoxPopuli, line 99: this was also upgraded, because when I was building the extension, the output told me to do upgrade

packages/core/package.json
```diff
@@ -37,6 +37,7 @@
     "semver": "^7.5.2",
     "silent-error": "^1.1.1",
     "uuid": "^8.3.2",
+    "vscode-uri": "^3.0.8",
```
dfreeman, line 40: Look 3 lines down from this

NullVoxPopuli: oh no lol

packages/vscode/package.json
dfreeman: Please revert this, since it has nothing to do with the stated purpose of the PR

NullVoxPopuli: done

packages/core/package.json
NullVoxPopuli: fixed

dfreeman:
Thank you! I'll do a bugfix release shortly so no one else gets bitten by this

Issue #626, opened 2023-10-03
Quest: let's forward _all_ the ServerCapabilities to tsserver


NullVoxPopuli:
Goal:
####  _Achieve parity between tsserver and glint so that folks don't feel the need to have both servers active (which would consume more RAM)_

Each capability should be its own PR -- some PRs will be more involved from others, and some may be straight-forward. 

------------

Here is where we need to configure: https://github.com/typed-ember/glint/blob/main/packages/core/src/language-server/binding.ts#L19

## "`ServerCapabilities`"

Here is the full list:
https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/src/common/protocol.ts#L1068

## Features we want? 

(and probably more)

Here are the features from reported requests we can add
- [ ] https://github.com/typed-ember/glint/issues/601
  Note from discord:
  > semantic highlighting is a little tricky mostly for performance reasons
  > the standard TS plugin for Code actually spawns a second tsserver instance just for semantic highlighting
  
- [ ] TypeScript language server code-folding
  Note from discord:
  > code folding should be straightforward

- [ ] textDocument/documentSymbol
- [ ] progress
      Shows ui indicator in $editor for booting/loading, pending, etc
      https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/src/common/protocol.ts#L1104

camerondubas:
@NullVoxPopuli I'd be happy to help with this. I have a little bit of experience with language servers, but not a ton. Could you point me in the right direction to get started?

NullVoxPopuli:
Hello!, given the list of server capabilities in protocol.ts (linked above), is there a particular one you'd like to start with implementing?

we only want one server capabilities addition per PR (for ease of testing)

IgnaceMaes:
Thanks for setting up this initiative!

Two features I would like to see as well are:

- The `TypeScript: Sort Imports` command

<img width="610" alt="image" src="https://github.com/typed-ember/glint/assets/10243652/657e9771-451f-4f3f-bfcd-3b09eba5642a">

- Inlay type hints to show inferred types inline

<img width="368" alt="image" src="https://github.com/typed-ember/glint/assets/10243652/5df2fdbe-1fdf-4172-8188-468d03f3eb73">

<img width="304" alt="image" src="https://github.com/typed-ember/glint/assets/10243652/4cd3330b-df23-49d5-86bc-2b837bfdf234">

I haven't done any digging so no clue how straight forward these should be.

[...]

PR #632, opened 2023-10-14
Add JS/TS code folding support to language server

camerondubas:
Part of https://github.com/typed-ember/glint/issues/626

This adds support for code folding in js & ts files. This implementation is [based on tsserver's](https://github.com/typescript-language-server/typescript-language-server/blob/651cfcbd28992ad599b00a4ae3bc30132b2a7892/src/lsp-server.ts#L1200). This does not add support for gts/gjs files.

Folding seems to work as expected, but two odd things had to be handled:
- The OutliningSpans have TextSpans with inclusive length, leading to off-by-one errors. I handled this inline, but I'm concerned that this is unintentional and may not always be the case.
- There are some additional OutliningSpans that incorrectly have a "0" starting offset and don't match any foldable code. I believe these are retrieved from the template file and are meant to represent their folding ranges. As is, these are superfluous and cause an incorrect FoldingRange on line 1 of every component file.

Any guidance on better ways to handle either of these would be appreciated!

https://github.com/camerondubas/glint/assets/6216460/b132d090-5ebb-46bb-bd65-1dd1113d9465

### DIFF
--- a/packages/core/__tests__/language-server/folding-ranges.test.ts
+++ b/packages/core/__tests__/language-server/folding-ranges.test.ts
@@ -1,0 +1,238 @@
+import { Project } from 'glint-monorepo-test-utils';
+import { describe, beforeEach, afterEach, test, expect } from 'vitest';
+import { stripIndent } from 'common-tags';
+
+describe('Language Server: Folding Ranges', () => {
+  let project!: Project;
+
+  beforeEach(async () => {
+    project = await Project.create();
+  });
+
+  afterEach(async () => {
+    await project.destroy();
+  });
+
+  test('function', () => {
+    project.write({
+      'example.ts': stripIndent`
+        function foo() {
+          return 'bar';
+        }
+      `,
+    });
+
+    let server = project.startLanguageServer();
+    let folds = server.getFoldingRanges(project.fileURI('example.ts'));
+
+    expect(folds).toEqual([
+      {
+        startLine: 0,
+        endLine: 1,
+        kind: undefined,
+      },
+    ]);
+  });
+
+  test('nested function', () => {
+    project.write({
+      'example.ts': stripIndent`
+        function topLevel() {
+
+          function nested() {
+            return 'bar';
+          }
+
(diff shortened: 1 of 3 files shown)


packages/core/__tests__/language-server/folding-ranges.test.ts
NullVoxPopuli, line 1: I really appreciate you add all these tests in your PRs <3

packages/core/__tests__/language-server/folding-ranges.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      `,
+    });
+
+    let server = project.startLanguageServer();
+    let folds = server.getFoldingRanges(project.fileURI('example.ts'));
+
+    expect(folds).toEqual([
+      {
+        startLine: 0,
+        endLine: 6,
```
NullVoxPopuli, line 57: is this excluding the last `}`?

camerondubas: The last line is excluded if it's just a closing bracket (https://github.com/typed-ember/glint/pull/632/files#diff-cadffc59ac4fd3c9fd1cd7488980fc9cfea40cd5a93129f1021c54c810980e20R503). This is so that folded functions keep the `}` visible:
```ts
// Like this when folded
function topLevel() {
}

// instead of this
function topLevel() {
```

this is consistent with the TS language server's folding.

camerondubas:
@dfreeman I updated this with the latest main. Friendly bump to take a look at this when you have some time. No rush though, life's busy! Cheers

lifeart:
Here is template part if needed: <https://github.com/lifeart/ember-language-server/blob/component-context-info-origin/src/folding-provider/template-folding-provider.ts>, quite easy to implement.

NullVoxPopuli:
looks like a floating dep dropped node 16, breaking our ci.

I'd consider this non-blocking for merge, myself, and fixing the floating dep issue can happen in a different PR

PR #635, opened 2023-10-16
Add "Sort Imports" command to vscode extension & lsp

camerondubas:
As requested here https://github.com/typed-ember/glint/issues/626, this adds a "Sort Imports" command to the vscode extension and implements a request in the language server to get the TextEdits needed to organize the imports of the current file.

- works with `gts` files as well as `ts`
- currently set to be non-destructive, so unused imports only reordered, not removed. Wasn't sure what the standard should be for this, or if it should be configurable by the user.

https://github.com/typed-ember/glint/assets/6216460/18df1dbd-670e-476e-ab0c-ce4b3dcb5e4d


packages/core/__tests__/language-server/requests.test.ts
```diff
@@ -0,0 +1,328 @@
...
+      `,
+    });
+
+    let server = project.startLanguageServer();
+
+    let formatting = ts.getDefaultFormatCodeSettings();
+    let preferences = {};
+    let edits = server.organizeImports(project.fileURI('index.ts'), formatting, preferences);
+
+    expect(edits).toEqual([
```
NullVoxPopuli, line 234: why are their edits if the imports are already sorted?

packages/core/src/language-server/binding.ts
```diff
@@ -219,6 +219,15 @@ export function bindLanguageServerPool({
     return pool.withServerForURI(uri, ({ server }) => server.getTransformedContents(uri));
   });
 
+  connection.onRequest(SortImportsRequest.type, ({ uri }) => {
+    return pool.withServerForURI(uri, ({ server }) => {
+      const language = server.getLanguageType(uri);
+      const formatting = configManager.getFormatCodeSettingsFor(language);
+      const preferences = configManager.getUserSettingsFor(language);
+      return server.organizeImports(uri, formatting, preferences);
```
NullVoxPopuli, line 227: :partying_face:

packages/core/__tests__/language-server/requests.test.ts
NullVoxPopuli: based on this code: https://github.com/typed-ember/glint/pull/635/files#diff-cadffc59ac4fd3c9fd1cd7488980fc9cfea40cd5a93129f1021c54c810980e20R432

it looks like "that's just the upstream behavior" :shrug:

NullVoxPopuli (approved):
thanks for doing this! it's exciting to see more features be added!

packages/core/src/language-server/messages.cts
```diff
@@ -7,10 +7,15 @@ export type Request<Name extends string, T> = {
...
+  ProtocolRequestType<RequestParams, GetIRResult | null, void, void, void>
 );
 
-export interface GetIRParams {
+export const SortImportsRequest = makeRequestType(
+  'glint/sortImports',
+  ProtocolRequestType<RequestParams, TextEdit[] | null, void, void, void>
+);
+
+export interface RequestParams {
```
dfreeman, line 18: It's arguably incidental that `getIR` and `sortImports` both just want a `uri` param, while other requests might require more (or less!) information. Given that, let's keep separate `xParams`/`xResult` types for each message.

packages/vscode/src/extension.ts
```diff
@@ -57,6 +60,32 @@ async function restartClients(): Promise<void> {
...
+    return;
+  }
+
+  const workspaceEdit = new WorkspaceEdit();
+
+  for (const edit of edits) {
+    const start = new Position(edit.range.start.line, edit.range.start.character);
+    const end = new Position(edit.range.end.line, edit.range.end.character);
+    const range = new Range(start, end);
+    workspaceEdit.replace(editor.document.uri, range, edit.newText);
```
dfreeman, line 83: Is there a reason you need to reconstruct a fresh `Range` object here?

```suggestion
    workspaceEdit.replace(editor.document.uri, edit.range, edit.newText);
```

packages/vscode/package.json
```diff
@@ -50,6 +50,10 @@
         "title": "Glint: Show IR for Debugging",
         "command": "glint.show-debug-ir",
         "enablement": "config.glint.debug == true"
+      },
+      {
+        "title": "Glint: Sort Imports",
+        "command": "glint.sort-imports"
```
dfreeman, line 56: Can we limit this command to only show up on files where it's relevant? It looks like the equivalent commands from Code's built-in JS/TS language features extension matches on [`editorLangId`](https://code.visualstudio.com/api/references/when-clause-contexts#available-context-keys).

packages/core/src/language-server/glint-language-server.ts
```diff
@@ -416,6 +416,45 @@ export default class GlintLanguageServer {
...
+  public organizeImports(
+    uri: string,
+    formatOptions: ts.FormatCodeSettings = {},
+    preferences: ts.UserPreferences = {}
+  ): TextEdit[] {
+    const transformInfo = this.transformManager.findTransformInfoForOriginalFile(
+      uriToFilePath(uri)
+    );
+
+    if (!transformInfo) {
```
dfreeman, line 428: An `organizeImports` call should still be actionable even if we haven't done any transformation on the file, as long as it's something we recognize as one of our known script extensions, right?

packages/core/src/language-server/glint-language-server.ts
```diff
@@ -416,6 +416,45 @@ export default class GlintLanguageServer {
...
+      preferences
+    );
+    const edits: TextEdit[] = [];
+
+    for (const fileTextChange of fileTextChanges) {
+      for (const textChange of fileTextChange.textChanges) {
+        const location = this.textSpanToLocation(fileTextChange.fileName, textChange.span);
+        if (location) {
+          edits.push({
+            range: location.range,
```
dfreeman, line 448: It's unusual for imports to appear anywhere other than the top of a file, but if they do, I believe TS will sort each group in-place.

If imports appear after a transformed span, then I think this range will be off—if so, it likely needs to be converted using something like `getOriginalRange` to convert from the 'transformed' version TS sees.

packages/core/__tests__/language-server/requests.test.ts
```diff
@@ -0,0 +1,328 @@
+import { Project } from 'glint-monorepo-test-utils';
+import { describe, beforeEach, afterEach, test, expect } from 'vitest';
+import { stripIndent } from 'common-tags';
+import * as ts from 'typescript';
+
+describe('Language Server: Requests', () => {
```
dfreeman, line 6: Can we call this something like `Language Server: Organize Imports`/`organize-imports.test.ts`?

Technically lots of messages that come in are requests; it's just that most of the ones we have implemented today have dedicated handlers in the `vscode-languageserver` API like `connection.onDefinition(...)` rather than using the generic `connection.onRequest(...)`.

packages/core/__tests__/language-server/requests.test.ts
```diff
@@ -0,0 +1,328 @@
...
+          range: {
+            start: { character: 0, line: 4 },
+            end: { character: 0, line: 5 },
+          },
+        },
+      ]);
+    });
+  });
+
+  test('many imports, already sorted', () => {
```
dfreeman, line 202: How many of these tests are exercising different code paths in Glint, versus different code paths within TypeScript itself? This is the opposite of the question I usually find myself asking in code reviews, but do you think each of these tests is providing unique value? 🙂

dfreeman (commented):
Thank you _so_ much for taking this on, @camerondubas! It's super exciting to see new contributors digging in and making hefty contributions like this 🙂

dfreeman:
@ijlee2 I imagine it would interact in exactly the same way as that plugin does with the built-in "Sort Imports" commands that are provided for regular JS and TS files today. Presumably if you're using the eslint plugin, then you aren't invoking this command and instead are relying on `eslint --fix` or fix-on-save instead.

ijlee2:
@dfreeman Sounds good, thanks for the clarification.

packages/vscode/src/extension.ts
camerondubas: The types for `vscode` (eg. [Range](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vscode/index.d.ts#L399)) differ from `vscode-language-server` (eg. [Range](https://github.com/microsoft/vscode-languageserver-node/blob/main/types/src/main.ts#L181)) slightly, so we have modify the response to match what the client expects. Kind of annoying :shrug: 

We could return a value that matches a `vscode` type, but then the language server wouldn't be editor-agnostic.

vscode's built-in extension adds [helper methods to `Range`](https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/src/typeConverters.ts#L75) to perform the conversion before [calling `applyEdit`](https://github.com/microsoft/vscode/blob/243e6cec5096b301259ca1f780e50b135f308d33/extensions/typescript-language-features/src/languageFeatures/organizeImports.ts#L106). We could do the same, but it seemed a bit premature?

packages/vscode/package.json
camerondubas: Good idea!  I'll set this up. How should the command be named? 
- `Glint: Sort Imports`: no language 
- `Glint: Sort Typescript Imports`: inline language 
- `Typescript: Sort Imports`: prefix language. This mimics the built-in extension. could be helpful or confusing, not sure which

GlimmerJS & GlimmerTS could be languages as well for their respective files.

packages/core/src/language-server/glint-language-server.ts
camerondubas: Not sure I completely understand, but I only had to get the transformInfo in order for `.gts` files. Passing them to tsserver directly would error. Using the original file name seemed to work fine for `.ts` files.

packages/core/__tests__/language-server/requests.test.ts
camerondubas: Ha, fair enough there's probably some duplication in what's being covered. I'll take another look at the suite

packages/vscode/src/extension.ts
dfreeman: Ah! Lovely. Thanks, `vscode` 😏 

It looks like `Range` has a secondary construct signature that accepts four numbers instead of requiring the intermediary `Position` objects. Maybe that's a slight improvement?

```ts
    const { start, end } = edit.range;
    const range = new Range(start.line, start.character, end.line, end.character);
    workspaceEdit.replace(editor.document.uri, range, edit.newText);
```

packages/core/src/language-server/glint-language-server.ts
dfreeman: As this stands, invoking this command on a regular `.ts` file with no accompanying template won't do anything, since we're short-circuiting here. Unless I'm misreading the code?

camerondubas: `transformInfo` will only be `null` for `.hbs` files without a corresponding `.ts`.  For regular TS files, a `transformInfo` object is returned with a valid `transformedFileName` but a null `transformedModule`.

Since this command is not available in `.hbs` files, I don't think it's a major concern

packages/core/src/language-server/glint-language-server.ts
camerondubas: `this.textSpanToLocation` calls `getOriginalRange` internally to account for this. 

Manually confirmed that `import`s after `<template>` tags in `.gts` files sort correctly and added this scenario to the test cases

NullVoxPopuli:
@camerondubas looks like there are lint failures? Can you address and push?
Thanks!

packages/core/src/language-server/glint-language-server.ts
dfreeman: > For regular TS files, a transformInfo object is returned with a valid transformedFileName but a null transformedModule.

Ahh, that's what I get for trying to review from memory on code I haven't actually looked at in detail for a long time. Thanks for clarifying!

dfreeman (approved):
I hope you don't mind I pushed the lint fixes myself @camerondubas! I have a bit of time this week and wanted to make sure this made it into the release I'll be putting out 🙂

We have a test issue right now that only occurs with the TS nightly build and is only reproducible in CI (💀), but that has nothing to do with your change here, so I think we're good to go. Thank you so much! 🎉

camerondubas:
Thanks for getting this in!

bartocc:
Thx for this feature 🙏

We've been using `source.organizeImports` for a while, and this both sorts, but also removes unused imports.
@camerondubas , I know you chose to not implement this feature in this PR, but are there any plans on adding this?

camerondubas:
@bartocc Ya I could take a look at that. Should have some time over the next week

Issue #597, opened 2023-07-04
Option to not clear the screen when in watch mode?


NullVoxPopuli:
Over in the v2-addon blueprint, we're proposing changing compilation to use:
```
    "start": "concurrently 'npm:start:*'",
    "start:js": "rollup --config --watch --no-watch.clearScreen",
    "start:types": "glint --build --watch",
```
https://github.com/embroider-build/addon-blueprint/pull/136

But, it seems `glint` doesn't have a flag to disable the screen clearing? 
I see that this is the file where the option would be added: https://github.com/typed-ember/glint/blob/main/packages/core/src/cli/index.ts, but how does screen clearing work? https://github.com/search?q=repo%3Atyped-ember%2Fglint%20clear&type=code

dfreeman:
> how does screen clearing work

That's not something we're doing ourselves, it's TypeScript. They have a `--preserveWatchOutput` flag, though, and that may be something we can pass through to the compiler API—but some of their CLI flags don't have any public programmatic equivalent, so someone will have to do a bit of investigation there.

dfreeman:
Fixed in #637

PR #637, opened 2023-10-19
Add --preserveWatchOutput flag to avoid clearing console output in watch mode

lukemelia:
Normally, output is cleared every time a change/re-check happens.

Correspondes to the tsc flag of the same name.


lukemelia:
Fixes #597

dfreeman (approved):
ESLint seems unhappy about something, but otherwise this looks great to me. Thank you, @lukemelia!

lukemelia:
Thanks for the review, @dfreeman. I fixed up the lint error.

PR #647, opened 2023-11-07
docs: Add "Template Imports" page

lukasnys:
Opening this PR as a result of the conversation [here](https://github.com/typed-ember/glint/discussions/641#discussioncomment-7496774).

This PR adds a `Template Imports` page to the docs to capture some details about how Glint can be used together with `ember-template-imports`

### DIFF
--- a/docs/ember/template-imports.md
+++ b/docs/ember/template-imports.md
@@ -1,3 +1,5 @@
+When adding Glint to an Ember project with `ember-template-imports` installed, there are a few additional things to consider.
+
 ## Installation
 
 In addition to the `@glint/core`, `@glint/template` and `@glint/environment-ember-loose` packages, you also need to install the `@glint/environment-ember-template-imports` package and configure it in `tsconfig.json` under `glint.environment`:
--- a/docs/ember/template-imports.md
+++ b/docs/ember/template-imports.md
@@ -64,7 +64,7 @@ louderPlease = (message: string) => message.toUpperCase();
 
 const Shout: TOC<ShoutSignature> = <template>
     <div ...attributes>
-        {{yield (louderPlease this.args.message)}}
+        {{yield (louderPlease @message)}}
     </div>
 </template>;
 export default Shout;
--- a/docs/ember/template-imports.md
+++ b/docs/ember/template-imports.md
@@ -42,3 +42,30 @@ import '@glint/environment-ember-template-imports';
 ```
 
 {% endcode %}
+
+## Template-Only Components
+
+When using `ember-template-imports`, you can define a backing module using the `TOC` type:
+
+{% code title="app/components/shout.gts %}
+
+```typescript
+import type { TOC } From '@ember/component/template-only';
+
+interface ShoutSignature {
+  Element: HTMLDivElement;
+  Args: { message: string };
+  Blocks: {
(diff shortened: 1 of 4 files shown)


docs/ember/template-imports.md
```diff
@@ -0,0 +1,73 @@
...
+
+```typescript
+import '@glint/environment-ember-template-imports';
+```
+
+{% endcode %}
+
+## Template-Only Components
+
+When using `ember-template-imports`, you can define a backing module using the `TOC` type:
```
dfreeman, line 48:
```suggestion
When using `ember-template-imports`, you can declare the type of a `<template>` component using the `TOC` type:
```

docs/ember/template-imports.md
```diff
@@ -0,0 +1,73 @@
...
+  Blocks: {
+    default: [shoutedMessage: string];
+  };
+}
+
+louderPlease = (message: string) => message.toUpperCase();
+
+const Shout: TOC<ShoutSignature> = <template>
+    <div ...attributes>
+        {{yield (louderPlease this.args.message)}}
```
dfreeman, line 67:
```suggestion
        {{yield (louderPlease @message)}}
```

Template-only components (including `<template>` expressions) don't have a `this` value—fortunately Glint would have flagged this in a real codebase 😄

docs/ember/template-imports.md
```diff
@@ -0,0 +1,73 @@
...
+
+interface ShoutSignature {
+  Element: HTMLDivElement;
+  Args: { message: string };
+  Blocks: {
+    default: [shoutedMessage: string];
+  };
+}
+
+louderPlease = (message: string) => message.toUpperCase();
```
dfreeman, line 63:
```suggestion
const louderPlease = (message: string) => message.toUpperCase();
```

docs/ember/template-only-components.md
```diff
@@ -26,6 +26,13 @@ declare module '@glint/environment-ember-loose/registry' {
 
 Note that the runtime content of this module (effectively `export default templateOnlyComponent();`) is exactly what Ember generates at build time when creating a backing module for a template-only component.
 
+{% hint style="info" %}
+
+Using `ember-template-imports`? See [Ember: Template Imports][etitoc] on how to create a backing module for a template-import template-only component.
```
dfreeman, line 31:
```suggestion
Using `ember-template-imports`? See [Ember: Template Imports][etitoc] on how to declare types for a `<template>`-only component.
```

dfreeman (commented):
Thank you, @lukasnys, this looks great! I think you'll also need to update [docs/contents.md](https://github.com/typed-ember/glint/blob/main/docs/contents.md?plain=1) for this to be published properly to Gitbook.

docs/ember/template-imports.md
lukasnys: Whoops 😅 good catch!

dfreeman:
For some reason that last commit doesn't seem to have triggered Gitbook to rebuild, so I can't tell whether that fix did the trick 🤔 

(It _looks_ right, though!)

dfreeman:
Based on our chat in Discord, I think this is good to go. Thank you so much for seeing this through @lukasnys!

Issue #628, opened 2023-10-06
Import of .gts emits broken declarations


simonihmig:
As [reported on Discord](https://discord.com/channels/480462759797063690/786312479620726804/1159538780919832698), there seems to be a bug when having an import with an explicit `.gts` extension and running `glint --declaration`. 

Reproduction:
* go to https://stackblitz.com/edit/github-yl7xy1
* `cd packages/accordion`
* `pnpm build`
* in `packages/accordion/declarations/components/accordion.d.ts` you can see an import of `item.gts`, which does not resolve, as there is only an `item.d.ts`, which TS/Glint does not see as the declaration that matches this import apparently:
![image](https://github.com/typed-ember/glint/assets/1325249/38924836-09f4-42d5-a013-a95a0088dd28)

The emitted declarations can be fixed by patching manually, either
* by changing the import to be extension-less
* or by renaming `item.d.ts` to `item.gts.d.ts`

So I think `glint --declaration` should be doing either of these.

NullVoxPopuli:
Demo: https://stackblitz.com/edit/stackblitz-starters-cj9ass?file=README.md

enspandi:
Not sure if useful for others, but I'm using this simple workaround for now:

<details>
  <summary>Patch writeFile to search/replace .gts to .ts</summary>
  
```patch
  diff --git a/lib/cli/perform-check.js b/lib/cli/perform-check.js
index 6bef94d92a5026f78073c8c35b6f8578df8e4173..632c5f3d2b98837b973c12dc28ff01b3346df8f1 100644
--- a/lib/cli/perform-check.js
+++ b/lib/cli/perform-check.js
@@ -47,6 +47,19 @@ function createCompilerHost(ts, options, transformManager) {
     host.fileExists = transformManager.fileExists;
     host.readFile = transformManager.readTransformedFile;
     host.readDirectory = transformManager.readDirectory;
+
+    // Postprocess .d.ts files to temporarily patch '.gts' to '.ts'
+    // https://github.com/typed-ember/glint/issues/628
+    const tsWriteFile = host.writeFile;
+    const matchGtsImport = /\.gts';/gm;
+    host.writeFile = (fileName, data, writeByteOrderMark, onError) => {
+        const isDts = fileName.endsWith('.d.ts');
+        if (isDts && matchGtsImport.test(data)) {
+            data = data.replace(matchGtsImport, ".ts';");
+        }
[...]

PR #648, opened 2023-11-07
Fix support for gts extensions remaining in emitted declarations

NullVoxPopuli:
Resolves: https://github.com/typed-ember/glint/issues/628

-------------

Given imports of:
```ts
// index.ts
import { Greeting } from './greeting.gts';
import { greeter } from './greeter.ts';
```
emitted declarations created via `glint --declaration` contain:
```ts
// index.d.ts
import { Greeting } from './greeting.gts';
import { greeter } from './greeter.ts';
```

This is incorrect, as in order to make `greeting.gts` and `greeter.ts` valid imports in declarations, the following declaration files would need to exist:
```
- ./index.d.ts <- the entrypoint (pictured above)
- ./greeting.gts.d.ts
- ./greeter.d.ts <- tsc has built in logic for this
```

A goal is not _just_ have Glint support gjs/gts imports, but have the emitted output of `glint --declarations` be compatible with `tsc`.

So, to do that, for `.gjs` and `.gts` imports, we'll need to include `.gjs` and `.gts` in the name of the emitted assets.

<details><summary>the old approach</summary>

So one way to deal with this is to, when transforming the files, chop off the extensions entirely so that the generated `index.d.ts` would contain:
```ts
// index.d.ts
import { Greeting } from './greeting';
import { greeter } from './greeter';
```
which means that, the currently emitted `.d.ts`-only files would resolve...
_if_ the resolving code is changed to support iterating over all possible file extensions -- which could lead to slower declaration building.

<details><summary>background</summary>

using the learnings from work on the rollup plugins, it was _much_ easier to work with all the different files with explicit extensions in the imports.

Resolving got _way_ easier, because the extensions were just right there in the import, and correct to what matched on disk. 

Unlike how TS has totally make extensions imports a nightmare for some projects, extension usage on `@embroider/addon-dev` _makes sense_ (specifying the author format, and it's compiled away in the resulting js).

-----------

I think it'd be easier and less code to emit `basename(<import specifier>).d.ts` for every import, than to
- remove the extension from all imports
- add resolving code everywhere to re-find the files

(which would def be a breaking change without a flag to opt in to this behavior)

</details>

</details>

-------------

- [x] failing test
- [ ] fix
  - [ ] add tsconfig.json [flag](https://github.com/typed-ember/glint/pull/648#discussion_r1385305024)

Support for resolving `gts`-extensioned files was added in: https://github.com/typed-ember/glint/pull/621 
This PR finishes that work by making sure emitted declarations can resolve other declarations.


packages/core/__tests__/cli/custom-extensions.test.ts
```diff
@@ -165,5 +173,24 @@ describe('CLI: custom extensions', () => {
         expect(result.stderr).toBe('');
       }
     );
+
+    test.runIf(semver.gte(typescript.version, '5.0.0'))(
+      'declarations work with `allowImportingTsExtensions: true`',
```
dfreeman, line 178: I think we decided in Discord that driving this emit behavior off of `allowImportingTsExtensions` was a bad idea?

NullVoxPopuli: For me, I'm trying to address one thing at a time -- if I can find where to delete `.gts` from the imports, then I can later un-do any `allowImportingTsExtensions`-specific behavior

dfreeman: Rather than typing it out again, I'm going to just paste a link to the conversation we had in which I said repeatedly that modifying imports likely isn't the path forward here 😉 

https://discord.com/channels/480462759797063690/491905849405472769/1166783585890615377

I believe you, I and @simonihmig all agreed there that a Glint-specific flag to allow including the `.gts` part in the emitted declarations made the most sense?

NullVoxPopuli: I don't actually recall / observe any agreement -- only [a suggestion](https://discord.com/channels/480462759797063690/491905849405472769/1166826588965445682).

Which, I don't know why we'd need it other than "abundance of caution".

dfreeman: > Which, I don't know why we'd need it other than "abundance of caution".

Simon found a real-world public example of where this would be a breaking change.

> I don't actually recall / observe any agreement

```
Simon Ihmig — 10/25/2023 9:51 PM
I'd rather introduce a new flag then tbh
NullVoxPopuli — 10/25/2023 9:55 PM
works for me / I have no problem with that / simpler = better
```

I read that as an agreement (and while I had `+1`ed an earlier comment, I'll make my agreement explicit here: I think that's the preferable approach)

NullVoxPopuli: :100:

packages/core/__tests__/transform/offset-mapping.test.ts
```diff
@@ -464,7 +464,12 @@ describe('Transform: Source-to-source offset mapping', () => {
       `,
     };
 
-    const rewritten = rewriteModule(ts, { script: source }, glimmerxEnvironment)!;
+    let rewritten;
+
+    beforeEach(() => {
```
NullVoxPopuli, line 469: without these changes, debugging was super troll time

packages/core/__tests__/cli/custom-extensions.test.ts
```diff
@@ -135,6 +135,15 @@ describe('CLI: custom extensions', () => {
...
           <template>Hello!</template>
         `,
+        're-export.gts': stripIndent`
+          export { default as Greeting } from './Greeting.gts';
+        `,
+        'vanilla.ts': 'export const two = 2;',
+        'barrel.ts': stripIndent`
+          export { default as Greeting } from './Greeting.gts';
+          export { Greeting as Greeting2 } from './re-export.gts';
+          export { two } from './vanilla.ts';
```
NullVoxPopuli, line 145: the declarations for this are, at present, emitted as `./vanilla.ts`, which is wrong.

To fix, either needs to happen:
- remove the `.ts` extension so `vanilla.d.ts` matches with the emitted corresponding `vanilla.js`
- or emit a goofy `vanilla.ts.d.ts` file, along the same lines as what I suggested originally with emitting `Greeting.gts.d.ts`

NullVoxPopuli:
Gonna close because I don't have a solution in mind.


Demo: https://stackblitz.com/edit/stackblitz-starters-cj9ass?file=README.md

Additional Context: https://discord.com/channels/480462759797063690/568935504288940056/1171853131521470534

PR #655, opened 2023-11-30
Use `content-tag` instead of `ember-template-imports`

NullVoxPopuli:
**`ember-template-imports @ v4` removed all the APIs that glint was relying on for working with `<template>`-tags**

Resolves: https://github.com/typed-ember/glint/issues/609
Resolves: https://github.com/typed-ember/glint/issues/706
Resolves: https://github.com/typed-ember/glint/issues/694
~~Blocked by: https://github.com/embroider-build/content-tag/issues/19~~
~~Blocked by: https://github.com/embroider-build/content-tag/pull/31~~

Unblocks: https://github.com/NullVoxPopuli/polaris-starter/pull/11
Supersedes: https://github.com/typed-ember/glint/pull/615

Extractions:
- https://github.com/typed-ember/glint/pull/656
- https://github.com/typed-ember/glint/pull/657

Allows:
- folks to `<pm> add ember-template-imports` in their apps and not have to downgrade to v3 if they want to use Glint
- better syntax handling
- better incorrect syntax handling

### DIFF
--- a/packages/core/src/transform/template/rewrite-module.ts
+++ b/packages/core/src/transform/template/rewrite-module.ts
@@ -71,7 +71,39 @@ function calculateCorrelatedSpans(
   let errors: Array<TransformError> = [];
   let partialSpans: Array<PartialCorrelatedSpan> = [];
 
-  let { ast, emitMetadata } = parseScript(ts, script, environment);
+  let { ast, emitMetadata, error } = parseScript(ts, script, environment);
+
+  if (error) {
+    if (typeof error === 'string') {
+      errors.push({
+        message: error,
+        location: { start: 0, end: script.contents.length - 1 },
+        source: script,
+      });
+    } else if ('isContentTagError' in error && error.isContentTagError) {
+      // these lines exclude the line with the error, because
+      // adding the column offset will get us on to the line with the error
+      let lines = script.contents.split('\n').slice(0, error.line);
+      let start = lines.reduce((sum, line) => sum + line.length, 0) + error.column - 1;
+      let end = start + 1;
+
+      errors.push({
+        isContentTagError: true,
+        // we have to show the "help" because content-tag has different line numbers
+        // than we are able to discern ourselves.
+        message: error.message + '\n\n' + error.help,
+        source: script,
+        location: {
+          start,
+          end,
+        },
+      });
+    }
+
+    // We've hit a parsing error, so we need to immediately return as the parsed
+    // document must be correct before we can continue.
+    return { errors, directives, partialSpans };
+  }
 
   ts.transform(ast, [
     (context) =>
@@ -103,9 +135,22 @@ function calculateCorrelatedSpans(
   return { errors, directives, partialSpans };
 }
 
+type ParseError =
+  | string
+  | {
+      isContentTagError: true;
+      message: string;
+      line: number;
+      column: number;
+      file: string;
+      help: string;
+      raw: string;
+    };
+
 type ParseResult = {
   ast: ts.SourceFile;
   emitMetadata: WeakMap<ts.Node, GlintEmitMetadata>;
+  error?: ParseError;
 };
 
 function parseScript(ts: TSLib, script: SourceFile, environment: GlintEnvironment): ParseResult {
@@ -116,15 +161,37 @@ function parseScript(ts: TSLib, script: SourceFile, environment: GlintEnvironmen
     void emitMetadata.set(node, Object.assign(emitMetadata.get(node) ?? {}, data));
 
   let { preprocess, transform } = environment.getConfigForExtension(extension) ?? {};
-  let preprocessed = preprocess?.(contents, filename) ?? { contents };
+
+  let original: {
+    contents: string;
+    data?: {
+      // SAFETY: type exists elsewhere (the environments)
+      templateLocations: any[];
+    };
+  } = { contents, data: { templateLocations: [] } };
+
+  let preprocessed = original;
+  let error: ParseError | undefined;
+
+  try {
+    preprocessed = preprocess?.(contents, filename) ?? original;
+  } catch (e) {
+    error = parseError(e, filename);
+  }
+
   let ast = ts.createSourceFile(
     filename,
+    // contents will be transformed and placeholder'd content
+    // or, in the event of a parse error, the original content
+    // in which case, TS will report a ton of errors about some goofy syntax.
+    // We'll want to ignore all of that and only display our parsing error from content-tag.
     preprocessed.contents,
     ts.ScriptTarget.Latest,
     true // setParentNodes
   );
 
-  if (transform) {
+  // Only transform if we don't have a parse error
+  if (!error && transform) {
     let { transformed } = ts.transform(ast, [
       (context) => transform!(preprocessed.data, { ts, context, setEmitMetadata }),
     ]);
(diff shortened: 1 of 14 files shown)


packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,54 +1,91 @@
...
-    delta += endTagLength - TEMPLATE_END.length;
+    deltaBytes += endTagLengthBytes - TEMPLATE_END.length;
 
-    sourceOffset = endTagOffset + endTagLength;
+    sourceOffsetBytes = endTagOffsetBytes + endTagLengthBytes;
 
+    // TODO: is there a way to convert bytes to chars?
+    //       I think maybe all of this code needs to live in content-tag,
+    //       and give us the option to generate this sort of structure
     templateLocations.push({
```
NullVoxPopuli, line 78: This is all likely wrong, as the original code used character-indices

yarn.lock
```diff
@@ -6461,7 +6466,7 @@ ember-template-imports@^1.1.1:
     ember-cli-version-checker "^5.1.2"
```
NullVoxPopuli, line 6466: above here, more reason to have glint not depend on content-tag:
```
=> Found "ember-template-imports@1.1.1"
info Reasons this module exists
   - "_project_#@glint#environment-glimmerx#@glimmerx#component" depends on it
   - Hoisted from "_project_#@glint#environment-glimmerx#@glimmerx#component#ember-template-imports"
```

yarn.lock
```diff
@@ -6461,7 +6466,7 @@ ember-template-imports@^1.1.1:
...
     validate-peer-dependencies "^1.1.0"
 
-ember-template-imports@^3.0.0, ember-template-imports@^3.1.1:
+ember-template-imports@^3.1.1:
```
NullVoxPopuli, line 6469: This'll soon be removed from ember-template lint as well:

```
=> Found "ember-template-lint#ember-template-imports@3.4.0"
info This module exists because "_project_#ts-ember-app#ember-template-lint" depends on it.
info Disk size without dependencies: "232KB"
```

NullVoxPopuli:
I'm marking as ready for review, as I think this is about at the point, where I want to make sure I'm not doing anything crazy.

There is still a failing test: https://github.com/typed-ember/glint/actions/runs/7254626609/job/19763687113?pr=655#step:6:107
which... I have no idea how this code affects the reporting of untyped `{{component}}` usage, since in the "debug extension" thing, the 4 use cases all error correctly (as expected).

cc @dfreeman @wagenet @gitKrystan 


Some things left for this PR tho:
- [ ] push more work in to content-tag, per @ef4's comment here: https://github.com/embroider-build/content-tag/issues/39#issuecomment-1832443310
- [ ] put content-tag in ember-template-lint
      ideally, when this PR is done, we'll see ember-template-imports removed from the lockfile entirely (tho, we'd also have to remove GlimmerX for that.... )

packages/core/src/transform/template/rewrite-module.ts
```diff
@@ -116,15 +159,77 @@ function parseScript(ts: TSLib, script: SourceFile, environment: GlintEnvironmen
...
+
+  let preprocessed = original;
+  let error: ParseError | undefined;
+
+  try {
+    preprocessed = preprocess?.(contents, filename) ?? original;
+  } catch (e) {
+    if (typeof e === 'object' && e !== null) {
+      // Parse Errors from the rust parser
+      if ('source_code' in e) {
```
NullVoxPopuli, line 179: It may be useful for content-tag to emit this error as a JSON object or something so I don't have to parse the error-string here.

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,54 +1,91 @@
...
+//
+// ...Except,
+//    > the referenced file is an ECMAScript module
+//
+//    package.json#exports does refer to a cjs file if required, so TS should be resolving the `require`
+//    entries not the `import` entries.
+//
+//    https://github.com/embroider-build/content-tag/blob/v1.2.2-content-tag/package.json#L13-L21
+//
+// @ts-expect-error see above
```
NullVoxPopuli, line 22: even though this ts-expect-error is here, we still get the correct type for `Preprocessor` and `p.parse` below

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,54 +1,91 @@
...
+  // in the root package.json
+  let sourceBuffer = Buffer.from(source);
 
   for (let template of templates) {
-    let { start, end } = template;
-    let startTagLength = template.start[0].length;
-    let endTagLength = template.end[0].length;
-    let startTagOffset = start.index ?? -1;
-    let endTagOffset = end.index ?? -1;
+    let startTagLengthBytes = template.startRange.end - template.startRange.start;
```
NullVoxPopuli, line 41: I added `Bytes` to all the variables here as a suffix to keep the units straight.

See also: https://github.com/embroider-build/content-tag/issues/45

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,54 +1,91 @@
...
 
-    segments.push(source.slice(sourceOffset, startTagOffset));
+    /**
+     * TODO: we want content-tag to manage all this for us, as managing indicies
+     *       can be error-prone.
+     *
+     * SEE: https://github.com/embroider-build/content-tag/issues/39#issuecomment-1832443310
+     */
+    let prefixingSegment = sourceBuffer.slice(sourceOffsetBytes, startTagOffsetBytes);
+    segments.push(prefixingSegment.toString());
```
NullVoxPopuli, line 57: is there a way to _not_ do this segments stuff?

@dfreeman since content-tag can transform to valid JS, could we just parse that instead?

I imagine this is what @patricklx may have run in to in the past as well while iterating on the eslint parser, we need a pre-transform to post-transform index mapping for all AST nodes

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,54 +1,91 @@
...
   let templateLocations: Array<TemplateLocation> = [];
   let segments: Array<string> = [];
-  let sourceOffset = 0;
-  let delta = 0;
+  let sourceOffsetBytes = 0;
+  let deltaBytes = 0;
+
+  // @ts-expect-error TS couldn't find @types/node, which are specified
+  // in the root package.json
+  let sourceBuffer = Buffer.from(source);
```
NullVoxPopuli, line 38: See https://runkit.com/nullvoxpopuli/content-tag-byte-vs-char-offsets for a brief overview of how `Buffer` solves the byte-index vs char-index issue.

(As well as https://github.com/embroider-build/content-tag/issues/45 )

packages/environment-ember-template-imports/-private/environment/preprocess.ts
gitKrystan: Interesting. I wonder why I don't get this in the prettier plugin.

NullVoxPopuli: is it using 1.2.2 / same tsconfig settings from glint?

lifeart:
I like this changes! We definitely need to push it forward and land it! Works just fine inside custom glint env <https://github.com/lifeart/glimmer-next/tree/master/glint-environment-gxt>

packages/environment-ember-template-imports/-private/environment/preprocess.ts
```diff
@@ -1,54 +1,91 @@
...
-    let endTagLength = template.end[0].length;
-    let startTagOffset = start.index ?? -1;
-    let endTagOffset = end.index ?? -1;
+    let startTagLengthBytes = template.startRange.end - template.startRange.start;
+    let endTagLengthBytes = template.endRange.end - template.endRange.start;
+    let startTagOffsetBytes = template.startRange.start;
+    let endTagOffsetBytes = template.endRange.start;
 
-    if (startTagOffset === -1 || endTagOffset === -1) continue;
+    // if (startTagOffset === -1 || endTagOffset === -1) continue;
```
wagenet, line 46: Should this be removed?

packages/environment-ember-template-imports/-private/environment/preprocess.ts
wagenet: Is this still likely wrong?

wagenet (approved):
I think this is ok. I still don't have 100% context.

packages/environment-ember-template-imports/-private/environment/preprocess.ts
NullVoxPopuli: yup, can do

packages/environment-ember-template-imports/-private/environment/preprocess.ts
NullVoxPopuli: yeah, char positions are expected, but I only have bytes here

patricklx: This is how it is currently done in eslint parser:
https://github.com/NullVoxPopuli/ember-eslint-parser/blob/d1b8796436670e496f4a21de8b1a9d9219038b66/src/parser/transforms.js#L25

NullVoxPopuli: ah ok, I'll try that

NullVoxPopuli: pushed!

this is looking excellent!

image
image
image

tsconfig.compileroptions.json
```diff
@@ -11,6 +11,6 @@
     "composite": true,
     "declaration": true,
     "sourceMap": true,
-    "types": []
+    "types": ["@types/node"]
```
NullVoxPopuli, line 14: this change was needed so that the types for `Buffer` existed

tsconfig.json
```diff
@@ -1,14 +1,14 @@
...
-    { "path": "packages/core" },
-    { "path": "packages/environment-ember-loose" },
-    { "path": "packages/environment-ember-template-imports" },
-    { "path": "packages/environment-glimmerx" },
-    { "path": "packages/template" },
-    { "path": "packages/type-test" },
-    { "path": "packages/vscode" },
-    { "path": "packages/scripts" },
-    { "path": "test-packages/test-utils" }
+    { "path": "./packages/core" },
```
NullVoxPopuli, line 3: in one of the commits, it's explained that this change was needed so that you can ctrl+click to go to each of these tsconfigs.

It's kinda silly that tsc works with both path-formats but vscode only works with leading ./

PR #659, opened 2023-12-14
Make explicit extensions in imports work.

NullVoxPopuli:
Resolves https://github.com/typed-ember/glint/issues/628

See repro repo: https://stackblitz.com/edit/stackblitz-starters-g58nhp?file=examples%2Fts-addon-node16%2Fpackage.json,examples%2Fts-addon-node16%2Fsrc%2Findex.ts
- `pnpm build`
- poke around in the `declarations` directories

Using the patch from here: https://github.com/typed-ember/glint/issues/628#issuecomment-1805416937 
(where TS would allow)

And the test from here: https://github.com/typed-ember/glint/pull/648/files#diff-a835cb215198fc0a50973a4ae3bf66eac5a4b0bc21293fa29fc70888d663f797

The downside to this approach is "what if" someone has a foo.gts.ts file (probably co-located with a foo.gts.hbs file) -- but like, realistically, this isn't a thing they can use unless they're going half-in on static component references.

Notes:
- subpath-imports: https://nodejs.org/api/packages.html#subpath-imports
  this feature can't be used because all subpath-imports must start with a `#` to differentiate from packages.
- subpath-exports: https://nodejs.org/api/packages.html#subpath-exports
  not relevant because from a library perspective, our own exports are not referenced except by consumers

### The Glint Terminal

- clone this branch
- `yarn`
- `yarn build`
- delete `glint-monorepo-test-utils` from `@glint/core`s package.json's devDependencies

### Repro Terminal

- download extract and cd into the repro
- `pnpm i`
- `pnpm link ~/path/to/this/branch/of/glint`
- `pnpm build`

----------

note that after every `pnpm link`, in order to work in the glint repo again, you'll need to re-run yarn, and un-delete the `glint-monorepo-test-utils` package.

### DIFF
--- a/packages/core/__tests__/cli/custom-extensions.test.ts
+++ b/packages/core/__tests__/cli/custom-extensions.test.ts
@@ -135,6 +135,15 @@ describe('CLI: custom extensions', () => {
         'Greeting.gts': stripIndent`
           <template>Hello!</template>
         `,
+        're-export.gts': stripIndent`
+          export { default as Greeting } from './Greeting.gts';
+        `,
+        'vanilla.ts': 'export const two = 2;',
+        'barrel.ts': stripIndent`
+          export { default as Greeting } from './Greeting.gts';
+          export { Greeting as Greeting2 } from './re-export.gts';
+          export { two } from './vanilla.ts';
+        `,
       });
     });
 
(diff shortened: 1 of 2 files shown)


lifeart:
Wondering, is `allowArbitraryExtensions` may help us here? https://www.typescriptlang.org/tsconfig#allowArbitraryExtensions

NullVoxPopuli:
@lifeart yeah, that sounds like a reasonable feature -- but we'd still need glint to to remove the extension during emitting.

So far, I've found it easiest, to work around the problem :(
Not ideal, but I don't know what the correct fix is yet. Maybe it's what you link to!

NullVoxPopuli:
this is resolved in v2 (currently alpha)

PR #663, opened 2024-01-02
provide suggestions and type checking for elements attributes

patricklx:
this would also make it possible to wrap WebComponents and have auto completions and type checking for those.
also allows any key. ideally we should only allow any key, if its prefixed with `data-`.

unfortunately `<WebComponent myoptions="<cursor here>" /> ` does not provide completions.
But `<WebComponent myoptions={{"<cursor here>"}} />` does


patricklx:
Screenshot 2024-01-03 at 12 45 12
Screenshot 2024-01-03 at 12 48 52

packages/template/-private/dsl/types.d.ts
```diff
@@ -10,7 +10,7 @@ export type ResolveOrReturn<T> = T & (<U>(item: U) => () => U);
  */
 export type ElementForTagName<Name extends string> = Name extends keyof HTMLElementTagNameMap
   ? Name extends keyof SVGElementTagNameMap
-    ? HTMLElementTagNameMap[Name] & SVGElementTagNameMap[Name]
+    ? HTMLElementTagNameMap[Name] | SVGElementTagNameMap[Name]
```
patricklx, line 13: comment also says that it should be an union `This will return a union for elements that exist both in HTML and SVG.`
But it was an intersection

patricklx:
hmmm, this needs a way to distinguish exactly between svg elements and html elements...
would need to add a parameter to `emitElement` when in svg context

dfreeman:
Thanks for this, but I think this is conflating HTML attributes and DOM properties in a way that doesn't quite work.

The properties on an `Element` subtype aren't the same as the attributes it accepts. Some attributes and properties do align 1-to-1, like `id`, but many either don't directly align, like `data-*` attributes that correspond to `dataset.*` properties, or only exist on one side or the other, like `aria-*` attributes or the `textContent` property.

patricklx:
😞
then i give up. would have been nice to have this

NullVoxPopuli:
does TS provide no map/list of attributes?

patricklx:
what about this?
https://www.npmjs.com/package/html-element-attributes

dfreeman:
> does TS provide no map/list of attributes?

We spent time looking into this earlier on in Glint's development—I think everyone involved agrees this would be nice to have! TS doesn't include anything like this in its standard library, though, and at least as of when we last checked we didn't find any third-party version either.

> what about this? [npmjs.com/package/html-element-attributes](https://www.npmjs.com/package/html-element-attributes)

At a types level, that just exposes `Record<string, Array<string>>`:
https://github.com/wooorm/html-element-attributes/blob/main/build.js#L88

> We could filter out DOM properties that aren't single-words, so at least the single-word attribute / dom properties match up.

Is that a valid heuristic? I feel pretty strongly that wrong completions are worse than no completions.

NullVoxPopuli:
>  I feel pretty strongly that wrong completions are worse than no completions.

yes, but a subset of incomplete, yet correct completions are better than no completions.

> At a types level, that just exposes Record<string, Array<string>>:

we could PR them to `as const` that whole structure in index.js, which would make the type more specific.
I like this.

dfreeman:
> yes, but a subset of incomplete, yet correct completions are better than no completions.

Yes, which is exactly why I asked if your suggested heuristic of only doing single-word ones actually works 😜

dfreeman:
In other words, my point is that this isn't just a casing issue. Driving off of `Element` properties would include, for instance, `dataset` as a suggestion, which isn't a valid HTML attribute.

patricklx:
I think making a pr to https://www.npmjs.com/package/html-element-attributes to generate a d.ts would then make ut usable

NullVoxPopuli:
Let's see what the maintainer says: https://github.com/wooorm/html-element-attributes/issues/9

I've worked with them before with remark/unified/codemirror stuff before, and iirc, they're pretty responsive.

patricklx:
Would also need https://github.com/wooorm/svg-element-attributes

patricklx:
well, I couldn't give up just yet :)
i build an own element.d.ts based on various packages from woorm.
its all interfaces, so it should be possible to augment them with custom definitions.

now i only get valid completions for element attributes. unfortunately it only works inside an AttrNode
so `<a h$cursor$ ></a>`

packages/core/src/language-server/glint-language-server.ts
```diff
@@ -219,8 +219,12 @@ export default class GlintLanguageServer {
       formatting
     );
 
+    const isInElementAttributes = mapping?.sourceNode.type === 'AttrNode';
+
     return completions?.entries.map((completionEntry) => ({
-      label: completionEntry.name,
+      label: isInElementAttributes
+        ? completionEntry.name.replace(/["']/g, '')
```
patricklx, line 226: to remove `"` or `'` from element attributes containing a `-`

packages/core/src/transform/template/template-to-typescript.ts
```diff
@@ -42,7 +42,7 @@ export function templateToTypescript(
   return mapTemplateContents(originalTemplate, { embeddingSyntax }, (ast, mapper) => {
     let { emit, record, rangeForLine, rangeForNode } = mapper;
     let scope = new ScopeStack([]);
-
+    let inSVG = false;
```
patricklx, line 45: it could be possible to have components inside svg, with this it would be possible to a glint-directive `@glint-in-svg` a the top of the template to specify that we are inside svg.

packages/template/-private/dsl/emit.d.ts
```diff
@@ -133,7 +137,25 @@ export declare function applySplattributes<
  *     <div foo={{bar}}></div>
  *     <AnotherComponent foo={{bar}} />
  */
-export declare function applyAttributes(element: Element, attrs: Record<string, AttrValue>): void;
+
+type WithSvgStrings<T> = {
+  [P in keyof T]?: T[P] extends SVGAnimatedString ? string : T[P];
+};
+
+// TODO: improve this to allow other keys
```
patricklx, line 145: anyone knowns how to achieve this?
basically would like to allow other keys that default to AttrValue.

Currently I do it with `Record<string, AttrValue>`, but that sets all properties to AttrValue

PR #669, opened 2024-01-05
Add  "Organize Imports" command

camerondubas:
As requested here https://github.com/typed-ember/glint/pull/635#issuecomment-1827566067, this adds a `Glint: Organize Imports` command. This is a destructive version of the "Sort Imports" command as it removes unused imports while also sorting.

These commands are analogous to the TS LSP's Sort/Organize commands.

This was just a matter of passing `skipDestructiveCodeActions: false` to the `organizeImports` language service method. As such, I've standardized on using organizeImports instead sortImports in the vscode extension.

### DIFF
--- a/packages/core/__tests__/language-server/organize-imports.test.ts
+++ b/packages/core/__tests__/language-server/organize-imports.test.ts
@@ -26,10 +26,16 @@ describe('Language Server: Organize Imports', () => {
       `,
     });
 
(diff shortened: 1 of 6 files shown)


machty:
Closing this because Volar-ized Glint gets an Organize Import command "for free" from Volar. Thank you for your contribution!

PR #668, opened 2024-01-05
npx create-release-plan-setup@latest --update

NullVoxPopuli:
What's release-plan?

A hybrid of the best parts of [release-it](https://github.com/release-it/release-it) and [changesets](https://github.com/changesets/changesets/).

Repo / docs here: https://github.com/embroider-build/release-plan

This allows folks with GH Maintainer access to preview and release to npm without actually needing to have any keys locally on their machine (or access to the npm package, which makes management of the repo separate from the npm package easy and encouraged).

- Everything is automated via GH Action workflows.
- We release by merging the "prepare release" preview PR.
- Changelog entries are determined by labels on the PRs (see release-plan readme)
  - we edit the changelog text via editing the PR title

See the newly added RELEASE.md for details.

PR #670, opened 2024-01-05
Switch to pnpm

NullVoxPopuli:
`yarn@v1` can't be trusted. this has been known for ages.
Also, newer tooling isn't supporting `yarn@v1`.

unblocks: https://github.com/typed-ember/glint/pull/668

Changes:
- remove yarn
- add pnpm
- Support isolated packages 
  (a requirement to get types passing)
  Previously, everything was all loosey goosey / incorrect.
  - `@glimmer/runtime` was not specified, adding it adds the ESM version of Glimmer. No modern version of `@glimmer/*` exports CJS, and thus is not compatible with Glint (which is compiled to CJS (because VSCode extensions don't support ESM))
  - `@glint/core` was added to any package that invokes the `glint` bin. This is required, because `bin` directories are are local to each package.
  - Other glint packages were added to each package that imports from them, because there is no more hoisting, which means TS can't find the packages unless they are declared in the package.json. This has been done as `devDependencies`
  - We can't have dependencies loosely importing from all over the place, so extra packages have been extracted for integration testing
    - Example: `@glint/core` _cannot_ have tests that consume any environment package, because the environment packages depend on `@glint/core`
  - There are resolution issues with the environments. :(


packages/template/__tests__/emit-content.test.ts
```diff
@@ -1,9 +1,8 @@
-import { SafeString } from '@glimmer/runtime';
```
NullVoxPopuli, line 1: we can't import from Glimmer, because Glimmer is ESM, and Glint is not.

NullVoxPopuli: Alternatively, the tests could be extracted to an ESM package, but then we'd lose the ability to test private things.

A solution could be vitest's in-source testing.
This would require revamping how all the glint packages are built though. (not using tsc)

NullVoxPopuli:
this happened in a separate PR

PR #671, opened 2024-01-06
build: migrate to yarn v4

JounQin:
A promise from https://twitter.com/JounQin/status/1743647752045690975

close #670


package.json
```diff
@@ -39,5 +39,6 @@
     "@types/node": "^20.10.6",
     "ember-cli-htmlbars": "^6.0.1"
   },
-  "version": "1.2.2"
+  "version": "1.2.2",
+  "packageManager": "yarn@4.0.2"
```
JounQin, line 43: `corepack` is required to be enabled.

.yarnrc.yml
```diff
@@ -0,0 +1,3 @@
+nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.0.2.cjs
```
JounQin, line 3: Not using `corepack` because it does not work quite well with Windows.

.yarn/releases/yarn-4.0.2.cjs
NullVoxPopuli, line 1: does this file need to be checked in? wouldn't it come from an npm package?

packages/vscode/__fixtures__/template-imports-app/src/Greeting.gts
```diff
@@ -8,6 +8,6 @@ export default class Greeting extends Component<GreetingSignature> {
   private message = 'Hello';
 
   <template>
-    {{this.message}}, {{@target}}!
+    {{@undocumentedProperty}} {{this.message}}, {{@target}}!
```
NullVoxPopuli, line 11: why this addition?

JounQin: I don't know, I just run `yarn test`

.yarn/releases/yarn-4.0.2.cjs
JounQin: It can be enabled via corepack, but the usage is a bit strict. The current approach is much flexible.

See also https://github.com/typed-ember/glint/pull/671#discussion_r1443797911

> wouldn't it come from an npm package?

It's not in npm.

image

.yarnrc.yml
NullVoxPopuli: Apparently there is some aversion to corepack as well 

https://github.com/nodejs/node/issues/50963#issuecomment-1885881236

Issue #130, opened 2021-04-01
Turning off built-in Typescript in VSCode loses useful features


wagenet:
For instance, if I type `assert` it no longer suggests importing from `@ember/debug`.

dfreeman:
It looks like we're not handling completion items with associated code actions (like adding an `import`) correctly.

bakerac4:
I can second this. I lost the ability to remove unused imports. The other thing is that if it does recommend adding an import it always seems to recommend the relative path which is not what I have in my settings

wagenet:
FWIW, I think the only conflict I've seen with built-in TS is the use of private scope in templates. For now I'll stick with protected as we had already been doing and leave built-in TS on.

runspired:
I lost most code-hinting and all hover-for-code-doc abilities.

dfreeman:
> I lost most code-hinting 

Can you give examples of what you mean by code hinting?

The main thing we know we drop on the floor today is what the LSP refers to as "code actions," either affiliated with a completion item (like adding the needed import statement when you pick an identifier) or as standalone directions like cleaning up unused imports. If things that don't fall into that category are missing, that may be something we're unaware of or it may be a bug.

> all hover-for-code-doc abilities.

This sounds like a definite bug, independent of the other things outlined here. Can you file a separate issue with an example of what's failing? (Ideally with some code that I could drop into a Glint-enabled project and see the issue live)

dfreeman:
If anyone is interested adding support for translating the TypeScript language service's code actions (auto-importing new identifiers, removing unused imports, etc) to the LSP equivalent, I'm happy to provide what guidance I can.

dfreeman:
Note that @chadhietala did a massive chunk of work to support code general actions in #499, which is included in recent 1.x beta releases.

We're still missing support for auto-import as part of accepting a completion item (rather than as an after-the-fact independent action), however.

NullVoxPopuli:
Anyone want to implement this?

machty:
Quick 👍  / 👎 : if you're a VS Code user running into this, are you 👍 leaving TS enabled alongside Glint, or 👎 doing without the TS niceties (like auto-import)?

wagenet:
@machty There's a third option here, which is that you turn it on and off as you get annoyed with one or the other 😬

wagenet:
One of the big issues with leaving TS on is that it doesn't recognize .gts files.

spuxx1701:
[...]

PR #677, opened 2024-01-11
Auto Imports, remove JS/TS config

machty:
See https://github.com/typed-ember/glint/issues/130

##  Basic Support for Auto-Import (for all supported filetypes, including .gts)

You are now offered code completions for auto-imports, and selecting them will add the import at the top of the file.

## Remove JS/TS configuration options

Previously we were attempting to read the JS/TS VScode configuration preferences in order to pass them through to the VS LSP within Glint, but since the goal is feature parity with the TS LSP such that you can (and should) disable the default TS LSP, we can't continue to use the default JS/TS config options because they get disabled then you disable the TS built-in extension. This PR replaces them with sensible hard-wired defaults, and we can consider adding configurable VS code preferences specific to Glint in a future PR.

## Add localhost dev/debug instructions to CONTRIBUTING.md

This would have been very helpful for me getting up and running.

### (Won't fix) Completions don't work on unclosed tags within `<template>`

Likely not fixable until we have some syntax-forgiving Glimmer parser.


machty:
cc @NullVoxPopuli @dfreeman

machty:
Update: got auto-complete working in both .ts and .gts with default TS engine disabled.

image

image

Going to spend some time unifying the text preview to make it clearer what kind of completion/import it is.

machty:
Pushed some updates that greatly improve the default formatting of auto imports.

Screenshot 2024-01-12 at 11 36 24 AM
Screenshot 2024-01-12 at 11 36 30 AM

CONTRIBUTING.md
```diff
@@ -17,3 +17,43 @@ Glint is a family of packages which all live in this repo as a Yarn workspace. T
 - Read the project’s [ARCHITECTURE.md](./ARCHITECTURE.md) to understand the basics of how the code base works.
 
 Once you have made changes and added tests to confirm they work correctly, you can then open a PR and we'll work with you to polish it up and get it landed!
+
```
NullVoxPopuli, line 20: Thanks a ton for adding this! This is all the information i got super stuck on not having when i was first getting in to glint! 

Very needed!

packages/core/__tests__/language-server/completions.test.ts
NullVoxPopuli, line 1: Thanks a ton for adding tests!!

packages/core/src/language-server/glint-language-server.ts
```diff
@@ -88,12 +96,36 @@ export default class GlintLanguageServer {
       directoryExists: this.ts.sys.directoryExists,
       getDirectories: this.ts.sys.getDirectories,
       realpath: this.ts.sys.realpath,
+      useCaseSensitiveFileNames: () => true,
+
+      // @ts-ignore Undocumented method.
+      getCachedExportInfoMap() {
+        // This hook is required so that when resolving a completion item, we can fetch export info
```
NullVoxPopuli, line 103: Thanks for adding nome background here! It's immensely helpful for folks to have inline context like this <3 💪

packages/core/src/language-server/binding.ts
```diff
@@ -141,8 +123,8 @@ export function bindLanguageServerPool({
           kind,
           range,
           diagnostics,
-          formating,
```
NullVoxPopuli, line 144: Were these old apis? Or were whey wrong before?

They didn't seem to be 'our' properties, as i couldn't find them, but maybe i just missed them

machty: Not totally sure what you're asking, but these formatting and preference hashes are read from the VSCode configuration for the built-in TS/JS language server, but once you disable that (for "takeover" mode where Glint does everything) these hashes become empty, so I'm overriding with a hard-wired set of defaults.

NullVoxPopuli: I just mean that it _looks_ sus that previously the key of `formatting` used and changing ut to all uppercase `FORMATTING` makes things work. Like, i assume then that `formatting` was always incorrect.

NullVoxPopuli: Oh nevermind, i thought this was object shorthand, rather than positional argn

machty: (Discussed on Discord:) these are positional params, not an options hash.

NullVoxPopuli (approved):

wagenet (approved):

spuxx1701:
This is great! Thanks for adding this <3

CONTRIBUTING.md
```diff
@@ -17,3 +17,69 @@ Glint is a family of packages which all live in this repo as a Yarn workspace. T
...
+       .templateForBackingValue(this, function(𝚪, χ: typeof import("@glint/environment-ember-loose/-private/dsl")) {
+  {
+    const 𝛄 = χ.emitComponent(χ.resolve(χ.Globals["FooComponent"])
+      ({ desc: "notchHeight _ footer", ...χ.NamedArgsMarker }));
+    𝛄;
+    {
+      const [] = 𝛄.blockParams["default"];
+      {
+  // ...
+```
```
MrChocolatine, line 77: Using these Greek letters makes the code/documentation less readable.
Why deviate from what is normally used.

machty: At the time of writing, these greek letters were already being used by Glint's "compiler" for transforming templates into type-checkable TS. Note that in the last year I changed these symbols to keywords like `__glintDSL` because I ran into upstream issues during the sourcemapping process due to the multi-byte unicode encoding for these characters.

PR #696, opened 2024-01-29
use new glimmer ElementNode sub nodes

patricklx:
there are now also following nodes:
* name node: ElementNameNode
* start tag node: ElementStartNode

example:
```js
/*
  <Foo.bar.x attr='2'></Foo.bar.x>
   ^-- Element PathExpression[0] node
       ^-- Element PathExpression[1] node
          ^- Element PathExpression[2] node
   ^-------- Element PathExpression node
  ^------------------ Element openTag span
                      ^----------- ElementEndNode
 */
```
afterwards completions/validations can be add for attributes and element tags
In #663

### DIFF
--- a/packages/core/__tests__/transform/template-to-typescript.test.ts
+++ b/packages/core/__tests__/transform/template-to-typescript.test.ts
@@ -821,6 +821,9 @@ describe('Transform: rewriteTemplate', () => {
       expect(templateBody(template, { globals: [] })).toMatchInlineSnapshot(`
         "{
         const __glintY__ = __glintDSL__.emitElement("div");
+        __glintDSL__.applyAttributes(__glintY__.element, {
+         
+        });
         __glintDSL__.applyModifier(__glintDSL__.resolve(modifier)(__glintY__.element, { foo: "bar" , ...__glintDSL__.NamedArgsMarker }));
         }"
       `);
@@ -832,6 +835,9 @@ describe('Transform: rewriteTemplate', () => {
       expect(templateBody(template, { globals: [] })).toMatchInlineSnapshot(`
         "{
         const __glintY__ = __glintDSL__.emitComponent(__glintDSL__.resolve(MyComponent)());
+        __glintDSL__.applyAttributes(__glintY__.element, {
+         
+        });
         __glintDSL__.applyModifier(__glintDSL__.resolve(modifier)(__glintY__.element, { foo: "bar" , ...__glintDSL__.NamedArgsMarker }));
         }"
       `);
@@ -973,6 +979,9 @@ describe('Transform: rewriteTemplate', () => {
         "{
         const __glintY__ = __glintDSL__.emitElement("div");
         __glintDSL__.applySplattributes(__glintRef__.element, __glintY__.element);
+        __glintDSL__.applyAttributes(__glintY__.element, {
+         
+        });
         }"
       `);
     });
@@ -1017,6 +1026,9 @@ describe('Transform: rewriteTemplate', () => {
         "{
         const __glintY__ = __glintDSL__.emitComponent(__glintDSL__.resolve(Foo)());
         __glintDSL__.applySplattributes(__glintRef__.element, __glintY__.element);
+        __glintDSL__.applyAttributes(__glintY__.element, {
+         
+        });
         }"
       `);
     });
--- a/packages/core/src/transform/diagnostics/augmentation.ts
+++ b/packages/core/src/transform/diagnostics/augmentation.ts
@@ -235,7 +235,7 @@ function checkImplicitAnyError(
     // This error may appear either on `<Foo />` or `{{foo}}`/`(foo)`
     let globalName =
       sourceNode.type === 'ElementNode'
-        ? sourceNode.tag.split('.')[0]
+        ? sourceNode.path.head.original
         : sourceNode.type === 'PathExpression' && sourceNode.head.type === 'VarHead'
           ? sourceNode.head.name
           : null;
--- a/packages/core/src/transform/template/map-template-contents.ts
+++ b/packages/core/src/transform/template/map-template-contents.ts
@@ -27,7 +27,7 @@ export type Mapper = {
    * Given a @glimmer/syntax AST node, returns the corresponding start
    * and end offsets of that node in the original source.
    */
-  rangeForNode: (node: AST.Node) => Range;
+  rangeForNode: (node: AST.Node, span?: AST.Node['loc']) => Range;
 
   /**
    * Captures the existence of a directive specified by the given source
@@ -81,6 +81,7 @@ export type Mapper = {
    * corresponding to the given AST node in the original source.
    */
   forNode(node: AST.Node, callback: () => void, codeFeaturesForNode?: CodeInformation): void;
+  forNodeWithSpan(node: AST.Node, span: AST.Node['loc'], callback: () => void, codeFeaturesForNode?: CodeInformation): void;
 
   /**
    * This needs to be called after any node that "consumes" a `glint-expect-error` directive.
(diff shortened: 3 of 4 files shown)


test-packages/ts-ember-app/app/components/foo.hbs
NullVoxPopuli, line 1: there are a lot of noisy changes here -- can you reduce the diff to what was meaningful? thanks!

patricklx:
@NullVoxPopuli any idea how I can fix this:
```
Error: node_modules/@glimmer/interfaces/index.d.ts(1,2[9](https://github.com/typed-ember/glint/actions/runs/7698348329/job/20977417907?pr=696#step:5:10)): error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
Error: node_modules/@glimmer/interfaces/index.d.ts(3,15): error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
Error: node_modules/@glimmer/interfaces/index.d.ts(4,15): error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
Error: node_modules/@glimmer/interfaces/index.d.ts(5,15): error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
Error: node_modules/@glimmer/interfaces/index.d.ts(6,15): error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
Error: node_modules/@glimmer/interfaces/index.d.ts(7,15): error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.
```

NullVoxPopuli:
Does skipLibCheck help?

packages/core/__tests__/transform/template-to-typescript.test.ts
NullVoxPopuli, line 1: what's up with these changes? :sweat_smile:

patricklx: i'm not sure... i just run vite --update & prettier

patricklx: ah, wait, I did add a space in case its empty attributes

tsconfig.compileroptions.json
```diff
@@ -11,6 +11,7 @@
     "composite": true,
     "declaration": true,
     "sourceMap": true,
+    "skipLibCheck": true,
```
NullVoxPopuli, line 14: this may be controversial with the former ember-TS members, but I don't know of another way to have a node16 project interoperable with non-none16 projects

patricklx:
i'm not seeing and difference

NullVoxPopuli (approved):

tsconfig.compileroptions.json
ef4: I would prefer to fix this issue upstream in the glimmer side. `skipLibCheck` is a pretty blunt tool that can hide problems globally (since type failures in a library can generate `any`s that then propagate through our own code).

NullVoxPopuli: https://github.com/glimmerjs/glimmer-vm/pull/1555

NullVoxPopuli (changes requested):
Let's block on https://github.com/glimmerjs/glimmer-vm/pull/1555
And get that update in here, too once released

NullVoxPopuli:
@patricklx new versions published, can you try the .1 release?

patricklx:
nope, some new errors-...
```
Error: node_modules/@types/ember__helper/index.d.ts(3,81): error TS147[9](https://github.com/typed-ember/glint/actions/runs/7739973176/job/21103924609?pr=696#step:5:10): The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@glimmer/manager")' call instead.
Error: node_modules/@types/ember__helper/index.d.ts(4,53): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@glimmer/runtime")' call instead.
Error: node_modules/@glimmer/modifier/dist/commonjs/index.d.ts(1,20): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("@glimmer/runtime")' call instead.
```
https://github.com/microsoft/TypeScript/issues/52529
https://github.com/microsoft/TypeScript/pull/53426

looks like @ef4 knows something about this

NullVoxPopuli:
ah yeah, we're probably blocked on ember-source _needing_ to actually ship as a `type=module` package.

patricklx:
@NullVoxPopuli the error is under @types folder.
So we can make that type: module?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember__helper/package.json

NullVoxPopuli:
does changing all of `@types/ember*` to type=module work locally? :thinking:

patricklx:
> does changing all of `@types/ember*` to type=module work locally? 🤔

it fixes some errors,
but also needed to fix `@glimmerx/modifier` and `@glimmer/modifier`.

still, at the end it says
`node_modules/@types/ember__helper/index.d.ts:1:24 - error TS2307: Cannot find module 'ember/-private/type-utils' or its corresponding type declarations.`

NullVoxPopuli:
does Glint need to be ESM? :(

everything is borken

chancancode:
@patricklx FYI 0.89.0 is out with https://github.com/glimmerjs/glimmer-vm/pull/1568

Should also fix #706

packages/core/src/transform/diagnostics/augmentation.ts
```diff
@@ -201,8 +210,8 @@ function checkImplicitAnyError(
 
     // This error may appear either on `<Foo />` or `{{foo}}`/`(foo)`
     let globalName =
-      sourceNode.type === 'ElementNode'
-        ? sourceNode.tag.split('.')[0]
+      sourceNode.type === 'ElementNameNode'
+        ? sourceNode.value.split('.')[0]
         : sourceNode.type === 'PathExpression' && sourceNode.head.type === 'VarHead'
```
chancancode, line 215: This kind of stuff should now work better, since `ElementNode`, `MustachesStatement` and `SubExpression` all have the same `path` property

packages/template/__tests__/emit-content.test.ts
```diff
@@ -1,4 +1,5 @@
-import { SafeString } from '@glimmer/runtime';
+// @ts-expect-error cannot import, but we only need the type...
+import type { SafeString } from '@glimmer/runtime';
```
chancancode, line 2: Why did it stop working? It seems to be here: https://github.com/glimmerjs/glimmer-vm/blob/v0.89.0/packages/%40glimmer/runtime/index.ts#L43

And I don't think we would want `@ts-expect-error` here, if the only reason we want to import it is for the type, but specifically the type import wasn't working, what does this even do? (making it `any`, probably)

patricklx: This was the error:
https://github.com/typed-ember/glint/actions/runs/7712187814/job/21019163076#step:9:139

NullVoxPopuli:
@patricklx i think this is unblocked now?

patricklx:
re-created as https://github.com/typed-ember/glint/pull/839

PR #699, opened 2024-02-04
feat(vscode): add jsx and tsx to the list of watched extensions

poulet42:
👋 I have a usecase where I'm getting types from a `.tsx` file, glint errors does show up but does not get updated when the tsx types changes. This PR adds .tsx (and .jsx for consistency(?)) to the list of extensions to watch. 

I understand this is a rather unlikely usecase in a glint codebase but I'm trying my luck 🙏 . Another alternative would be to expose the list of extensions to watch as a setting but it's a bit more work

### DIFF
--- a/packages/vscode/src/extension.ts
+++ b/packages/vscode/src/extension.ts
@@ -21,7 +21,7 @@ import type { Request, GetIRRequest, SortImportsRequest } from '@glint/core/lsp-
 
 const outputChannel = window.createOutputChannel('Glint Language Server');
 const clients = new Map<string, LanguageClient>();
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
I think this should probably be configurable. While it's common for folks to mix in React with Ember, I don't want to give React special treatment, and would like to equally support extending this extension list. That way, if folks are using Ember with Vue, Svelte, etc, they have a path to do so

poulet42:
@NullVoxPopuli Thank you for your answer 🙏 I'm closing this MR in favour of a more generic "glint.watchExtensions" setting. I'll take some time to work on it

Issue #672, opened 2024-01-10
Glint doesn’t support multiple Typescript extends


andrewtimberlake:
The Ember Typescript docs suggest enabling TypeScript’s strictest support by adding `@tsconfig/strictest` before the ember config at https://guides.emberjs.com/release/typescript/additional-resources/faq/#toc_strictness

This breaks Glint as seen in #662

NullVoxPopuli:
unstable Glint (main) has this same problem

PR #701, opened 2024-02-20
Loading tsconfig with extends array

BarryThePenguin:
I put together a failing test for #672

I might have a go at putting together a fix, just wanted to make this available for others


NullVoxPopuli:
Thank you for your reproduction! this is resolved in glint @ v2 / alpha!!

PR #709, opened 2024-03-13
Fix project argument to match tsc

bendemboski:
The `--project` argument was being treated as a path to some location inside the project, rather than as either a path to a tsconfig file or a path to a folder containing a tsconfig file, which is how `tsc` implements it. This meant that there was no way to run `glint` and use a config file named anything other than `tsconfig.json`.

This is a breaking change because if anybody was passing a project path that pointed deeply into the project, that will no longer work. Similarly, the `analyzeProject()` unstable API's behavior has changed, and `loadConfig` export has been replaced with `loadClosestConfig` and `loadConfigFromProject` exports.

We could switch back to only having a single `loadConfig()` method that checks to see if its path points to a config file, and if not falls back on the pre-existing logic, but I thought that would be kinda confusing, and it would be better for the behavior to match that of `tsc` even though that requires a technically-breaking change.

### DIFF
--- a/packages/core/src/config/loader.ts
+++ b/packages/core/src/config/loader.ts
@@ -27,7 +27,7 @@ export class ConfigLoader {
    */
   public configForProjectPath(configPath: string): GlintConfig | null {
     let tsConfigPath = path.join(configPath, 'tsconfig.json');
-    let jsConfigPath = path.join(configPath, 'tsconfig.json');
+    let jsConfigPath = path.join(configPath, 'jsconfig.json');
 
     if (fileExists(configPath)) {
       return this.configForConfigFile(configPath);
--- a/packages/core/src/config/loader.ts
+++ b/packages/core/src/config/loader.ts
@@ -1,5 +1,6 @@
 import { createRequire } from 'node:module';
 import * as path from 'node:path';
+import * as fs from 'node:fs';
 import SilentError from 'silent-error';
 import { GlintConfig } from './config.js';
 import { GlintConfigInput } from '@glint/core/config-types';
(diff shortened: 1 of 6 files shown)


packages/core/src/config/loader.ts
```diff
@@ -10,32 +11,72 @@ const require = createRequire(import.meta.url);
...
 
+  /**
+   * Given the path to a configuration file, or to a folder containing a
+   * `tsconfig.json` or `jsconfig.json`, load the configuration. This is meant
+   * to implement the behavior of `glint`/`tsc`'s `--project` command-line
+   * option.
+   */
+  public configForProjectPath(configPath: string): GlintConfig | null {
+    let tsConfigPath = path.join(configPath, 'tsconfig.json');
+    let jsConfigPath = path.join(configPath, 'tsconfig.json');
```
dfreeman, line 30: Should this be `jsconfig.json`?

dfreeman (commented):
Thank you for fixing this!

packages/core/src/index.ts
```diff
@@ -40,6 +41,6 @@ export function analyzeProject(projectDirectory: string = process.cwd()): Projec
   };
 }
 
-export { loadConfig };
```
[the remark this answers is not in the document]
bendemboski: Yeah, makes sense! `console.warn` or do we have another method for issuing deprecation warnings?

packages/core/src/config/loader.ts
bendemboski: 🤦 yes it should

bendemboski:
@dfreeman I've addressed your feedback in a separate commit that I'll squash into the first commit if it all looks good

NullVoxPopuli:
thanks for your efforts!

this is now fixed (in another PR) and released in v2/alpha.

You can see it in action in the v2 addon blueprint, here: https://github.com/ember-cli/ember-addon-blueprint/blob/main/files/rollup.config.mjs#L66

PR #713, opened 2024-03-18
check if template is in heritage clause

patricklx:
support 
```gjs
class A extends X(<template></template>) {

}
```

currently the template part tries to have `A` as backing value

### DIFF
--- a/packages/core/__tests__/transform/rewrite.test.ts
+++ b/packages/core/__tests__/transform/rewrite.test.ts
@@ -467,6 +467,30 @@ describe('Transform: rewriteModule', () => {
   });
 
   describe('ember-template-imports', () => {
+    test('in class extends', () => {
+      let customEnv = GlintEnvironment.load(['ember-loose', 'ember-template-imports']);
+      let script = {
+        filename: 'test.gts',
+        contents: stripIndent`
+          import Component, { hbs } from 'special/component';
+          export default class MyComponent extends Component(<template></template>) {
+            
+          }
+        `,
+      };
+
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Any chance for a test? <3

patricklx:
@NullVoxPopuli added a test

NullVoxPopuli (approved):

patricklx:
@NullVoxPopuli any chance to release this?

NullVoxPopuli:
it "should be" on the `unstable` releases rn.

if you want it released on the 1.x series, can you PR to the 1.x branch? https://github.com/typed-ember/glint/tree/v1.x

apologies -- the timing of this merge + the last release + the beginning of the volar stuff was a little cray <3

machty:
I'm having trouble understanding what this syntax does, could you explain?

```gts
class A extends X(<template></template>) {

}
```

What is `X` here? What would you put an embedded template within the heritage clause?

NullVoxPopuli:
I wouldn't expect `X` should matter, rather than _it's a function who's first argument receives a component_ (similar to ember-route-template), and it must return a class (so that A can extend from it)

machty:
TIL [ember-route-template](https://github.com/discourse/ember-route-template) -- is it fair to say this library is the main use cases / proponent for this syntax at this time?

NullVoxPopuli:
It's probably the most popular use case, yea. But at the same time,

We support this:
```js
const x = <template>foo</template>

doSomething(x);
```

Why not support support someone wanting to inline it?
```js
doSomething(<template>foo</template>);
```

PR #716, opened 2024-03-19
Remove GlimmerX

machty:
https://github.com/emberjs/rfcs/pull/1014

Removes any reference of GlimmerX from Glint.

Justification from the RFC:

> most people don't even know what GlimmerX is, it's not under active development, it doesn't play nicely with our plans to use TS Plugins. Therefore, we should remove it from Glint before proceeding further with the proposals in this RFC.

## Notes

- It seems that a lot of the ideas that wound up in ember-template-imports (i.e. `<template>` tags within `.gts` files) were first tested out via inline `hbs` template tags in the glimmerx environment; in order to not lose test coverage, I did my best to convert these glimmerx test to .gts ember-template-import tests rather than wholesale removing them.


NullVoxPopuli:
Before merge, we should probably release the current stuff on main

wagenet (approved):

Issue #720, opened 2024-05-17
If route is named `components` model types are not inferred 


aklkv:
If we have structure like:

```
templates/
  components/
    text.hbs
routes/
  components/
    text.ts
 ```
 
 and the content of the `route.ts`:
 
 ```
 import Route from '@ember/routing/route';

export default class ComponentsTextRoute extends Route {
  model() {
    return {
      message: 'Components Text Route',
    };
  }
}
```

we get the following glint error:

<img width="650" alt="Screenshot 2024-05-17 at 12 17 09 PM" src="https://github.com/typed-ember/glint/assets/312210/a0b3e754-bb71-4881-ab5b-2755e026b2b2">

Anything outside of `template/components` folders works just fine.
I am guessing there might be some naming clash.

I can also publish repro if needed

NullVoxPopuli:
this is an extra troll as well, because `templates/components` was the folder used for all component hbs files pre-Octane.

It'd be good to drop support for that pattern entirely in Glint so that this bug can be fixed.

ef4:
I would be surprised if glint is the only thing that breaks when using a route named "components".

We're planning to drop support for components putting their template files into `templates/components/*.hbs` at ember 6.0, but before that point components very much might resolve template files there. In your example, if you happen to have a component named "text", ember will probably try to resolve your `templates/components/text.hbs` as its template.

It would be good to do a review across the whole stack to see what still treats `templates/components` as special, since those are all places that can get cleaned up at ember 6.0.

PR #721, opened 2024-05-18
feat: drop support for classic component layout

aklkv:
fixes: #720


NullVoxPopuli (approved):

packages/environment-ember-loose/__tests__/environment.test.ts
```diff
@@ -69,14 +69,6 @@ describe('Environments: Ember Loose', () => {
       '/route.js',
     ]);
 
-    // Classic component
```
aklkv, line 72: Should we keep a parts of this test or drop it all together 🤔

NullVoxPopuli: We should drop everything related

NullVoxPopuli:
yeah, I mostly want to see where @machty is at with volar stuff, because if we're going to introduce some churn, I'd prefer we do it once with the volar stuff, and then put the v1.4 in bugfix-only mode its own branch (which I guess could happen now if folks have non-breaking bugfixes to PR)

ef4:
If this was green we would consider merging it.

machty:
So long as this only removes support for `templates/components` and NOT co-located templates, seems good to me.

FYI Volar will initially support Polaris .gts-only projects and then add support for two-file components, at which point we'll only support co-located components and not reinstate `templates/components`.

machty:
Merging this so i can work atop it as i reinstate loose mode in Volar.

PR #723, opened 2024-05-27
Add template import type for uniqueId helper

vstefanovic97:
This in combination with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69687 should make importing `uniqueId` from `@ember/helper` work


NullVoxPopuli:
Why does glint need code changes at all?

vstefanovic97:
> Why does glint need code changes at all?

@NullVoxPopuli as far as I can tell the types defined in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69687/files#diff-ee310ca708dd6726cf0487230fae7935b0eecd94c32c37e022b0776a82506bb7R15
don't have any real type information and we rely on extending the interfaces here in glint

NullVoxPopuli:
We may need to start a 1.x branch, because main is currently breaking, preparing for v2 / a new approach to glint

NullVoxPopuli:
This'll be in the main/2.x release (which may be unstable for a while).
(looks like you have lint issues though)

We can release this in 1.x if you also make this PR for here: https://github.com/typed-ember/glint/tree/v1.x

vstefanovic97:
@NullVoxPopuli I rebased it and repointed to the `v1.x` branch

NullVoxPopuli:
so sorry to do this, but glint v2 is presently only going to be focused on gjs/gts, which means that we only need be able to use the types from `@ember/helper` already -- however, those are also broken -- I've opened this issue to track: https://github.com/emberjs/ember.js/issues/20947

PR #725, opened 2024-06-17
docs: use gjs as codefence lang

IgnaceMaes:
Glimmer JS/TS syntax highlighting support was added to GitHub a couple of months ago. This PR uses that lang for codefences where `<template>` tag is being used.

### DIFF
--- a/docs/ember/template-imports.md
+++ b/docs/ember/template-imports.md
@@ -49,7 +49,7 @@ When using `ember-template-imports`, you can declare the type of a `<template>` 
 
 {% code title="app/components/shout.gts %}
 
-```typescript
+```gts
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
gjs works:

```gjs
const two = 2;

<template>
  {{two}}
</template>
```

gts:

```gts
const two = 2;

<template>
  {{two}}
</template>
```

glimmer-ts:

```glimmer-ts
const two = 2;

<template>
  {{two}}
</template>
```

did we misconfigure something in linguist?

NullVoxPopuli:
hmm. that's silly

NullVoxPopuli (approved):

PR #749, opened 2024-07-09
Primitives for supporting ember-loose mode (.ts + .hbs)

machty:
Starts in on support for ember-loose mode, the classic Ember component authoring format when you provide (optional) backing .ts/.js class with a separate .hbs template file.

## Learnings / Takeways:

TL;DR, we can't follow support loose mode until we finish migrating to TS Plugin. Next steps are to find out how much of a lift that'll be.

- Volar only recently added support for two-file templating system (initially added for Angular-Webstorm; Angular has .ts files with .html template files)
- Volar's support only applies to TS Plugins, and does not currently offer support for two files in the (quickly-becoming-legacy) Language Server APIs
  - This means Glint will need to convert to using TS Plugin before we can support this
  - Alternatively Johnson from Volar might possibly be convinced to add support to the LS, but I think if we can get TS Plugin working, better to align with the future.
- Basic Volar scheme for two-file component systems:
  - In order to typecheck an .hbs file, you need to regard the backing `.ts` file as Volar's "entrypoint", in that you need to provide a VirtualCode for .ts files and teach those files how to find the "associated" .hbs template file. If you correctly/successfully do that, then when you attempt to typecheck the .hbs file, Volar will already know that it has been associated with the .ts file that it'll use as part of the transformation.
    - Note this doesn't cover the case of standalone .hbs files, which Ember classically supports
      - It is TBD whether we'll try to support this vs some other (hopefully scriptable) migration method, but if we do:
      - We would probably have to implement another HandlebarsVirtualCode and rely on Volar's internal algorithm to use the TS+associated HBS transformation/type-checking algorithm to "win" over the fallback HandlebarsVirtualCode used for standalone templates.

### DIFF
--- a/packages/core/__tests__/cli/check.test.ts
+++ b/packages/core/__tests__/cli/check.test.ts
@@ -176,7 +176,7 @@ describe('CLI: single-pass typechecking', () => {
     `);
   });
 
-  test.only('reports diagnostics for a companion template type error', async () => {
+  test.skip('reports diagnostics for a companion template type error', async () => {
     project.setGlintConfig({ environment: 'ember-loose' });
 
     let script = stripIndent`
--- a/packages/core/__tests__/cli/check.test.ts
+++ b/packages/core/__tests__/cli/check.test.ts
@@ -176,7 +176,7 @@ describe('CLI: single-pass typechecking', () => {
     `);
   });
 
-  test('reports diagnostics for a companion template type error', async () => {
+  test.only('reports diagnostics for a companion template type error', async () => {
     project.setGlintConfig({ environment: 'ember-loose' });
 
     let script = stripIndent`
--- a/packages/core/__tests__/cli/check.test.ts
+++ b/packages/core/__tests__/cli/check.test.ts
@@ -176,7 +176,7 @@ describe('CLI: single-pass typechecking', () => {
     `);
   });
 
-  test.skip('reports diagnostics for a companion template type error', async () => {
+  test.only('reports diagnostics for a companion template type error', async () => {
     project.setGlintConfig({ environment: 'ember-loose' });
 
     let script = stripIndent`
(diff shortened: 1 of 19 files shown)


packages/core/__tests__/cli/check.test.ts
```diff
@@ -200,9 +200,8 @@ describe('CLI: single-pass typechecking', () => {
 
     let checkResult = await project.check({ reject: false });
 
-    expect(checkResult.exitCode).toBe(1);
-    expect(checkResult.stdout).toEqual('');
-    expect(stripAnsi(checkResult.stderr)).toMatchInlineSnapshot(`
+    expect(checkResult.exitCode).not.toBe(0);
+    expect(stripAnsi(checkResult.stdout)).toMatchInlineSnapshot(`
```
machty, line 204: This test currently failing (diagnostics returning empty array `[]`)

packages/core/__tests__/cli/check.test.ts
```diff
@@ -176,7 +176,7 @@ describe('CLI: single-pass typechecking', () => {
     `);
   });
 
-  test.skip('reports diagnostics for a companion template type error', async () => {
+  test.only('reports diagnostics for a companion template type error', async () => {
```
machty, line 179: This test is failing for strange reasons, but might be specific to CLI tests, so not focusing on it for now. Root cause seems to be that the ember component `.ts` file is not ever getting passed to `createVirtualCode` the language plugin.

packages/core/__tests__/language-server/diagnostic-augmentation.test.ts
```diff
@@ -940,7 +940,8 @@ describe('Language Server: Diagnostic Augmentation', () => {
     });
 
     let server = await project.startLanguageServer();
-    let diagnostics = server.getDiagnostics(project.fileURI('index.hbs'));
+    const { uri } = await server.openTextDocument(project.filePath('index.hbs'), 'handlebars');
+    let diagnostics = await server.sendDocumentDiagnosticRequest(uri);
 
     expect(diagnostics.items.reverse()).toMatchInlineSnapshot(`
```
machty, line 946: This test is failing because diagnostics returning empty array `[]`

PR #781, opened 2024-12-18
Resolve config from extends package

iarroyo:
Resolve glint configuration when your tsconfig.json extends from a package, ex

path: tsconfig.json
```json
{
  "extends": "@package1"
}
```

path: node_modules/@package1/tsconfig.json
```json
{
  "glint": {
    "environment": ["ember-loose", "ember-template-imports"]
  }
}
```

**Note**: I'm uncertain if this is the correct branch to merge with. If it isn't, please let me know.

### DIFF
--- a/packages/core/__tests__/config/load-config.test.ts
+++ b/packages/core/__tests__/config/load-config.test.ts
@@ -1,8 +1,9 @@
 import * as fs from 'node:fs';
 import * as os from 'node:os';
-import { describe, beforeEach, afterEach, test, expect } from 'vitest';
+import { describe, beforeEach, afterEach, test, expect, vi } from 'vitest';
 import { loadConfig } from '../../src/config/index.js';
 import { normalizePath } from '../../src/config/config.js';
+import { require } from '../../src/config/loader.js';
 
 describe('Config: loadConfig', () => {
   const testDir = `${os.tmpdir()}/glint-config-test-load-config-${process.pid}`;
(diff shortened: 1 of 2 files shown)


packages/core/src/config/loader.ts
```diff
@@ -1,11 +1,12 @@
 import { createRequire } from 'node:module';
 import * as path from 'node:path';
+import * as fs from 'node:fs';
 import SilentError from 'silent-error';
 import { GlintConfig } from './config.js';
 import { GlintConfigInput } from '@glint/core/config-types';
 import type * as TS from 'typescript';
 
-const require = createRequire(import.meta.url);
+export const require = createRequire(import.meta.url);
```
NullVoxPopuli, line 9: why export this? 
conventionally, we really only want to use createRequire in the file it's used in, since the resolution rules are specific to the import.meta.url

packages/core/__tests__/config/load-config.test.ts
```diff
@@ -51,4 +52,49 @@ describe('Config: loadConfig', () => {
...
+          checkStandaloneTemplates: false,
+        },
+      })
+    );
+    fs.writeFileSync(
+      `${directory}/tsconfig.json`,
+      JSON.stringify({
+        extends: '@package1/tsconfig.json',
+        glint: {
+          environment: '../local-env',
```
NullVoxPopuli, line 89: what's local-env?

NullVoxPopuli (commented):
This is looking good -- left some questions.
Also, would you be willing to PR this to the main branch as well?

packages/core/src/config/loader.ts
NullVoxPopuli: ah, it's for mocking in tests.

can you add a block-comment with something like:
```js
/**
 * @private 
 *
 * Only exported for testing purposes. Do not import.
 */
```

packages/core/__tests__/config/load-config.test.ts
iarroyo: To be honest, I simply followed the example from the previous test which uses a predefined environment in the [beforeEach](https://github.com/typed-ember/glint/blob/53f68241d6f4224fceb8b318f2566e5eb9c9a703/packages/core/__tests__/config/load-config.test.ts#L14)

packages/core/src/config/loader.ts
iarroyo: **Done**

iarroyo:
> This is looking good -- left some questions. Also, would you be willing to PR this to the main branch as well?

Definitely, I will create a PR against the main branch

iarroyo:
PR Created
https://github.com/typed-ember/glint/pull/782

packages/core/src/config/loader.ts
```diff
@@ -85,7 +85,11 @@ function loadConfigInput(ts: TypeScript, entryPath: string): GlintConfigInput |
     if (currentContents.extends) {
       currentPath = path.resolve(path.dirname(currentPath), currentContents.extends);
       if (!fs.existsSync(currentPath)) {
-        currentPath = require.resolve(currentContents.extends);
+        try {
+          currentPath = require.resolve(currentContents.extends);
+        } catch (error) {
+          // control the exception, do nothing.
```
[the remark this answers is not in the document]
iarroyo: If I log anything this [test](https://github.com/typed-ember/glint/blob/v1.x/packages/core/__tests__/cli/build.test.ts#L1151) and others will fail

Please refer to the image below that shows the error being displayed.
Screenshot 2025-01-08 at 15 34 49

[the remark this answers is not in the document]
NullVoxPopuli: ahh ok. thanks!

NullVoxPopuli:
thanks for this! do you want to make the same changes on main as well?

iarroyo:
> thanks for this! do you want to make the same changes on main as well?

Yep, let's make sure that the exception is not propagated.
I´ll add a try/catch

PR #784, opened 2025-01-08
Upgrade to Volar 2.4.11

machty:


package.json
```diff
@@ -40,7 +40,7 @@
     "eslint": "^8.27.0",
     "prettier": "^3.3.2",
     "release-it": "^15.5.0",
-    "typescript": ">=5.4.0"
+    "typescript": ">=5.6.0"
```
machty, line 43: Volar hugs TS pretty closely; this was necessary due to output lib code referencing SetIterator which is new to TS 5.6. Perhaps at some point we might need some upstream tweaks to make a little room for older Ember apps that are tied to older TS?

packages/core/__tests__/cli/build.test.ts
```diff
@@ -466,8 +466,13 @@ describe('CLI: single-pass build mode typechecking', () => {
               2 const A = 2 * C;
                               ~
 
+              src/index.gts:2:15 - error TS2307: Cannot find module '@glint-test/a' or its corresponding type declarations.
```
machty, line 469: Since the TS/Volar upgrade, there seems to be some internal changes in the build algorithm, producing more error messages than before if one of the subprojects fails to build.

packages/core/__tests__/cli/build.test.ts
```diff
@@ -1614,7 +1664,7 @@ describe('CLI: --build --dry', () => {
           let buildResult = await projects.root.buildDeclaration({ flags: ['--dry'] });
           expect(buildResult.exitCode).toBe(0);
           expect(stripAnsi(buildResult.stdout)).toMatch(
-            `A non-dry build would build project '${projects.main.filePath('tsconfig.json')}'`,
+            ` A non-dry build would update timestamps for output of project '${projects.main.filePath('tsconfig.json')}'`,
```
machty, line 1667: This also seems to be a change/optimization within TSC that's smart enough to realize root project would have no changed output and hence would only update timestamps (?)

packages/core/__tests__/language-server/diagnostic-augmentation.test.ts
```diff
@@ -583,7 +583,24 @@ describe('Language Server: Diagnostic Augmentation', () => {
...
+                  "start": {
+                    "character": 69,
+                    "line": 18,
+                  },
+                },
+                "uri": "file:///PATH_TO_MODULE/@glint/template/-private/integration.d.ts",
+              },
+              "message": "'[InvokeDirect]' is declared here.",
+            },
+          ],
```
machty, line 603: This additional snapshotted data just means that Volar has since enhanced the metadata provided on each diagnostic, e.g. you should start to see more "SYMBOL is declared here" messages with links to the declaration source.

packages/core/__tests__/cli/watch.test.ts
```diff
@@ -164,8 +164,6 @@ describe('CLI: watched typechecking', () => {
     output = await watch.awaitOutput('Watching for file changes.');
     expect(output).toMatch('Found 0 errors.');
 
-    expect(watch.allOutput).toMatch(/\033c/); // output should include screen reset control sequence
```
machty, line 167: removed this and the test below because the tests were breaking but likely due to an artifact with tty / CI. I checked that `preserveWatchOutput` still works (and it does); I figure we don't need to double-test stuff that Volar already kind of guarantees for us.

NullVoxPopuli (approved):

PR #786, opened 2025-01-08
Clarify typing for no arguments/blocks

elwayman02:
Update verbiage and add code example to make it clear how to handle scenarios where a component has no arguments or no block invocation

### DIFF
--- a/docs/ember/component-signatures.md
+++ b/docs/ember/component-signatures.md
@@ -75,9 +75,7 @@ import Component from '@glimmer/component';
 export interface SimpleHelloSignature {
   // We have a `<div>` as our root element
   Element: HTMLDivElement;
-  // We accept no arguments
-  Args: Record<string, never>;
-  // We do not accept block form, so don't specify it in the signature
+  // We accept no arguments or block form, so don't specify them in the signature
 }
 
 export default class SimpleHello extends Component<SimpleHelloSignature> {}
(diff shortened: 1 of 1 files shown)


docs/ember/component-signatures.md
```diff
@@ -67,6 +67,32 @@ export default class SuperTable<T> extends Component<SuperTableSignature<T>> {}
...
+{% code title="app/components/simple-hello.ts" %}
+
+```typescript
+import Component from '@glimmer/component';
+
+export interface SimpleHelloSignature {
+  // We have a `<div>` as our root element
+  Element: HTMLDivElement;
+  // We accept no arguments
+  Args: Record<string, never>;
```
bendemboski, line 79: I think I would omit `Args` here as well. Including it but not `Blocks` seems confusing as the two have pretty much identical usages (aside from their contents of course), and as far as I know you never need to do this -- the scenario we discussed over discord is where you also override the constructor (not the most common/simplest case), in which case you can just do

```ts
interface SimpleHelloSignature {
  Element: HTMLDivElement;
}

export default class SimpleHello extends Component<SimpleHelloSignature> {
  constructor(owner: Owner, args: Record<string, never>) {
    super(owner, args);
  }
}
```

elwayman02: Seems fine to me

NullVoxPopuli (approved):

NullVoxPopuli:
thank you!!

PR #811, opened 2025-03-07
Migrate tests to TS Plugin mode, remove option for old LS mode

machty:
In a recent [PR](https://github.com/vuejs/language-tools/pull/5248), Vue tooling has:

1. Removed their support between toggling between classic LS-driven typechecking (aka takeover mode) and the new TS Plugin driven diagnostics
2. Provided examples for how to test the TS Plugin directly via tsserver harness

The timing is good for me to remove these features in Glint and to re-express old LS tests as TS Plugin instead. Otherwise we would experience continued awkward fracturing in Glint's codebase due to the way to way express certain TS tests in the old (and soon to be unused) LS mode, but then have/had no way to test TS Plugin-only stuff (e.g. hbs+ts loose mode two file components) other than to use slow VSCode acceptance tests.


packages/core/__tests__/language-server/references-ts-plugin.test.ts
```diff
@@ -0,0 +1,230 @@
...
+        }
+      `,
+    );
+
+    expect(
+      await performReferencesRequest(
+        greetingDoc,
+        offset,
+      ),
+    ).toMatchInlineSnapshot(`
```
machty, line 60: This is wrong; there should be 4 references total, but 2 references are missing from `ephemeral-app.gts`. I've tested in F5 debug mode and it seems to be working; wondering if there is something strange about the test setup.

machty: Fixed via on-disk empty-fixture.gts

johanrd:
@machty FWIW: When upgrading from `1.4.1-unstable.130d85f` (The last published version before this PR) to `1.4.1-unstable.45026b1` (this PR), Glint stopped working with the following error message:

```Request initialize failed with message: typescript.tsdk is required```

Screenshot 2025-03-18 at 09 21 21


– Are there perhaps another set of dependencies required after this PR? thanks!

machty:
@johanrd we have a testable TS Plugin mode available from VSCode Marketplace; try the pre-release version 1.4.3+.

See here for more details: https://github.com/typed-ember/glint/blob/main/GLINT_V2.md

PR #816, opened 2025-03-15
Allow each-in to work on `Map`

### LINKED ISSUE #645: {{#each-in}} breaks if used with a Map
The value of `{{#each-in}}` has an unexpected type if being used with a `Map`. Even trying to render the value in the template throws an error.

Simple reproduction:

```ts

export default class PlaygroundComponent extends Component {
  map: Map<string, number> = new Map([
    ['a', 1],
    ['b', 2],
    ['c', 3],
  ]);
```

```hbs
{{#each-in this.map as |key value|}}
  {{key}}
  {{value}}
{{/each-in}}
```

```
app/components/playground.hbs:3:3 - error TS2345: Only primitive values and certain DOM objects (see `ContentValue` in `@glint/template`) are usable as top-level template content.
  Argument of type 'number | ((callbackfn: (value: number, key: string, map: Map<string, number>) => void, thisArg?: any) => void) | (() => IterableIterator<[string, number]>) | ... 6 more ... | ((key: string) => boolean)' is not assignable to parameter of type 'ContentValue'.
    Type '(callbackfn: (value: number, key: string, map: Map<string, number>) => void, thisArg?: any) => void' is not assignable to type 'ContentValue'.

3   {{value}}
    ~~~~~~~~~
```

Please note that it is working fine if using a POJO instead of a `Map`.

typescript: 5.2.2
@glint/template: 1.2.1

NullVoxPopuli:
from [Devin](https://app.devin.ai/sessions/afe0bdb5d2824d6595d3e077e38e4233)

Resolves #645

### DIFF
--- a/packages/template/-private/index.d.ts
+++ b/packages/template/-private/index.d.ts
@@ -35,6 +35,7 @@ export type ContentValue =
   | void
   | SafeString
   | Node
+  | Map<any, any>
   | ArglessCurlyComponent;
 
 // This encompasses both @glimmer/runtime and @ember/template's notion of `SafeString`s,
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Better implementation here: https://github.com/typed-ember/glint/pull/826

PR #820, opened 2025-03-15
Drop support for @types/ember* only use native/built-in types

NullVoxPopuli:
Unblocks: https://github.com/typed-ember/glint/pull/818


NullVoxPopuli:
The one failing test exists on main

packages/core/src/transform/template/template-to-typescript.ts
```diff
@@ -20,6 +20,22 @@ export type TemplateToTypescriptOptions = {
   specialForms?: Record<string, GlintSpecialForm>;
 };
 
+/**
+ * NOTE: this is tech debt. Tho solving it will require more work than polyfilling old behavior of path.parts
```
machty, line 24: why / what specifically is tech debt here?

machty (commented):
Looks good just think one comment could clarify

packages/core/src/transform/template/template-to-typescript.ts
NullVoxPopuli: I'm adapting an old behavior for the sake of easier PR review, rather than inlining needed changes, and propagating the impact of those changes.

machty: ok sounds good

Issue #824, opened 2025-03-21
Glint doesn't understand implied "export default <template>"


ef4:
A `<template>` tag by itself at module scope gets an implied `export default`, per the original template tag RFC.

Until very recently, the implementation  in `content-tag` didn't do this correctly for typescript, but that bug is now fixed and the actual runtime behavior is fully working.

But Glint doesn't infer the `export default`, so your component is not importable into other TS or GTS files.

PR #825, opened 2025-03-21
Support implicit default export of template tags with satisfies keyword

mogstad:
Try resolving  #824


mogstad:
I think upgrading to content-tag v3 would solve the test issue, any reason not to upgrade it?

NullVoxPopuli:
We should use content-tag v3, but probably in a different PR.
My next batch of work in this repo is converting to pnpm so deps make sense, which may help us with the error received here, and figuring out the TS config incantation to resolve

PR #826, opened 2025-03-21
Explicitly accept an Iterable in `each-in`

mogstad:
This allows us to extract the generic arguments and use them as the types for the default block.

Resolves: #645


NullVoxPopuli (approved):
Yay!

PR #827, opened 2025-03-21
Update ARCHITECTURE and GLINT_V2 docs

machty:


bartocc (commented):
Thanks for these updated docs @machty!

PR #828, opened 2025-03-22
volar-service-typescript dep - replace NPM tag with version number

bartocc:
Yarn does not seem to understand the NPM tag volar-2.4

Would it be possible to use the version number instead?

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -47,7 +47,7 @@
     "semver": "^7.5.2",
     "silent-error": "^1.1.1",
     "uuid": "^8.3.2",
-    "volar-service-typescript": "volar-2.4",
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
We're using pnpm now, pnpm understands tags 🎉

bartocc (commented):
I know, I know ;-) but it would be too bad that those using Yarn could not try Glint V2.

Yarn, by the way, understands tags, but its regex is too restrictive, see
https://discord.com/channels/226791405589233664/1353033398049968320

For the time being, I am going to patch the regex in our vendored Yarn bin

bartocc:
I've rebased on main

PR #829, opened 2025-03-23
Working VSCode extension pre-release

machty:
This PR introduces a number of eslint and vsce tweaks to properly bundle the dependencies needed for pre-release Glint to work when downloaded from the VSCode marketplace.

Example of Glint in new TS Plugin mode kicking in in VSCode (.gts file):

image

Example of Glint in new TS Plugin mode kicking in in VSCode (.hbs file):

image

Version 1.4.3 available in VSCode marketplace for pre-release download.


package.json
```diff
@@ -14,7 +14,7 @@
     ]
   },
   "scripts": {
-    "reset": " git clean -Xfd; git clean -fd; pnpm i && pnpm build",
+    "reset": "git clean -Xfd; git clean -fd; pnpm i && pnpm build > /dev/null 2>&1; pnpm i -f; pnpm build; pnpm sync",
```
machty, line 17: @NullVoxPopuli had to reinstate your original reset script (main is broken without this fix too)

NullVoxPopuli: Hmm sus. I'll investigate

packages/vscode/src/extension.ts
```diff
@@ -234,6 +232,10 @@ try {
 
   const languageIdsQuoted = config.server.includeLanguages.map((lang) => `'${lang}'`).join(',');
 
+  // TODO: as of March 23 2025 and perhaps a few days before, this fs.readFileSync hacky
+  // is failing with "Cannot set property readFileSync of #<Object> which has only a getter".
+  // Stay tuned for another hack from Volar... or hopefully an upstream fix from VSCode.
```
machty, line 237: This is the no-longer-working hack described above.

PR #838, opened 2025-03-25
Support extending multiple config files

### LINKED ISSUE #672: Glint doesn’t support multiple Typescript extends

mogstad:
Since TS5, tsconfig.json `extends` supports being an array of config files to extend.

Resolved #672


packages/core/__tests__/config/load-config.test.ts
NullVoxPopuli, line 1: What happens in these new tests without the fix implemented?

I don't really understand why the expects are what they are. I would have expected an assertion of a value being present in the resolved config that is defined in the extended configs?

Why assert with normalize path?

Thank you!

NullVoxPopuli (approved):
thank you!!

PR #842, opened 2025-03-25
Improve multi-byte character handling

NullVoxPopuli:
Summary of the issue: https://github.com/embroider-build/content-tag/issues/45
If folks need to do anything with content-tag in the future, there is this: https://github.com/NullVoxPopuli/content-tag-utils

Hoping to resolve:
- https://github.com/typed-ember/glint/issues/841
- https://github.com/typed-ember/glint/issues/840 (not a multi-byte issue, but "feels" similar)

PR #844, opened 2025-03-25
Use transformed string as source for transformed byte to string index

mogstad:
Previously it was genrating the offset from the original source, but as the transformed code is different, the offseted character wasn't offset at the right spot


mogstad:
Closing for #842

PR #849, opened 2025-03-26
add html attribute validation & completions 

patricklx:
Screenshot 2024-01-03 at 12 45 12
Screenshot 2024-01-03 at 12 48 52


patricklx:
unable to test completions at the moment...

NullVoxPopuli:
Current test failure looks like a path hash mismatch.

@patricklx can you update that snapshot?

I wonder also if you happen to know how to do custom stapshot serializers, and have an idea for these sorts of paths?

Thank you!

packages/template/-private/dsl/lib.dom.augmentation.d.ts
NullVoxPopuli, line 1: Why are these needed on the global type?

On rather, why do we need to add a property to the existing elements?

packages/template/package.json
```diff
@@ -19,7 +19,12 @@
   "devDependencies": {
     "@glimmer/component": "^1.1.2",
     "expect-type": "^0.15.0",
-    "sums-up": "^2.1.0"
+    "sums-up": "^2.1.0",
+    "aria-attributes": "^2.0.1",
+    "html-element-attributes": "^3.3.0",
+    "html-event-attributes": "^2.2.0",
```
NullVoxPopuli, line 25: Is there a package for mathml?

scripts/build-augmentations.mjs
```diff
@@ -0,0 +1,47 @@
...
+});
+
+let augmentations = [];
+for (const element of elements) {
+  augmentations.push(`interface ${element} {
+  [symbol]: '${element}';
+}`);
+}
+
+const prefix = `//generated by scrips/build-augmentations.mjs\n`;
```
NullVoxPopuli, line 41: Can you add to the comment block about why this is needed, how it's used, etc?

Thanks!

scripts/build-elements.mjs
NullVoxPopuli, line 1: Tbh, i'm surprised no existing package has this information for us already.

How strange

NullVoxPopuli:
Very excited for this!

packages/template/package.json
patricklx: yes, https://github.com/wooorm/mathml-tag-names
it only has tag names, no attributes exists for them

packages/template/-private/dsl/lib.dom.augmentation.d.ts
patricklx: That doesn't work...
I cannot make a map of type to another type...

scripts/build-elements.mjs
NullVoxPopuli, line 1: should mathml elements be in here?

packages/template/package.json
NullVoxPopuli: https://github.com/typed-ember/glint/pull/849/files#r2021317938

scripts/build-elements.mjs
patricklx: No, they are already in typescript lib.dom

NullVoxPopuli (approved):

PR #851, opened 2025-03-26
Refrain creating ember-loose virtual code in non ember-loose environment

### LINKED ISSUE #846: V2: JS/TS language service crashes if only `ember-template-imports` is specified in tsconfig glint.environment
1. Clone [this repo](https://github.com/johanrd/gts-2025-03-23) and open it in VS code with the glint extension v1.4.3 and see that the language service works as expected
2. Remove "ember-loose" from tsconfig:

```diff
{
  "extends": "@tsconfig/ember/tsconfig.json",
  "glint": {
-    "environment": ["ember-loose", "ember-template-imports"]
+    "environment": "ember-template-imports"
  },
```

Restart VS code, and see that the language service now crashes:

Image

```
The JS/TS language service immediately crashed 5 times. The service will not be restarted. This may be caused by a plugin contributed by one of these extensions: GitHub.copilot-chat, lifeart.vscode-glimmer-syntax, typed-ember.glint-vscode. Please try disabling these extensions before filing an issue against VS Code.
```

mogstad:
I didn’t find a great place to test this, so ended up adding to the vscode test suite. Any better suggestions are very welcome.
Closes #846

### DIFF
--- a/packages/vscode/__tests__/ts-plugin-tests/smoketest-template-imports-ts-plugin.test.ts
+++ b/packages/vscode/__tests__/ts-plugin-tests/smoketest-template-imports-ts-plugin.test.ts
@@ -55,6 +55,31 @@ describe('Smoke test: ETI Environment (TS Plugin Mode)', () => {
       ]);
     });
 
+    test.only('gives diagnostics for TypeScript file', async () => {
+      let scriptURI = Uri.file(`${rootDir}/src/file.ts`);
+      let scriptEditor = await window.showTextDocument(scriptURI, { viewColumn: ViewColumn.One });
+
+      await hackishlyWaitForTypescriptPluginToActivate(scriptEditor, scriptURI);
+
+      // Comment out a property in the script that's referenced in the template
+      await scriptEditor.edit((edit) => {
+        edit.delete(new Range(0, 0, 0, 7));
+      });
+
+      // Wait for the diagnostic to show up
+      await waitUntil(() => languages.getDiagnostics(scriptURI).length);
+
+      // Verify it's what we expect
+      expect(languages.getDiagnostics(scriptURI)).toMatchObject([
+        {
+          message: "'greeting' is declared but its value is never read.",
+          source: 'ts',
+          code: 6133,
+          range: new Range(new Position(0, 4), new Position(0, 12)),
+        },
+      ]);
+    });
+
     describe('codeactions args', () => {
       test('adds missing args from template into Args type', async () => {
         let scriptURI = Uri.file(`${rootDir}/src/Greeting.gts`);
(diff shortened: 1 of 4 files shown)


NullVoxPopuli (approved):
lgtm

packages/vscode/__tests__/ts-plugin-tests/smoketest-template-imports-ts-plugin.test.ts
```diff
@@ -55,6 +55,31 @@ describe('Smoke test: ETI Environment (TS Plugin Mode)', () => {
       ]);
     });
 
+    test.only('gives diagnostics for TypeScript file', async () => {
```
machty, line 58: This `.only()` needs to be removed so that other tests run. Normally vitest catches this on CI but these VSCode tests I think use mocha or something that doesn't catch cases like these.

NullVoxPopuli: Bah. I missed this :(

mogstad: Ah sorry!

machty:
Nice, I can see some perf gains here; I wanted to guard for this originally but with some of the trickiness in TS Plugin's synchronous API it wasn't clear that we'd be able to synchronously determine the presence of `ember-loose`. Glad you were able to bring it back.

machty:
FWIW I think there might be further potential perf gains here, because with your PR I believe we are still going to be parsing EVERY .ts file into the Loose VirtualCode upon the possibility that it contains a classic loose mode Component definition, i.e. even a file containing an Ember Data / WarpDrive model is going to be parsed into VirtualCode.

Part of the reason I did it this way was 1. just to get something working and 2. based on angular-webstorm's experience. With angular (pardon me bc I could be getting something wrong), a .ts file is considered a component once the `@Component` decorator is introduced, which might seem like an obvious / easy thing to detect to determine whether to parse into VirtualCode, but I think they opted to just parse ALL .ts into VirtualCode because, unless you do, with the way Volar works, Volar wouldn't be able to detect that a .ts file that previously didn't have a `@Component` should suddenly then be parsed into VirtualCode once `@Component` is added. 

That said angular-webstorm might have additional perf checks to do a light-weight parse within the VirtualCode if no `@Component` is detected.

In our case, perhaps it's feasible to check whether a .ts file has `/components/` in its path? Are there any cases of classic loose mode components being defined in, say, a `src/` directory but not specifically in a `components/` directory?

NullVoxPopuli:
In v2 addons they can be anywhere. But default paths are /components/ 

Its worth the perf. Maybe we can add an option to add more component paths in TSconfig if needed

machty:
Summarizing some of our discussions in DM:

The only potential downside/uncovered case by this perf improvement would be that we can't support the use case of v2 addons that still use loose mode components AND their components don't have a `components/` folder in their folder ancestry. In such a case the addon authors wouldn't see type-checking diagnostics in their .hbs files. This is pretty tiny corner case seeing as how most folk on v2 have probably converted to .gts, and if not it's easy to have them move their components into `components/` (or keep them there if they're already there) if they want the Glint 2 experience.

PR #856, opened 2025-03-27
Support (neo)vim filetypes for gts and gjs

NullVoxPopuli:
glimmer-ts and glimmer-js is just what VSCode-flavored editors use.

neovim has a couple conventions for mixed syntax file-types:
- `${base_language}.${variant}`, e.g.:
  - javascript.jsx
  - javascript.glimmer
  - typescript.tsx
  - typescript.glimmer
   
- `${base_language}${variant}`, (though, I don't like this one, and haven't implemented any equiv for glimmer -- I even removed these in favor of the above) e.g.:
  - javascriptreact
  - typescriptreact
  
  These in particular, _may_ be intended for the (incorrect) flavors of `.js` and `.ts` files that have jsx inside them and don't declare so via their extension (this was common when jsx was new)
  

Solves this problem when using ts-plugin mode in neovim:
```
[Info] Using Typescript version (bundled) 5.8.2 from path "🏠/.local/share/nvim/mason/packages/typescript-language-server/node_modules/typescript/lib/tsserver.js"
[Error] Notification handler 'textDocument/didOpen' failed with message: Cannot open document 'file://<repo>/app/templates/view.gts' (languageId: typescript.glimmer).
[Error] Unexpected resource file://<repo>/app/templates/view.gts
```


machty:
@NullVoxPopuli lgtm, but just curious, does Vue's tooling avoid this because VSCode's languageID for .vue files' is the same as Neovim's?

NullVoxPopuli:
yea, vue chose just `vue` for their filetype in every environment

NullVoxPopuli:
but they also have in-file differentiation of js/ts. so it makes sense that their _outer_ file can be "just vue"

our outer syntax is just js/ts, so we don't have as much control in that regard

machty:
LGTM but i think maybe just one of these tests would suffice because they all share the same branch point essentially.

PR #874, opened 2025-04-01
Fix organize imports

### LINKED ISSUE #867: Organize Imports removes all imports
When I save a gts file or run organize imports manually, all of my imports are removed.

@glint/* 1.5.2
glint addon 1.4.5

machty:
The hacks we were using to trick TS into considering `.gts`/`.gjs` to be valid implicit extensions (when omitted from import module specifiers) were causing issues with Organize Imports, due to the fact that the hacky fake imports were detected as "unused" and hence removed by Organize Imports in a manner that caused havok when the changes were reverse-source-mapped by Volar.

After some tinkering I landed on a solution that uses `await import()` which doesn't seem to get touched by Organize Imports.

Closes #867


NullVoxPopuli (approved):
seems ok

PR #877, opened 2025-04-03
Reinstate GJS Support

machty:


machty:
Released in pre-release VSCode extension v1.4.9

PR #878, opened 2025-04-04
Breaking test for SVG

machty:
Issue reported in Discord that SVG is yielding incorrect diagnostics for attributes like `class`.

https://discord.com/channels/480462759797063690/1357364002258424059/1357394721999425748

@patricklx you did work on this stuff recently, do you see an obvious fix?

Undesired diagnostics from this breaking test:

```
[
        {
          "category": "error",
          "code": 2353,
          "end": {
            "line": 7,
            "offset": 14,
          },
          "start": {
            "line": 7,
            "offset": 9,
          },
          "text": "Object literal may only specify known properties, and 'xmlns' does not exist in type 'Partial<WithDataAttributes<SVGSVGElementAttributes>>'.",
        },
        {
          "category": "error",
          "code": 2353,
          "end": {
            "line": 9,
            "offset": 56,
          },
          "start": {
            "line": 9,
            "offset": 51,
          },
          "text": "Object literal may only specify known properties, and 'class' does not exist in type 'Partial<WithDataAttributes<SVGRectElementAttributes>>'.",
        },
      ]
```

PR #880, opened 2025-04-06
fix missing generic attrs for svg

patricklx:
fixes #878

### DIFF
--- a/bin/build-elements.mjs
+++ b/bin/build-elements.mjs
@@ -85,6 +85,7 @@ export declare namespace HtmlElementAttributes {
 }
 
 function createSvgElementAttributesMap() {
+  svgElementAttributes['*'] = htmlElementAttributes['*'];
   let svgElementsContent = `
 export declare namespace SvgElementAttributes {
 `;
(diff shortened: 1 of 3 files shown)


patricklx:
I think this should fix the issue. Just need to run script to update elements.d.ts

NullVoxPopuli:
>  Just need to run script to update

I ran both and added the changes to this branch

johanrd:
@patricklx Thanks! This seems to fix most namespace cases within svg!

The `xlmns` attribute is still yelled at with glint `1.4.1-unstable.6e222ff`:

Screenshot 2025-04-11 at 08 08 58


```xhtml
<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'> 
  <path d='M16.32 14.9l5.39 5.4a1 1 0 0 1-1.42 1.4l-5.38-5.38a8 8 0 1 1 1.41-1.41zM10 16a6 6 0 1 0 0-12 6 6 0 0 0 0 12z' />
</svg>
```

Even though xmlns [is not required](https://github.com/typed-ember/glint/commit/6286d96f7a94cbb97f589bd8103eff8ff57eda7f#diff-fb2c4693b419c3d4c761d3cdc3a381b48f476147de483dc80e454456f42d9171L647), it should be allowed by glint, or? 

Since it is optional/unnecessary in inline svgs, it may be a case for ember-template-lint or similar, rather than glint?

PR #895, opened 2025-04-16
VSCode: Add activationEvents for glimmer-ts/js file types

machty:


machty:
I cannot for the life of me get these tests to be un-flaky.

PR #904, opened 2025-06-14
fix html element type

patricklx:


NullVoxPopuli:
This fix included in https://github.com/typed-ember/glint/pull/903

PR #914, opened 2025-07-08
Re-enable types tests

machty:


packages/template/__tests__/attributes.test.ts
```diff
@@ -213,7 +211,7 @@ class MyComponent extends TestComponent<{ Element: HTMLImageElement }> {
     'aria-atomic': undefined,
     // @ts-expect-error: setting a `void` return as an attr makes no sense
     'aria-autocomplete': undefined as void,
-    // @ts-expect-error: DOM nodes aren't valid values
+    // @disabled-ts-expect-error: DOM nodes aren't valid values
```
machty, line 214: With some of @patricklx 's changes, it seems we've shifted away from using `AttrValue` for typing attributes of an element and instead are using maps of known attrs (and extending this for SVG and other use cases). Perhaps we can reinstate the "strictess" of `AttrValue` but maybe the time spent on this won't be worth the value add?

NullVoxPopuli: I dunno how others feel about correctness, or the DX of being more permissive here, but does HTML allow arbitrary attributes? maybe we could do `KnownKey & string` -- I think the key thing is that we want attribute autocomplete, and errors to show when types of known attributes are incorrect

bin/build-augmentations.mjs
```diff
@@ -51,7 +51,7 @@ const filePath = resolve(
   '../../packages/template/-private/dsl/lib.dom.augmentation.d.ts',
 );
 let content = prefix;
-content += 'export const GlintSymbol: unique symbol = Symbol();\n\n';
+content += 'export const GlintSymbol: unique symbol;\n\n';
```
machty, line 54: This is needed because otherwise I get errors due to runtime code (instantiating a Symbol) trying to operate in ambient .d.ts context.

packages/template/__tests__/attributes.test.ts
gossi: I literally added this to my types yesterday (required to let the build pass):

```ts
import type { AttrValue } from '@glint/template/-private/dsl';

declare module '@glint/template/-private/dsl/elements' {
  export declare namespace HtmlElementAttributes {
    interface GenericAttributes {
      ['part']: AttrValue;
    }

    interface HTMLTextAreaElementAttributes {
      ['type']: AttrValue;
    }
  }
}
```

... and was wondering, why not map over existing interfaces that already include all attributes?

Happy to see, you are already one step ahead 🙌

Ad `KnownKey & string`: such a loose mode might be an escape-hatch for only when your build is failing, but code actually is working.

PR #942, opened 2025-08-20
feature: add support for single file type checking

BojanRatkovic:
Previous idea and discussion: https://github.com/typed-ember/glint/discussions/940

This feature adds support for single file type checking in Glint, allowing developers to type-check individual files instead of the entire project. It enables faster feedback during development by running type checks on specific files rather than performing a full project compilation. The implementation includes CLI support for targeting individual files and updates to the configuration system to handle single-file scenarios.

### DIFF
--- a/docs-v2/getting-started.md
+++ b/docs-v2/getting-started.md
@@ -51,6 +51,33 @@ To this:
 
 You can also use the `glint` command locally with the `--watch` flag to monitor your project as you work!
 
+#### Single File Checking
+
+Glint supports checking individual files or a specific set of files instead of your entire project. This can be useful for faster feedback during development or when working with large codebases.
+
+```bash
+# Check a single file
+npx glint src/components/my-component.gts
+
+# Check multiple files
+npx glint src/components/header.gts src/components/footer.gts
+
+# Check files with different extensions
+npx glint src/helpers/format-date.ts src/components/date-picker.gts
+```
+
+When checking specific files, Glint:
+- Uses your project's `tsconfig.json` configuration
+- Applies the same type checking rules as project-wide checking
+- Only analyzes the specified files for faster performance
+- Maintains all your project's compiler options and path mappings
+
+This is particularly useful for:
+- **IDE integrations** that need to check files on save
+- **Git hooks** that validate only changed files
+- **Development workflows** where you want quick feedback on specific components
+- **CI optimizations** for incremental builds
+
 ### Glint Editor Extensions
 
 You can install an editor extension to display Glint's diagnostics inline in your templates and provide richer editor support&mdash;typechecking, type information on hover, automated refactoring, and more&mdash;powered by `glint-language-server`:
--- a/docs-v2/getting-started.md
+++ b/docs-v2/getting-started.md
@@ -72,12 +72,6 @@ When checking specific files, Glint:
 - Only analyzes the specified files for faster performance
 - Maintains all your project's compiler options and path mappings
 
-This is particularly useful for:
-- **IDE integrations** that need to check files on save
-- **Git hooks** that validate only changed files
-- **Development workflows** where you want quick feedback on specific components
-- **CI optimizations** for incremental builds
-
 ### Glint Editor Extensions
 
 You can install an editor extension to display Glint's diagnostics inline in your templates and provide richer editor support&mdash;typechecking, type information on hover, automated refactoring, and more&mdash;powered by `glint-language-server`:
@@ -92,5 +86,3 @@ To get Ember/Glimmer and TypeScript working together, Glint creates a separate T
 1. Type `@builtin typescript` in the extension search box
 1. Click the little gear icon of "TypeScript and JavaScript Language Features", and select "Disable (Workspace)".
 1. Reload the workspace. Glint will now take over TS language services.
-
-![Disabling built-in TS language service per workspace](https://user-images.githubusercontent.com/108688/111069039-6dc84100-84cb-11eb-8339-18a589be2ac5.png)
--- a/docs-v2/getting-started.md
+++ b/docs-v2/getting-started.md
@@ -69,8 +69,7 @@ npx glint src/helpers/format-date.ts src/components/date-picker.gts
 When checking specific files, Glint:
 - Uses your project's `tsconfig.json` configuration
 - Applies the same type checking rules as project-wide checking
-- Only analyzes the specified files for faster performance
-- Maintains all your project's compiler options and path mappings
+- Analyzes the specified files and all their dependencies for faster performance
 
 ### Glint Editor Extensions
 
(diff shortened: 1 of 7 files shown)


docs-v2/getting-started.md
```diff
@@ -51,6 +51,33 @@ To this:
...
+```
+
+When checking specific files, Glint:
+- Uses your project's `tsconfig.json` configuration
+- Applies the same type checking rules as project-wide checking
+- Only analyzes the specified files for faster performance
+- Maintains all your project's compiler options and path mappings
+
+This is particularly useful for:
+- **IDE integrations** that need to check files on save
```
NullVoxPopuli, line 76: This bullet point makes no sense.

All glint is is typescript integration, and we have this already

docs-v2/getting-started.md
```diff
@@ -51,6 +51,33 @@ To this:
...
+npx glint src/components/header.gts src/components/footer.gts
+
+# Check files with different extensions
+npx glint src/helpers/format-date.ts src/components/date-picker.gts
+```
+
+When checking specific files, Glint:
+- Uses your project's `tsconfig.json` configuration
+- Applies the same type checking rules as project-wide checking
+- Only analyzes the specified files for faster performance
```
NullVoxPopuli, line 72: Neededly, everything those files reference as well.

docs-v2/getting-started.md
```diff
@@ -51,6 +51,33 @@ To this:
...
+- Maintains all your project's compiler options and path mappings
```
NullVoxPopuli, line 73: This point is redundant

docs-v2/getting-started.md
```diff
@@ -66,4 +93,4 @@ To get Ember/Glimmer and TypeScript working together, Glint creates a separate T
 1. Click the little gear icon of "TypeScript and JavaScript Language Features", and select "Disable (Workspace)".
 1. Reload the workspace. Glint will now take over TS language services.
 
-![Disabling built-in TS language service per workspace](https://user-images.githubusercontent.com/108688/111069039-6dc84100-84cb-11eb-8339-18a589be2ac5.png)
\ No newline at end of file
+![Disabling built-in TS language service per workspace](https://user-images.githubusercontent.com/108688/111069039-6dc84100-84cb-11eb-8339-18a589be2ac5.png)
```
NullVoxPopuli, line 96: Why was this added? This is wrong.
We integrate with ts, not disable it

docs/getting-started.md
```diff
@@ -36,6 +36,33 @@ To this:
 
 You can also use the `glint` command locally with the `--watch` flag to monitor your project as you work!
 
+#### Single File Checking
```
NullVoxPopuli, line 39: This whole section could probably be deleted.
These are old docs, afaik

packages/core/src/cli/run-volar-tsc.ts
```diff
@@ -21,16 +44,56 @@ export function run(): void {
...
-      } else {
-        return [];
+  if (hasSpecificFiles) {
+    // For specific files, create temporary tsconfig that inherits from project config
+    const { tempConfigPath, cleanup } = createTempConfigForFiles(cwd, files);
+    const originalArgv = process.argv;
+    
+    try {
+      // Build TypeScript arguments for single file checking
+      const tscArgs = ['node', 'tsc', '--project', tempConfigPath];
```
NullVoxPopuli, line 59: We can't invoke tsc for type cheking.

That evade's all the transforming glint does.

docs-v2/getting-started.md
```diff
@@ -51,6 +51,33 @@ To this:
 
 You can also use the `glint` command locally with the `--watch` flag to monitor your project as you work!
 
+#### Single File Checking
+
+Glint supports checking individual files or a specific set of files instead of your entire project. This can be useful for faster feedback during development or when working with large codebases.
```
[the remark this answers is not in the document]
BojanRatkovic: Yes, > 5000 files. 

Running the `glint` command on all files can take up to 60 seconds (when used with lint-staged). Since we're experiencing issues with the Glint extension frequently crashing, broken code sometimes gets pushed upstream, and TypeScript errors are only caught later when the full command is executed in CI.

packages/core/src/config/index.ts
NullVoxPopuli, line 1: Why import stuff and not use it?

docs-v2/getting-started.md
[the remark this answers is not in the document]
BojanRatkovic: > Lint staged has a glint plugin? Do you have to use that?

No need for specific plugins, you can use it like this. 
```
{
  "lint-staged": {
    "*.{gts,ts}": [
      "glint --noEmit"
    ]
  }
}
```
I wouldn't say we have to use it that way, but it would be super convenient if, along with other linters, it could run glint (lint) on the changed files at commit time together with stylelint, for example, and the others...

> Also, have you tried the v2 alphas and prererease in vscode?

I don't think so...

packages/core/src/config/index.ts
BojanRatkovic: Leftovers during development, I'll fix it.

docs-v2/getting-started.md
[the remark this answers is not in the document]
NullVoxPopuli: > it could run glint (lint) on the changed files at commit time together with stylelint, for example, and the others...

For sure, but typescript doesn't work this way, it has to resolve the types of everything imported.

But! I will concede if with this branch, the outputs are significantly different:
```bash
# for reference 
glint --version
glint -v
# measure
time glint
time glint ./app/templates/application.gts
time glint ./app/components/something.gts
```

If you could get me these numbers, that'd be a huge help

> I don't think so...

You'll want to, as we've frozen glint v1 and are only working on glint v2 right now. It's already way faster.

Caveat tho, we dropped support for hbs.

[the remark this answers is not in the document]
NullVoxPopuli: That's pretty good results!

packages/core/src/cli/run-volar-tsc.ts
```diff
@@ -21,16 +44,56 @@ export function run(): void {
...
+      
+      // Convert compiler options back to command line arguments
+      // Skip conflicting options that we control
+      const filteredOptions = { ...compilerOptions };
+      delete filteredOptions.project;
+      
+      // Add --noEmit as default only if user hasn't specified emit-related flags
+      const hasEmitFlag = Boolean(
+        compilerOptions.noEmit ||
```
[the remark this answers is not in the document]
BojanRatkovic: It makes sense to support emit, as long as the default is `noEmit`, so why not let users decide? Maybe someone just wants to test a single file and see the final JS output for whatever reason. 

But whatever you say, I'm more or less fine with either way.

machty:
@BojanRatkovic thank you for your contribution, and the perf improvements definitely seem nice.

That said I would like to investigate alternatives before bringing in an extension to the `glint` binary, which since v2/volar is basically a decorated clone of the `tsc` binary. Once we introduce these changes then it'll be a little bit harder to follow along with Vue/Volar as enhancements to those projects are made upstream.

It seems that what you're describing would be useful for all `tsc` variants:

- vanilla `tsc`
- `vue-tsc`
- `glint`

Is it possible for you to build a CLI tool that wraps the underyling `tsc` / `vue-tsc` / `glint` binary and does all of the temporary `tsconfig` build-up, invoking the underlying binary, and then teardown? Then you could solve the problem for all variants and we could avoid this custom enhancement specifically to glint.

PR #946, opened 2025-08-29
Support .gjs.d.ts and .d.gjs.ts

wagenet:
Glint V1 will pick up `.d.ts` files adjacent to a `.gjs` file because `.gjs` is transformed to `.js` behind the scenes. However, this does not work in Glint V2 and you must explicitly import with a `.gjs` extension in which case it looks for `.gjs.d.ts` or `.d.gjs.ts`. We should support both of these extensions in Glint v1 to make migration easier.


NullVoxPopuli:
we now defer to the ts7 convention for this --- see: 
- https://github.com/microsoft/TypeScript/issues/64053

and
- https://github.com/microsoft/TypeScript/pull/64120

PR #939, opened 2025-08-19
Absorb `@glint/environment-ember-template-imports` into `@glint/core`

machty:
With this PR there will be no more "environments", only the "ember template import" aka "template tag format" baked into `@glint/core`.

PR #966, opened 2025-09-26
Reinstate additionalSpecialForms

machty:
https://github.com/typed-ember/glint/pull/939#issuecomment-3338330763

cc @bwbuchanan

### DIFF
--- a/packages/core/src/config/environment.ts
+++ b/packages/core/src/config/environment.ts
@@ -38,9 +38,11 @@ export class GlintEnvironment {
   ): GlintEnvironment {
     // Glint V1:
     let additionalGlobals = null;
+    let additionalSpecialForms = {};
     if (topLevelGlintConfigObject) {
       if (topLevelGlintConfigObject.additionalGlobals) {
         additionalGlobals = topLevelGlintConfigObject.additionalGlobals;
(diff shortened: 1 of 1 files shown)


packages/core/src/config/environment.ts
```diff
@@ -38,9 +38,11 @@ export class GlintEnvironment {
   ): GlintEnvironment {
     // Glint V1:
     let additionalGlobals = null;
+    let additionalSpecialForms = {};
     if (topLevelGlintConfigObject) {
       if (topLevelGlintConfigObject.additionalGlobals) {
```
bwbuchanan, line 43: Should be `topLevelGlintConfigObject.additionalGlobals || topLevelGlintConfigObject.additionalSpecialForms`?

machty: Yes it should be, sorry, I'll do a followup commit with a few tests.

machty: Hopefully this should do it. https://github.com/typed-ember/glint/releases/tag/v2.0.0-alpha.10-%40glint%2Fcore

bwbuchanan: Thanks! All looks good here.

PR #1000, opened 2025-10-17
Support HTML properties for HTML elements

mogstad:
Ember supports setting both attributes and properties for HTML elements. As properties typed compared to attributes, I've included support for setting the specific type.

I've not made sure this is a comprehensive list of properties on HTML elements, there are likely more properties.


NullVoxPopuli:
Seems good! tho if anything's goofy, we can keep doing prs.

However! Can we add a test or something (code that would type check) to ensure this doesn't regress? Ty! <3

PR #1052, opened 2026-02-04
Support environments with no tsconfig or jsconfig

NullVoxPopuli:
This tweaks the logic for when glint should activate (which has to happen independent of opening a file (for vscode reasons))

To test it out:

[glint2-vscode-1.0.3-no-config.zip](https://github.com/user-attachments/files/25222917/glint2-vscode-1.0.3-no-config.zip)

> [!IMPORTANT]
> in order for template-aware intellisense, you _MUST_ have `@glint/template` and `@glint/ember-tsc` installed (which means also having typescript) -- this is because the Volar virtual files require `@glint/template` and `@glint/ember-tsc` and import _as if_ your app defined these dependencies.

> [!NOTE]
> in order to get intellisense for ember-source things, you _MUST_ have a jsconfig.json or tsconfig.json

This PR is about making the experience not terrible for folks that haven't done either of the above

### DIFF
--- a/packages/core/src/config/loader.ts
+++ b/packages/core/src/config/loader.ts
@@ -24,6 +24,8 @@ type TypeScript = typeof TS;
 export class ConfigLoader {
   private configs = new Map<string, GlintConfig | null>();
   private logInfo?: (message: string) => void;
+  private loggedNoConfig = new Set<string>();
+  private loggedCachedConfig = new Set<string>();
 
   constructor(logInfo?: (message: string) => void) {
     this.logInfo = logInfo;
@@ -44,14 +46,19 @@ export class ConfigLoader {
     }
 
     let configPath = findNearestConfigFile(ts, directory);
-    if (!configPath) {
+    let cacheKey = configPath ?? directory;
+
+    if (!configPath && !this.loggedNoConfig.has(cacheKey)) {
       this.log(`No tsconfig.json or jsconfig.json found from ${directory}.`);
+      this.loggedNoConfig.add(cacheKey);
     }
 
-    let cacheKey = configPath ?? directory;
     let existing = this.configs.get(cacheKey);
     if (existing !== undefined) {
-      this.log(`Using cached Glint config from ${cacheKey}.`);
+      if (!this.loggedCachedConfig.has(cacheKey)) {
+        this.log(`Using cached Glint config from ${cacheKey}.`);
+        this.loggedCachedConfig.add(cacheKey);
+      }
       return existing;
     }
 
--- a/packages/core/src/config/loader.ts
+++ b/packages/core/src/config/loader.ts
@@ -41,27 +41,26 @@ export class ConfigLoader {
     let ts = findTypeScript(directory);
     if (!ts) {
       this.log(`No TypeScript installation found from ${directory}.`);
-      return null;
     }
 
     let configPath = findNearestConfigFile(ts, directory);
     if (!configPath) {
       this.log(`No tsconfig.json or jsconfig.json found from ${directory}.`);
-      return null;
     }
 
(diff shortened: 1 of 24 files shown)


packages/core/src/volar/ember-language-plugin.ts
```diff
@@ -34,17 +87,21 @@ export function createEmberLanguagePlugin<T extends URI | string>(
       }
```
evoactivity: Should this just return `getLanguageIdForFileName(String(fileNameOrUri))`

evoactivity (changes requested):
Just a couple of things.

evoactivity (approved):

void-mAlex:
@NullVoxPopuli 
this is what the editor looks like after the last steps
image

NullVoxPopuli:
Note for myself -- find a way to make this the default jsconfig for no-config situations:

```json
{
  "compilerOptions": {
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "bundler",
    "experimentalDecorators": true,
    "skipLibCheck": true,
    "strict": false,
    "checkJs": false,
    "noEmit": true,
    "isolatedModules": true,
    "allowImportingTsExtensions": false,
    "resolveJsonModule": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "types": [
      "ember-source/types",
      "@embroider/core/virtual",
      "vite/client",
      "@glint/ember-tsc/types"
    ],
  },
  "exclude": ["dist", "node_modules"]
}
```

from https://discord.com/channels/480462759797063690/1470939378871435557/1473088988561674324

void-mAlex:
> > tried on windows with a brand new ember app based on `pnpm dlx ember-cli@latest new my-app-name -b @ember/app-blueprint --pnpm --skip-npm --no-git --no-welcome`
> 
> This is exactly what I did on windows too, worked first time, no need to add deps manually. Strange 🤔

the extension doesn't activate without the packages which are not in the default bp for non ts projects - it says as much in the extension log
also no ide features were working to suggest glint was doing anything on a gjs file

evoactivity:
image

image

image

🤷 I can use features like rename symbol from the template too without that extra config

NullVoxPopuli:
do you two have different setups somehow?

NullVoxPopuli:
it's unrelated to ts/glint

packages/core/src/config/loader.ts
```diff
@@ -44,9 +44,40 @@ export class ConfigLoader {
       return null;
     }
```
evoactivity: If typescript is not installed in the project none of the following code runs.

If you remove `return null` we end up with breakage,.

Adding typescript then makes it run, and in volar labs I can see the virtual file, but we still end up with `any` types in the template.

A combination of typescript, @glint/ember-tsc (with the glint config setting telling vscode to use the bundled copy) and this jsconfig.js file makes things work 
```json
{
  "compilerOptions": {
    "module": "ESNext",
    "moduleResolution": "bundler",
  }
}
```

I don't have time now to keep digging.

NullVoxPopuli:
[glint2.zip](https://github.com/user-attachments/files/25562890/glint2.zip)

Here is the latest build -- I think I've fixed things, but I'm still checking -- more to do as well

image


with 0 config, 0 typescript, we have this in JS projects now (Volar Labs reports the transform is working)

but a bunch of stuff is `any` typed, and ultimately we want suggestions to import `Component` and `@tracked`

hovering over the HTML parts is useful tho 
image

ef4:
Is this PR superseded? Asking because it's linked from RFC tracking.

NullVoxPopuli:
yup, we landed support for this already and is shipped in the latest release

PR #1102, opened 2026-03-30
Support Glint in JS-only Ember projects without tsconfig/jsconfig

NullVoxPopuli-ai-agent:
## Summary

- Adds `createDefaultConfig()` to `@glint/ember-tsc` that provides `ember-template-imports` environment defaults without requiring a config file
- Removes the tsconfig/jsconfig early bail-out gates from the tsserver plugin and the VSCode extension's `launch()` function
- Falls back to default config in the language server when tsserver reports an inferred project (e.g. `/dev/null/inferredProject1*`)
- The bundled `ember-tsc` is used automatically when `@glint/ember-tsc` is not installed in the workspace

This enables Glint to provide IntelliSense, code actions, import suggestions, and diagnostics in JavaScript-only Ember apps created via `pnpm dlx ember-cli@latest new my-app --pnpm` — with **no tsconfig.json, no jsconfig.json, no TypeScript installed, and no `@glint/*` dependencies** in the consumer project.

## Test plan

- [x] Verified via `pnpm dlx ember-cli@latest new my-app --pnpm` (JS-only project, no tsconfig/jsconfig, no TS, no @glint/* deps)
- [x] Installed the built VSIX extension and opened `app/templates/application.gjs`
- [x] Confirmed tsserver plugin loads via bundled ember-tsc (tsserver log: `Plugin validation succeeded`)
- [x] Confirmed zero syntax and semantic diagnostics on the `.gjs` file
- [x] Confirmed imports (`ember-page-title`, `ember-welcome-page`) resolve correctly
- [x] Confirmed `_glint:projectInfo` and other protocol handlers are active
- [x] Confirmed the Glint language server connects to the correct workspace root
- [ ] Verify existing projects with tsconfig/jsconfig still work as before (regression test)

### DIFF
--- a/packages/vscode/src/extension.ts
+++ b/packages/vscode/src/extension.ts
@@ -505,42 +505,6 @@ function resolveBundledEmberTscServerPath(): string | undefined {
   }
 }
 
-/**
- * Check whether a tsconfig.json or jsconfig.json exists anywhere in the
- * workspace directory tree (searching recursively). Returns the first
- * matching path found, or undefined if none.
- */
-function findConfigFileInWorkspace(workspaceRoot: string): string | undefined {
-  // Quick check: look for tsconfig.json / jsconfig.json in the workspace root first.
-  for (const name of ['tsconfig.json', 'jsconfig.json']) {
-    const candidate = path.join(workspaceRoot, name);
-    if (fs.existsSync(candidate)) {
-      return candidate;
-    }
-  }
-
-  // Walk immediate sub-directories (1 level deep) to cover monorepo layouts
-  // where the config lives in a child package. We intentionally skip
-  // node_modules and hidden directories to keep this check fast.
-  try {
-    const entries = fs.readdirSync(workspaceRoot, { withFileTypes: true });
-    for (const entry of entries) {
-      if (!entry.isDirectory()) continue;
-      if (entry.name === 'node_modules' || entry.name.startsWith('.')) continue;
-      for (const name of ['tsconfig.json', 'jsconfig.json']) {
-        const candidate = path.join(workspaceRoot, entry.name, name);
-        if (fs.existsSync(candidate)) {
-          return candidate;
-        }
-      }
-    }
-  } catch {
-    // If reading the directory fails, fall through to return undefined.
-  }
-
-  return undefined;
-}
-
 // We need to activate the default VSCode TypeScript extension so that our
 // TS Plugin kicks in. We do this because the TS extension is (obviously) not
 // configured to activate for, say, .gts files:
--- a/packages/vscode/src/extension.ts
+++ b/packages/vscode/src/extension.ts
@@ -341,16 +341,6 @@ function launch(
     return undefined;
   }
 
-  // Early bail-out: don't launch the Glint language server if there's no
-  // tsconfig.json or jsconfig.json anywhere in the workspace root tree.
-  // This prevents unnecessary overhead for pure JS projects without config.
-  if (!findConfigFileInWorkspace(workspaceFolder.uri.fsPath)) {
-    outputChannel.appendLine(
-      '[Activation] No tsconfig.json or jsconfig.json found in workspace; not launching Glint.',
-    );
-    return undefined;
-  }
-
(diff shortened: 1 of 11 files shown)


NullVoxPopuli:
omg, the mad lad did it

here is a zip of the vscode extension for folks to try

[glint2-vscode-1.1.0.vsix.zip](https://github.com/user-attachments/files/26356064/glint2-vscode-1.1.0.vsix.zip)

NullVoxPopuli:
I see import suggestions for Component
image


I see jsdoc for existing things:
image


I don't get intellisense for decorators, so this may require a jsconfig in the blueprint 🤔

NullVoxPopuli:
We have element support 
image

evoactivity:
> I don't get intellisense for decorators, so this may require a jsconfig in the blueprint 

I see the same behaviour in a plain .js file too for decorators, so might just be a limitation when using JS?

evoactivity:
@NullVoxPopuli can we have a js-app fixture/test adding so we can ensure we don't introduce regressions to JS support in future.

evoactivity:
@void-mAlex can you have a test of this please.

NullVoxPopuli (approved):

test-packages/js-ember-app/package.json
```diff
@@ -0,0 +1,28 @@
...
+    },
+    "@glint/tsserver-plugin": {
+      "injected": true
+    }
+  },
+  "devDependencies": {
+    "@glimmer/component": "^2.0.0",
+    "@glint/ember-tsc": "workspace:*",
+    "@glint/template": "workspace:*",
+    "@glint/tsserver-plugin": "workspace:*",
```
void-mAlex: why is the tsserver plugin a dependency here?

NullVoxPopuli: Probably just convention from the other projects. This is the first test project that shouldn't have these deps

lifeart (approved):

test-packages/js-ember-app/package.json
NullVoxPopuli: fixed

ef4 (approved):

NullVoxPopuli:
for observers of this PR, we're waiting a day or so for @void-mAlex to test out their setup with this version of glint

packages/vscode/src/extension.ts
```diff
@@ -341,16 +341,6 @@ function launch(
     return undefined;
   }
 
-  // Early bail-out: don't launch the Glint language server if there's no
-  // tsconfig.json or jsconfig.json anywhere in the workspace root tree.
-  // This prevents unnecessary overhead for pure JS projects without config.
-  if (!findConfigFileInWorkspace(workspaceFolder.uri.fsPath)) {
```
void-mAlex, line 347: we should clean up the now unused findConfigFileInWorkspace function

void-mAlex:
generally this is a big improvement

remaining issue that I found was that with a config in place but no typescript installed at the root of the project you get 3 popup errors from the extension as it fails to boot up
functionality still remains in place for the most part being able to go to definition and some intellisense but limited

NullVoxPopuli:
Next PR will attempt to fix that -- thanks for your explorations, @void-mAlex !

evoactivity:
Should be able to solve that with something like this to target the vscode version of tsc

```js
function resolveTsdkPath() {
    const vscodeTsdk = path.join(vscode.env.appRoot, 'extensions', 'node_modules', 'typescript', 'lib');
    if (fs.existsSync(vscodeTsdk)) {
        return vscodeTsdk;
    }

    const tsExt = vscode.extensions.getExtension('vscode.typescript-language-features');
    if (tsExt) {
        // Eclipse Theia
        // see: https://github.com/eclipse-theia/vscode-builtin-extensions/blob/65c70ec636bd879ef9529d0a2da36f4b99139c40/src/package-vsix.js#L71
        const theiaTsdk = path.join(tsExt.extensionPath, 'deps', 'typescript', 'lib');
        if (fs.existsSync(theiaTsdk)) {
            return theiaTsdk;
        }
    }
}
```

PR #1109, opened 2026-05-13
feat(ember-tsc): support Ember 7.1 built-in keywords

aklkv:
Adds support for the new built-in template keywords introduced in ember-source 7.1.0 (RFCs 562, 470, 997, 998, 999, 1000, 560, 561, 389): on, fn, array, hash, and, or, not, eq, neq, lt, lte, gt, gte, element.

These can be referenced in .gjs/.gts templates without an explicit import from @ember/helper or @ember/modifier, matching ember-source 7.1+ runtime behavior.

Detection is automatic — no consumer opt-in required:

- The Keywords interface in types/-private/dsl/globals.d.ts probes '@ember/helper' for the EqHelper symbol (introduced in RFC 561, shipped in ember-source 7.1.0). When present, each new keyword resolves to a real HelperLike / ModifierLike / ComponentLike type. When absent, each resolves to 'never', so un-imported usage in a template fails to type-check on older Ember versions.

- src/environment-ember-template-imports/-private/environment/index.ts: add the 14 keywords to the environment's globals list so the parser recognizes them.

Behavior matrix:

  ember-source >= 7.1.0 -> full types; un-imported keywords type-check ember-source <  7.1.0 -> un-imported keywords resolve to 'never'; Glint reports a clear type error.
  no @ember/helper      -> keywords are enabled (fallback for non-Ember
                           Glimmer projects, gated by a @ts-ignore on
                           the probe import).


packages/ember-tsc/src/environment-ember-template-imports/-private/environment/index.ts
```diff
@@ -61,6 +61,24 @@ export default function emberTemplateImportsEnvironment(
...
+            'and',
+            'or',
+            'not',
+            'eq',
+            'neq',
+            'lt',
+            'lte',
+            'gt',
+            'gte',
+            'element',
```
ijlee2: I noticed that the strings above (`action` to `yield`) are alphabetized. Can we alphabetize the list for Ember 7.1 similarly, and create variables so that it's easier to document and maintain the lists?

The code below is what I mean. (I sorted the RFC IDs, too. I wrote `Ember7_1` but see that the name `Ember71` is used in `types/-private/dsl/globals.d.ts`. Feel free to rename `Ember7_1` to achieve consistency.)

```ts
export default function emberTemplateImportsEnvironment(
  options: Record<string, unknown>,
): GlintEnvironmentConfig {
  let additionalSpecialForms =
    typeof options['additionalSpecialForms'] === 'object'
      ? (options['additionalSpecialForms'] as GlintSpecialFormConfig)
      : {};

  const additionalGlobalSpecialForms = additionalSpecialForms.globals ?? {};

  const additionalGlobals = Array.isArray(options['additionalGlobals'])
    ? options['additionalGlobals']
    : [];

  const globalsForEmber = [
    'action',
    'component',
    'debugger',
    'each',
    'each-in',
    'has-block',
    'has-block-params',
    'helper',
    'if',
    'in-element',
    'let',
    'log',
    'modifier',
    'mount',
    'mut',
    'outlet',
    'unbound',
    'unless',
    'with',
    'yield',
  ];

  /*
    Ember 7.1+ built-in keywords (RFCs 389, 470, 560, 561, 562, 997, 998, 999, 1000).
    Their template-side signatures live in `@glint/ember-tsc/globals/ember-7.1` (opt-in),
    and they only resolve at runtime on `ember-source >= 7.1.0`.
  */
  const globalsForEmber7_1 = [
    'and',
    'array',
    'element',
    'eq',
    'fn',
    'gt',
    'gte',
    'hash',
    'lt',
    'lte',
    'neq',
    'not',
    'on',
    'or',
  ];

  return {
    tags: {
      '@glint/ember-tsc/environment-ember-template-imports/-private/tag': {
        hbs: {
          typesModule: '@glint/ember-tsc/-private/dsl',
          specialForms: {
            globals: {
              if: 'if',
              unless: 'if-not',
              yield: 'yield',
              component: 'bind-invokable',
              modifier: 'bind-invokable',
              helper: 'bind-invokable',
              ...additionalGlobalSpecialForms,
            },
            imports: {
              '@ember/helper': {
                array: 'array-literal',
                hash: 'object-literal',
                ...additionalSpecialForms.imports?.['@ember/helper'],
              },
              ...additionalSpecialForms.imports,
            },
          },
          globals: [
            ...globalsForEmber,
            ...globalsForEmber7_1,
            ...Object.keys(additionalGlobalSpecialForms),
            ...additionalGlobals,
          ],
        },
      },
    },
    extensions: {
      '.gts': {
        kind: 'typed-script',
        preprocess,
        transform,
      },
      '.gjs': {
        kind: 'untyped-script',
        preprocess,
        transform,
      },
    },
  };
}
```

packages/ember-tsc/types/-private/dsl/globals.d.ts
```diff
@@ -199,6 +224,117 @@ interface Keywords {
 
   // `{{yield}}` is implemented directly in `ember-tsc`
   yield: void;
+
+  // ---- Ember 7.1+ built-in keywords ----
+  // These are gated on `HasEmber71BuiltIns` above. On `ember-source < 7.1.0`
+  // each one resolves to `never`, so un-imported usage in a template fails
+  // to type-check.
```
ijlee2: **suggestion**

Similarly to https://github.com/typed-ember/glint/pull/1109/changes#r3231711703, maybe we can create temporary variables and sort keys to ease maintenance?

```ts
interface KeywordsForEmber {
  /* ... */
}

interface KeywordsForEmber7_1 {
  /* ... */
}

interface Keywords extends KeywordsForEmber, KeywordsForEmber7_1 {}
```

ijlee2: Thank you!

NullVoxPopuli (approved):

PR #1114, opened 2026-05-15
fix(ember-tsc): respect local imports over template globals

### LINKED ISSUE #1113: Regression in @glint/ember-tsc@1.6.1
## Background

`@embroider/try` scenarios for `ember-intl` passed yesterday, but failed today. I suspect that, when installing the changed dependencies, `@glint/ember-tsc` got updated to the latest.

- https://github.com/ember-intl/ember-intl/issues/2147#issuecomment-4457897474
- https://github.com/ember-intl/ember-intl/blob/8.2.6/docs/my-v2-addon/src/components/select-locale.gts#L81

```sh
src/components/select-locale.gts:81:7 - error TS2349: This expression is not callable.
  Type 'never' has no call signatures.

81       {{on "change" this.updateValue}}
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Found 1 error in src/components/select-locale.gts:81
```

I reproduced the issue locally by updating `@glint/ember-tsc` to `1.6.1`, then running `pnpm build` in `docs/my-v2-addon`.

aklkv:
When the user has imported a binding with the same name as one of the template's configured globals (e.g. `import { on } from '@ember/modifier'` where `on` is also listed in the Ember 7.1+ built-in keywords introduced in 1.6.1), prefer the local import.

Without this filter the template transform routes the path through `__glintDSL__.Globals[<name>]` and silently shadows the user's import. That is a regression for projects on `ember-source < 7.1` where the corresponding `Globals` entry is gated by `Ember71Only<T>` and resolves to `never` -- producing TS2349 'This expression is not callable. Type "never" has no call signatures.' on otherwise valid code such as

    import { on } from '@ember/modifier';
    <template>
      <select {{on "change" this.update}} />
    </template>

Fixes #1113.

### DIFF
--- a/packages/ember-tsc/src/transform/template/inlining/tagged-strings.ts
+++ b/packages/ember-tsc/src/transform/template/inlining/tagged-strings.ts
@@ -36,6 +36,19 @@ export function calculateTaggedTemplateSpans(
     );
 
     let { typesModule, globals } = info.tagConfig;
+
+    // If the user has imported a binding with the same name as one of the
+    // template's configured globals (e.g. `import { on } from '@ember/modifier'`
+    // when `on` is also listed in the Ember 7.1+ built-in keywords), prefer
+    // the local import. Without this filter the template transform would
+    // silently shadow the user's import by routing the path through
+    // `__glintDSL__.Globals[<name>]`, which is a regression for projects on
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
thank you! I've pulled this in to here: https://github.com/typed-ember/glint/pull/1121

PR #1123, opened 2026-05-16
Give better error when gjs/gts content-tag mismatch/parse-error

NullVoxPopuli:


test-packages/package-test-core/__tests__/language-server/diagnostics.test.ts
```diff
@@ -837,6 +837,32 @@ describe('Language Server: Diagnostics (ts plugin)', () => {
...
+      }
+    `;
+
+    const diagnostics = await requestTsserverDiagnostics(
+      'ts-template-imports-app/src/empty-fixture.gts',
+      'glimmer-ts',
+      code,
+    );
+
+    expect(diagnostics).toMatchInlineSnapshot(`[]`);
```
NullVoxPopuli: we should report the content-tag error here

PR #1139, opened 2026-05-27
fix(ember-tsc): omit 7.1 keywords from Globals on ember-source < 7.1

### LINKED ISSUE #1138: Auto suggestion for (and, eq...) doesn't work anymore for ember < 7.1
Starting with glint v1.6.0+ the auto suggestion was broken, when you want to use (and, or, eq.... from ember-truth-helpers)

```ts
import type { TemplateOnlyComponent } from '@ember/component/template-only';

const test = 'test';
const test2 = 'test2';

export default <template>
  {{#if (and test test2)}}
    
  {{/if}}
</template> satisfies TemplateOnlyComponent<{ Args: unknown; }>;
```

Image

Also the tooptip on mouseover is wrong see:

Image

aklkv:
Fixes #1138.

## Background

PR #1125 introduced `KeywordsForEmber71` to expose the ember-source 7.1 built-in
keyword set (`and`, `or`, `eq`, `fn`, `hash`, `not`, `gt`/`gte`/`lt`/`lte`/`neq`,
`on`, `element`, `array` — RFCs 389, 470, 560, 561, 562, 997, 998, 999, 1000) on
the template `Globals` interface, gated by a `HasEmber71BuiltIns` probe against
`@ember/helper`'s value exports. To keep behaviour identical on ember-source
< 7.1, each keyword type was wrapped in `Ember71Only<T>` so it resolved to
`never`. The keyword names were *also* added unconditionally to the runtime
`globals` array in the `ember-template-imports` environment config (so the
transform would emit `__glintDSL__.Globals.and` rather than a bare `and`).

That combination shadows user-imported helpers on ember-source < 7.1:

- The property keys remain present on `Globals`, so TypeScript reports no
  missing-property diagnostic for `{{and}}` / `{{eq}}` / etc.
- Because the diagnostic is absent and the names are in the strict-mode
  `globals` set, the transform emits `__glintDSL__.Globals.and` instead of a
  bare `and`. Even if a diagnostic *did* fire, it would be a property-access
  error on `Globals`, which TypeScript does not offer auto-import quick-fixes
  for.
- Completion in templates suggests the `never`-typed builtin instead of the
  user's scope-imported `and` from `ember-truth-helpers`, and hover shows
  `never`.

## Fix

Two-layer fix so consumers on ember-source < 7.1 get both a "Cannot find name"
diagnostic *and* the auto-import quick-fix that points at `ember-truth-helpers`
/ `@ember/helper`:

1. **Type layer (`packages/ember-tsc/types/-private/dsl/globals.d.ts`)** —
   rename `KeywordsForEmber71` to `KeywordsForEmber71Members`, unwrap each
   field (no more `Ember71Only<T>`), and gate the whole member set at the
   `Globals` intersection:

   ```ts
   type KeywordsForEmber71 = HasEmber71BuiltIns extends true
     ? KeywordsForEmber71Members
     : {};

   export const Globals: Keywords & KeywordsForEmber71 & Globals;
   ```

   On ember-source >= 7.1 behaviour is unchanged (the JSDoc and the dotted
   emit from #1125 keep working). On < 7.1 the keys are absent from `Globals`
   entirely, so TypeScript surfaces a property-access diagnostic.

2. **Runtime layer
   (`packages/ember-tsc/src/environment-ember-template-imports/-private/environment/index.ts`)**
   — add a `hasEmber71BuiltIns()` probe that does
   `createRequire(import.meta.url)('ember-source/package.json')` and parses
   `major.minor`. On < 7.1 the 7.1 names are omitted from the strict-mode
   `globals` set, so the transform emits a bare `and` instead of
   `__glintDSL__.Globals.and`. TypeScript then reports the standard
   `Cannot find name 'and'` diagnostic, which carries an auto-import
   quick-fix.

   Any probe failure (package missing, malformed JSON, unparseable version)
   collapses to `false` so the safe default is to assume pre-7.1 and let
   users import explicitly.

## Behaviour matrix

| ember-source       | `{{eq}}` with no import | Auto-import quick-fix |
| ------------------ | ----------------------- | --------------------- |
| < 7.1 (before fix) | no diagnostic, shadows user import | not offered           |
| < 7.1 (after fix)  | `Cannot find name 'eq'` | `ember-truth-helpers` / `@ember/helper` |
| >= 7.1 (before/after) | resolved as builtin     | n/a                   |

## Verification

- Built `@glint/ember-tsc` cleanly.
- `pnpm --filter @glint/type-test run test:typecheck` passes.
- Patched the build into a downstream consumer project on ember-source 6.x via
  pnpm `patchedDependencies` and confirmed:
  - `{{and …}}` / `{{or …}}` / `{{eq …}}` in `.gts` templates surface
    `Cannot find name 'and'` etc.
  - VS Code "Quick Fix" lightbulb offers `Add import { and } from
    'ember-truth-helpers'`.
  - User-scoped imports flow through completion and hover correctly.
- On ember-source >= 7.1 the original `#1125` behaviour (dotted emit and full
  keyword set on `Globals`) is preserved by the conditional-type intersection
  and the runtime probe returning `true`.

### DIFF
--- a/test-packages/ts-template-imports-app/src/globals/the-7-1-globals-are-not-present.gts
+++ b/test-packages/ts-template-imports-app/src/globals/the-7-1-globals-are-not-present.gts
@@ -1,22 +1,39 @@
-import { expectTypeOf, to } from '@glint/type-test';
-
 <template>
-  {{! these must not exist prior to 7.1 }}
-  {{expectTypeOf on to.beNever}}
-  {{expectTypeOf fn to.beNever}}
-  {{expectTypeOf array to.beNever}}
-  {{expectTypeOf hash to.beNever}}
+  {{! These names must not be present on `Globals` prior to ember-source 7.1.
+      Each `@glint-expect-error` below asserts TypeScript reports
+      "Cannot find name 'X'", which is the diagnostic that carries the
+      auto-import quick-fix (`ember-truth-helpers` / `@ember/helper`) in
+      consumer projects on < 7.1. See packages/ember-tsc/types/-private/dsl/globals.d.ts. }}
 
-  {{expectTypeOf and to.beNever}}
-  {{expectTypeOf or to.beNever}}
-  {{expectTypeOf not to.beNever}}
+  {{! @glint-expect-error -- `on` is not a global on ember-source < 7.1 }}
+  {{on}}
+  {{! @glint-expect-error -- `fn` is not a global on ember-source < 7.1 }}
+  {{fn}}
+  {{! @glint-expect-error -- `array` is not a global on ember-source < 7.1 }}
+  {{array}}
+  {{! @glint-expect-error -- `hash` is not a global on ember-source < 7.1 }}
+  {{hash}}
 
-  {{expectTypeOf eq to.beNever}}
-  {{expectTypeOf neq to.beNever}}
-  {{expectTypeOf gt to.beNever}}
-  {{expectTypeOf lt to.beNever}}
-  {{expectTypeOf gte to.beNever}}
-  {{expectTypeOf lte to.beNever}}
+  {{! @glint-expect-error -- `and` is not a global on ember-source < 7.1 }}
+  {{and}}
+  {{! @glint-expect-error -- `or` is not a global on ember-source < 7.1 }}
+  {{or}}
+  {{! @glint-expect-error -- `not` is not a global on ember-source < 7.1 }}
+  {{not}}
 
-  {{expectTypeOf element to.beNever}}
+  {{! @glint-expect-error -- `eq` is not a global on ember-source < 7.1 }}
+  {{eq}}
+  {{! @glint-expect-error -- `neq` is not a global on ember-source < 7.1 }}
+  {{neq}}
+  {{! @glint-expect-error -- `gt` is not a global on ember-source < 7.1 }}
+  {{gt}}
+  {{! @glint-expect-error -- `lt` is not a global on ember-source < 7.1 }}
+  {{lt}}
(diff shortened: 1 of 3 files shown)


aklkv:
Removed `test-packages/ts-template-imports-app/src/globals/the-7-1-globals-are-not-present.gts` — it asserted each 7.1 keyword was *present-but-typed-`never`* on `Globals`, which is the exact contract this PR intentionally changes. The new contract is that the keys are **absent** from `Globals` on ember-source < 7.1, so referencing `{{on}}` / `{{and}}` / etc. surfaces TypeScript's standard `Cannot find name 'X'` diagnostic (which carries the auto-import quick-fix). That contract is self-enforced by TypeScript — any regression would immediately break compilation in this app (which is on ember-source ^6.2.0). The positive 7.1 path remains covered by `test-packages/ts-gts-7-1-app/src/globals/*.gts`.

NullVoxPopuli (approved):

test-packages/ts-template-imports-app/src/globals/the-7-1-globals-are-not-present.gts
NullVoxPopuli, line 1: We can test this a different way with glint expect error

Let's not delete this file

aklkv: Good call — restored the file and rewrote each assertion as a `{{! @glint-expect-error -- ... }}` over the bare reference. Each comment now positively asserts that TypeScript reports `Cannot find name 'X'` (the diagnostic that surfaces the auto-import quick-fix). Pushed as f11b9cc5.

PR #1156, opened 2026-06-20
fix: used claude to address crash loop in a monorepo, this did fix it… had it add a test

runspired:


packages/ember-tsc/src/environment-ember-template-imports/-private/environment/transform.ts
```diff
@@ -182,14 +196,9 @@ function isETITemplateLiteral(ts: TSLib, node: ts.Node): node is ETITemplateLite
 function findTemplateLocation(
   locations: Array<TemplateLocation>,
   node: ETITemplateExpression | ETITemplateProperty,
-): TemplateLocation {
-  let location = locations.find((loc) => loc.transformedStart === node.getStart());
-
-  if (!location) {
-    throw new Error('Internal error: missing location info for template');
```
NullVoxPopuli, line 189: this error/guard should come back and we probably don't want to change the return type.


as is, this hides errors for a class of things we have yet to discover

NullVoxPopuli (changes requested):
.

NullVoxPopuli:
@runspired in the future, it's probably best to not use the main branch of your fork -- I have no idea why me pushing to it closed this PR -- in any case, I've continued the work over here: https://github.com/typed-ember/glint/pull/1162

PR #1157, opened 2026-06-23
 Fix `tsc -b` / `ember-tsc --build` crash on `.gts` declaration emit ("Extension .gts is unsupported")

NullVoxPopuli:
`ember-tsc --build` errors during declaration emit with:

> Error: Debug Failure. Extension .gts is unsupported:: FileName:: app/components/usage.gts

Turns out `.gts`/`.gjs` are registered as supported TS extensions (so `hasTSFileExtension` is happy), but TS's internal
`tryGetJSExtensionForFile` never learned about them. When declaration emit synthesizes a module specifier with a `"js"`
ending preference pointing at a `.gts` file, it falls through to `getJSExtensionForFile` and `Debug.fail`s, killing the
whole build.

`--noEmit` checking doesn't compute those specifiers, so this only shows up in `--build`.

Fix is to teach `tryGetJSExtensionForFile` that `.gts`/`.gjs` emit `.js` (same as `.ts` -> `.js`), patched in through the
same compiled-tsc-source interception we already use in `run-volar-tsc.ts` for the `proxyCreateProgram` stuff.

> [!NOTE]
> This really belongs in volar's `transformTscContent` (it already patches `changeExtension` + the supported-extension
lists). Once it lands there this monkey-patch can go. The patcher throws a loud error if the tsc source shape changes so a
TS bump won't silently break it.

Test lives in `test-packages/ts-extensionless-app` — a composite `emitDeclarationOnly` config + a fixture shaped to actually
hit the path (one `./util.js` import kept in the `.d.ts` to force the `.js` ending, plus a return type that comes
transitively from another `.gts` so a specifier has to be synthesized). Fails without the fix, green with it.

### DIFF
--- a/packages/ember-tsc/src/cli/run-volar-tsc.ts
+++ b/packages/ember-tsc/src/cli/run-volar-tsc.ts
@@ -13,7 +13,7 @@ const require = createRequire(import.meta.url);
 const fs = require('node:fs') as typeof import('node:fs');
 
 export function run(): void {
-  patchVolarProxyForExtensionlessImports();
+  patchCompilerSourcesForGts();
   patchVolarDecorateProgramForContentTagErrors();
 
   let cwd = process.cwd();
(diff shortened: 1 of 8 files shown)


NullVoxPopuli:
The `tryGetJSExtensionForFile` patch added here is a temporary monkey-patch — the proper home is volar's `transformTscContent`, which already patches tsc's extension awareness in every other spot (`supportedTSExtensions`, `changeExtension`, `extensionsToRemove`).

Filed upstream so this can eventually be removed:
- Issue: volarjs/volar.js#312
- PR: volarjs/volar.js#313

Once a `@volar/typescript` release including that ships, the `patchTscJsExtensionForGts` half of this change can be dropped (same as the plan for the `proxyCreateProgram` patch once volarjs/volar.js#309 lands).

gbudjeakp (approved):

PR #1162, opened 2026-06-24
fix array parsing `[<template>...</template>]`

NullVoxPopuli:
Continuation of https://github.com/typed-ember/glint/pull/1156


NullVoxPopuli:
@runspired, it looks like this didn't fix anything, but just hid the error. I removed your return of missing locations -- but locations should never be missing

PR #1167, opened 2026-06-26
Improve the types for the `element` helper

boris-petrov:
This is what I've been using so far, works pretty nicely.

### DIFF
--- a/packages/ember-tsc/types/-private/intrinsics/element.d.ts
+++ b/packages/ember-tsc/types/-private/intrinsics/element.d.ts
@@ -6,8 +6,28 @@ import { DirectInvokable } from '@glint/template/-private/integration';
  * name. Built-in keyword in ember-source >= 7.1 (RFC 389).
  */
 export type ElementHelper = DirectInvokable<{
-  (tagName: string): ComponentLike<{
-    Element: Element;
+  // `const K` preserves the literal tag name (e.g. `"video"`) through Glint's
+  // `resolve()` inference; without it TypeScript widens the argument to `string`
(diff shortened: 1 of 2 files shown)


boris-petrov:
I'm not sure why this check fails. It worked fine when testing this locally in my project (albeit using Glint 1). `Tag` should be typed as `never` in this case.

NullVoxPopuli:
I'll take a look ! thanks for getting some tests going for me! <3

NullVoxPopuli (approved):

boris-petrov:
Thanks for merging this! Now that I think about it, however, I don't think the case for `''` is very useful/convenient. Check the following example:

```ts
readonly tagName: keyof HTMLElementTagNameMap | '';

{{#let (element @tagName) as |Tag|}}
  <Tag class="foo">Hello</Tag>
{{/let}}
```

This gives an error right now because `Tag` is `null`. But that's kind of inconvenient. I want my code to compile in such a case - I've obviously thought about the cases - if a tag name is passed, an element will be there. If `''` was passed, yes, the `class` that I pass is unused but that's not "wrong" per se. So the types here actually are in my way. Not sure if there is a nice way of solving this with the current types or a new PR would have to be done and make the `''` case narrow down to an `HTMLElement` so that such code compiles.

NullVoxPopuli:
Null would get you a runtime error, too, due to:

https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/-internals/glimmer/lib/helpers/element.ts#L122

You may want an if block

boris-petrov:
Yes, I specifically talk only about an empty string, not about null.

PR #1171, opened 2026-06-27
Report missing-argument errors on helper/component invocations (#1168)

### LINKED ISSUE #1168: [Glint 2] Helpers do not type-check correctly
```gts
import Helper, { helper } from '@ember/component/helper';

class Foo extends Helper<{
  Args: {
    Positional: [number];
    Named: {
      readonly str: string;
    };
  };
  Return: number;
}> {}

const bar = helper((_positional: [number], _named: { readonly str: string }): number => 1);

const baz = (_positional: number, _named: { readonly str: string }): number => 1;

<template>
  {{Foo str=123}}
  {{bar str=123}}
  {{baz str=123}}
</template>
```

Neither of those gives an error but they should - a number of errors. 1) the positional argument is mandatory but it's missing; 2) `str` should be a string, not a number.

NullVoxPopuli-ai-agent:
Fixes #1168.

## Problem

Glint 2 fails to report a missing required positional argument on a template invocation. Given a helper that takes one positional `number` and a named `str: string`, all three of these type-check clean when they should error:

```gts
class Foo extends Helper<{ Args: { Positional: [number]; Named: { readonly str: string } }; Return: number }> {}
const bar = helper((_positional: [number], _named: { readonly str: string }): number => 1);
const baz = (_positional: number, _named: { readonly str: string }): number => 1;

<template>
  {{Foo str=123}}
  {{bar str=123}}
  {{baz str=123}}
</template>
```

## Root cause

The error *is* present in the generated TypeScript — `resolve(Foo)({ str: 123, ...marker })` passes one argument where two are required. But TypeScript anchors the `TS2554 "Expected 2 arguments, but got 1"` diagnostic on the **callee** (`__glintDSL__.resolve(Foo)`), which is synthetic boilerplate, rather than on the args.

`toVolarMappings` only emitted **zero-length boundary** mappings around a node's synthetic output, plus full-length mappings for leaf identifiers. The arity diagnostic's span (`__glintDSL__.resolve(Foo)`) starts and ends in synthetic text that no verification mapping covers, so Volar can't translate it back to the template and silently drops it. (Type errors on *named* args still surfaced, because those map to a real `str` leaf — which is why the bug looked like "helpers don't type-check at all".)

## Fix

Introduce an opt-in `wideVerification` mapping. `forNode` can request an extra **verification-only** Volar mapping (via `generatedLengths`) that covers a node's entire generated span, so a diagnostic anchored anywhere within it — including the synthetic callee — maps back to source.

`emitResolve` flags the `resolve(<path>)` callee and maps it to the tight `path` node. Keeping it narrow (the path, not the whole invocation, which for a block could span unrelated statements) means:

- `@glint-expect-error` suppression stays correctly scoped — a blanket wide mapping on every parent would re-expose diagnostics that a narrower directive suppressed, because an ancestor spanning multiple directive regions carries its own unsuppressed verification.
- Narrower leaf mappings still win for diagnostics that fall squarely within them (the wide mapping is pushed *after* children), so precise squiggles (e.g. a wrong-typed named arg) are unchanged.

The error now reports on the helper name:

```
repro.gts(6,5): error TS2554: Expected 2 arguments, but got 1.
```

## Fixture fallout

Surfacing these previously-dropped diagnostics exposed latent issues in existing fixtures (the emitted TypeScript is byte-for-byte unchanged — only whether the error maps back changed):

- `ts-gts-7-1-app/src/globals/fn.gts` — `(fn handler greet)` bound `greet` (a `(string, string) => string`) to `handler`'s `MouseEvent` parameter; no `fn` overload matches. Reworked into a valid binding.
- `ts-gts-7-1-app/src/globals/array.gts` — `(hash)` with zero args violates `HashHelper`'s required args parameter. Changed to `(hash a=1)`.
- `package-test-core/.../helper.test.gts` — revealed that currying a **generic** positional arg via `{{helper}}` doesn't produce a correctly-typed bound helper (the bound helper is still invalid to call). Documented as a known limitation with `@glint-expect-error`; the underlying generic-currying gap is orthogonal to this change.

A regression test for plain (non-curried) helper-invocation arg checking is added to `helper.test.gts`.

## Testing

- `pnpm --filter '*' run test:typecheck` — pass
- `pnpm --filter '*' run test:tsc` — pass
- `pnpm --filter glint2-vscode test` — pass (4/4)
- `pnpm lint` — pass

### DIFF
--- a/test-packages/package-test-core/__tests__/type-tests/intrinsics/helper-invocation-args.test.gts
+++ b/test-packages/package-test-core/__tests__/type-tests/intrinsics/helper-invocation-args.test.gts
@@ -1,0 +1,56 @@
+import Helper, { helper } from '@ember/component/helper';
+
+// Regression test for https://github.com/typed-ember/glint/issues/1168.
+//
+// A plain (non-curried) helper invocation must type-check its arguments. In
+// particular, a missing required positional argument surfaces TypeScript's
+// "Expected N arguments, but got M" diagnostic. That diagnostic anchors on the
+// synthetic `resolve(...)` callee rather than on the args, and its mapping was
+// previously dropped — so these invocations type-checked clean when they should
+// have errored. The issue reproduced across all three ways of authoring a
+// helper, so each is covered below.
+
+// 1. Class-based helper
+class ClassHelper extends Helper<{
+  Args: {
+    Positional: [count: number];
+    Named: { str: string };
+  };
+  Return: number;
+}> {}
+
+// 2. Functional helper
+const functionalHelper = helper(
+  (_positional: [count: number], _named: { str: string }): number => 1,
+);
+
+// 3. Plain function helper
+const plainFunctionHelper = (_count: number, _named: { str: string }): number => 1;
+
+<template>
+  {{! Valid invocations type-check cleanly. }}
+  {{ClassHelper 1 str="ok"}}
+  {{functionalHelper 1 str="ok"}}
+  {{plainFunctionHelper 1 str="ok"}}
+
+  {{! @glint-expect-error: missing required positional arg }}
+  {{ClassHelper str="ok"}}
+  {{! @glint-expect-error: missing required positional arg }}
+  {{functionalHelper str="ok"}}
+  {{! @glint-expect-error: missing required positional arg }}
+  {{plainFunctionHelper str="ok"}}
+
+  {{! @glint-expect-error: positional arg is the wrong type }}
+  {{ClassHelper "nope" str="ok"}}
+  {{! @glint-expect-error: positional arg is the wrong type }}
+  {{functionalHelper "nope" str="ok"}}
+  {{! @glint-expect-error: positional arg is the wrong type }}
+  {{plainFunctionHelper "nope" str="ok"}}
+
+  {{! @glint-expect-error: named arg is the wrong type }}
+  {{ClassHelper 1 str=123}}
+  {{! @glint-expect-error: named arg is the wrong type }}
+  {{functionalHelper 1 str=123}}
+  {{! @glint-expect-error: named arg is the wrong type }}
+  {{plainFunctionHelper 1 str=123}}
+</template>;
--- a/packages/ember-tsc/src/transform/template/glimmer-ast-mapping-tree.ts
+++ b/packages/ember-tsc/src/transform/template/glimmer-ast-mapping-tree.ts
@@ -58,6 +58,14 @@ export default class GlimmerASTMappingTree {
     public children: Array<GlimmerASTMappingTree> = [],
     public sourceNode: MappingSource,
     public codeInformation?: CodeInformation,
+    /**
+     * When set, `toVolarMappings` emits an additional verification-only mapping
+     * covering this node's entire generated span (even when it differs in length
+     * from the source span). This lets diagnostics that TS anchors on synthetic
+     * generated boilerplate — e.g. the `__glintDSL__.resolve(foo)` callee, where
+     * a missing-positional-arg error lands — map back to the template (#1168).
+     */
+    public wideVerification: boolean = false,
   ) {
     children.forEach((child) => (child.parent = this));
   }
--- a/packages/ember-tsc/src/transform/template/map-template-contents.ts
+++ b/packages/ember-tsc/src/transform/template/map-template-contents.ts
@@ -93,8 +93,18 @@ export type Mapper = {
   /**
    * Map all content emitted in the given callback to the span
    * corresponding to the given AST node in the original source.
+   *
+   * When `wideVerification` is set, an extra verification-only mapping covering
+   * the callback's entire generated span is emitted so that diagnostics anchored
+   * on synthetic generated boilerplate within it still map back to the node
+   * (see `GlimmerASTMappingTree#wideVerification`).
    */
-  forNode(node: AST.Node, callback: () => void, codeFeaturesForNode?: CodeInformation): void;
+  forNode(
+    node: AST.Node,
+    callback: () => void,
+    codeFeaturesForNode?: CodeInformation,
+    wideVerification?: boolean,
+  ): void;
   forNodeWithSpan(
     node: AST.Node,
     span: AST.Node['loc'],
@@ -193,6 +203,7 @@ export function mapTemplateContents(
     allowEmpty: boolean,
     callback: () => void,
     codeFeaturesForNode?: CodeInformation,
+    wideVerification = false,
   ): void => {
     let start = offset;
     let mappings: GlimmerASTMappingTree[] = [];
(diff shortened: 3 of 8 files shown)


test-packages/package-test-core/__tests__/type-tests/intrinsics/helper.test.gts
```diff
@@ -82,6 +100,12 @@ const NoopGenericTest = <template>
 
   const CurriedGenericPosTest = <template>
     {{#let (helper BoundMakeArray 123) as |makeNumberArray|}}
+      {{! KNOWN LIMITATION: currying a generic positional arg via {{helper}}
+          does not currently produce a correctly-typed bound helper, so
```
NullVoxPopuli: so lets fix it

NullVoxPopuli: (in a separate PR)

test-packages/ts-gts-7-1-app/src/globals/array.gts
```diff
@@ -3,7 +3,7 @@
     {{item}}
   {{/each}}
 
-  {{#each (array "a" 1 (hash)) as |item|}}
+  {{#each (array "a" 1 (hash a=1)) as |item|}}
```
NullVoxPopuli: why change this? these tests are not for hash

test-packages/ts-gts-7-1-app/src/globals/fn.gts
NullVoxPopuli, line 1: make new tests, don't change existing ones

NullVoxPopuli-ai-agent:
Thanks for the review — reworked to address all of it. Force-pushed.

The key change: I **scoped the fix to direct mustache invocations** (the exact case in the issue). The three fixture failures my first attempt ran into were all **sub-expressions** — `(fn handler greet)`, `(hash)`, `(makeNumberArray)` — whose synthetic-callee diagnostics are now left dropped, as before. So:

- **`array.gts`** (`why change this? these tests are not for hash`) — reverted, untouched. The `(hash)` sub-expression is no longer affected.
- **`fn.gts`** (`make new tests, don't change existing ones`) — reverted, untouched. Added a new test file `helper-invocation-args.test.gts` covering the class-based, functional, and plain-function helper forms from the issue.
- **`helper.test.gts`** (`so lets fix it (in a separate PR)`) — reverted, untouched. The generic-positional-currying gap it exposed is a sub-expression case, so it's no longer surfaced here and can be fixed separately as you noted.

Net diff is now just the mapping mechanism + `emitResolve` opting mustache invocations into it, plus the one new test file — no existing fixtures changed.

For the curious: the diagnostic was being dropped because TS anchors `"Expected N arguments"` on the synthetic `resolve(foo)` callee, and `toVolarMappings` only emitted zero-length boundary mappings there, so Volar couldn't translate it back. The new `wideVerification` mapping covers the callee's full generated span (via `generatedLengths`) mapped to the tight `path` node, keeping `@glint-expect-error` scoping intact.

test-packages/package-test-core/__tests__/type-tests/intrinsics/helper-invocation-args.test.gts
```diff
@@ -0,0 +1,48 @@
...
+
+  {{! @glint-expect-error: missing required positional arg }}
+  {{ClassHelper str="ok"}}
+  {{! @glint-expect-error: missing required positional arg }}
+  {{functionalHelper str="ok"}}
+  {{! @glint-expect-error: missing required positional arg }}
+  {{plainFunctionHelper str="ok"}}
+
+  {{! @glint-expect-error: positional arg is the wrong type }}
+  {{ClassHelper "nope" str="ok"}}
```
NullVoxPopuli, line 44: why not also test functionHelper and plainFunction helper here as well?

NullVoxPopuli-ai-agent: Good call — done. Added wrong-type cases (positional and named) for `functionalHelper` and `plainFunctionHelper` alongside `ClassHelper`, so all three forms now cover both the missing-positional-arg and wrong-arg-type paths.

NullVoxPopuli-ai-agent:
CI was red on two existing snapshots in `package-test-core` — both updated to reflect the fix (not to paper over failures):

- **`diagnostic-augmentation.test.ts` → "expected argument count"** — this one's a nice validation. Its fixture has `{{expectsTwoArgs "one"}}` (too few) and `{{expectsAtLeastOneArg}}` (too few), and the old snapshot only ever captured the *too-many*-args errors. The fix now also surfaces `Expected 2 arguments, but got 1.` and `Expected at least 1 arguments, but got 0.` (with the "argument for 'b' was not provided" related info), so the snapshot grew to include them.
- **`rewrite.test.ts`** — the mapping debug snapshots gain the `PathExpression` callee-wrapper node that `emitResolve` now emits for mustache invocations. Emitted TS text is unchanged (so `template-to-typescript.test.ts` needed no update).

Full `package-test-core` suite is green locally now.

test-packages/package-test-core/__tests__/type-tests/intrinsics/helper-invocation-args.test.gts
NullVoxPopuli, line 1: we should add another file, similar to this one, but with optional named args, and then also assess what is allowed with optional positional args

NullVoxPopuli-ai-agent: Added `helper-invocation-optional-args.test.gts`. It covers:

- **Optional named arg** (`str?`) — may be omitted (`{{OptionalNamedHelper 1}}`) or provided, and is still type-checked when provided.
- **Optional trailing positional arg** (`second?`) — same: may be omitted or provided, still type-checked.
- Required positional args remain required in both cases.

And the interesting find while assessing optional **positional** args: an optional positional that sits *before* a required named-args hash currently **can't** be omitted. `{{OptionalPositionalThenNamedHelper 1 flag=true}}` (dropping the optional `label`) errors with `Expected 3 arguments, but got 2`, because Glint models the named args as the final positional parameter — so dropping the optional positional shifts the hash into its slot. I captured that as a documented `@glint-expect-error` with a comment; happy to file it as a separate issue if you'd like it tracked for a real fix.

NullVoxPopuli-ai-agent:
Small cleanup: while double-checking that everything in the diff is load-bearing, I found one dead branch and removed it.

In `transformed-module.ts`'s `recurse()`, the `wideVerification` → `pushWide` handling was in both the parent (`children.length > 0`) and leaf (`children.length === 0`) branches. The leaf one is unreachable: `wideVerification` is only set on the `forNode(node.path, …)` wrapper from `emitResolve`, and `emitExpression(node.path)` always emits at least one child mapping, so a flagged node always has children and always takes the parent branch.

Confirmed empirically by swapping the leaf body for a `throw` and running the full suite — 0 hits across all `package-test-core` test files plus the type-tests, with diagnostics still surfacing correctly. Removed the 4 dead lines and rebuilt the branch into two clean commits; CI is green.

NullVoxPopuli (approved):

PR #1174, opened 2026-06-29
Handle `''` specially in the `element` helper

boris-petrov:
I've written in the comment why I think this is much better. Also [here](https://github.com/typed-ember/glint/pull/1167#issuecomment-4815797703).

cc @NullVoxPopuli

### DIFF
--- a/test-packages/ts-gts-7-1-app/src/globals/element.gts
+++ b/test-packages/ts-gts-7-1-app/src/globals/element.gts
@@ -15,6 +15,8 @@ const rect = null as unknown as ElementComponent<SVGRectElement>;
 const empty = null as unknown as ElementComponent<null>;
 const anyElement = null as unknown as ElementComponent<Element>;
 
+const videoOrNot = null as unknown as 'video' | '';
+
 const dynamicTag: string = 'div';
 
 <template>
@@ -34,6 +36,9 @@ const dynamicTag: string = 'div';
 
   {{! Empty string renders no wrapping element -> `null` (no element) }}
   {{expectTypeOf (element "") to.equalTypeOf empty}}
+
+  {{! A tag or empty is typed as the tag }}
+  {{expectTypeOf (element videoOrNot) to.equalTypeOf video}}
 
   {{! A dynamic (non-literal) tag name falls back to the base `Element` }}
   {{expectTypeOf (element dynamicTag) to.equalTypeOf anyElement}}
--- a/test-packages/ts-gts-7-1-app/src/globals/element.gts
+++ b/test-packages/ts-gts-7-1-app/src/globals/element.gts
@@ -40,8 +40,8 @@ declare function narrow(loose: string): loose is 'input' | 'button';
   {{! Empty string renders no wrapping element -> `null` (no element) }}
   {{expectTypeOf (element "") to.equalTypeOf empty}}
 
-  {{! A tag or empty is typed as the tag }}
-  {{expectTypeOf (element videoOrNot) to.equalTypeOf video}}
+  {{! A tag or empty -- because element can't know which will be received }}
+  {{expectTypeOf (element videoOrNot) to.equalTypeOf anyElement}}
 
   {{! A dynamic (non-literal) tag name falls back to the base `Element` }}
   {{expectTypeOf (element dynamicTag) to.equalTypeOf anyElement}}
@@ -84,7 +84,7 @@ declare function narrow(loose: string): loose is 'input' | 'button';
       no element for attributes to apply to }}
   {{#let (element "") as |Tag|}}
     <Tag>hello</Tag>
-    {{! @glint-expect-error: a tagless element has no attributes }}
+    {{! @glint-expect-error: a tagless element has no attributes (no element rendered) }}
     <Tag id="foo" />
   {{/let}}
 </template>
--- a/test-packages/ts-gts-7-1-app/src/globals/element.gts
+++ b/test-packages/ts-gts-7-1-app/src/globals/element.gts
@@ -9,6 +9,7 @@ type ElementComponent<El> = ComponentLike<{
 }>;
 
 const video = null as unknown as ElementComponent<HTMLVideoElement>;
+const inputOrButton = null as unknown as ElementComponent<HTMLInputElement | HTMLButtonElement>;
 const anchor = null as unknown as ElementComponent<HTMLAnchorElement>;
 const svg = null as unknown as ElementComponent<SVGSVGElement>;
 const rect = null as unknown as ElementComponent<SVGRectElement>;
@@ -18,6 +19,8 @@ const anyElement = null as unknown as ElementComponent<Element>;
 const videoOrNot = null as unknown as 'video' | '';
 
 const dynamicTag: string = 'div';
+
+declare function narrow(loose: string): loose is 'input' | 'button';
 
 <template>
   {{! ---- (RFC 389) element ---- }}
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
thanks for writing up your thoughts!

I have a concern tho -- when the source for element helper becomes the canonical type, we're back at '' => null (as is the runtime behavior)

boris-petrov:
Sorry, not sure I follow - can you explain in a bit more detail and/or give an example?

NullVoxPopuli:
the long term plan for the types in glint is to bring them all in to ember-source -- where the `element` helper lives. 

right now, the element helper is typed like this: 
- https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/helper/index.ts#L575
  - (fake types for now)
- implementation comes from: https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/helper/index.ts#L22-L23
- implementation: https://github.com/emberjs/ember.js/blob/main/packages/%40ember/-internals/glimmer/lib/helpers/element.ts
  - when tagName is falsey (`''`), we return null: https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/-internals/glimmer/lib/helpers/element.ts#L42 
  - here is where we look up the component used for element: https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/-internals/glimmer/lib/helpers/element.ts#L132
    - https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/-internals/glimmer/lib/helpers/element.ts#L73
    - here is where we set the component manager: https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/-internals/glimmer/lib/helpers/element.ts#L65
      - (which is how this ends up getting called, through all the component manager machinery) https://github.com/emberjs/ember.js/blob/faeea2fe31f2f2c86669a08f33e567905aa220de/packages/%40ember/-internals/glimmer/lib/helpers/element.ts#L41 -- null == no element, which means no attributes can be applied --- so when we define a public type for element eventually, this null relationship / _no element_ will need to be correctly represented so that our tooling doesn't lie about the runtime behavior 



can you describe more about the use case you're running in to? It's clear you've found a rough edge, and I'd like to know more about the types in the component/template you're working in

thanks!

boris-petrov:
Well, it's pretty much what I've written in the comment. You can imagine that no one will type `(element '')` - that's pointless. Every use-case of `element` that allows `''` will be combined with other possible elements - say `div` or `keyof HTMLElementTagNameMap`. For example, most components that I've created that use `element` have the following type: `titleTagName: keyof HTMLElementTagNameMap | ''` - which means the user of this component can either pass some HTML tag name or an empty string. In the implementation I do:

```hbs
{{#let (element @titleTagName) as |TitleTagName|}}
  <TitleTagName class="foo">
    ...
  </TitleTagName>
{{/let}}
```

Without this commit, this doesn't compile because `TitleTagName` will include `null` because of the `''`. My point is that this is not useful. Passing `class` isn't **wrong** - it just won't do anything if `''` is passed - but it will do something when a tag name is passed. Both cases are fine. I don't want to have an error on such code.

NullVoxPopuli:
sounds good to me -- so the problem is that the argument passed to `element` is _too wide_, which **accidentally** includes `''`.

I think your change probably fine then if you can get tests passing.

> Without this commit, this doesn't compile because

a bit of pendanticness -- type checking isn't compilation -- your code compiles just fine via babel <3 
(tho, I understand this is still frustrating)


Could you comment on the RFC, linking back to this PR as the fix?
- https://github.com/emberjs/rfcs/pull/389

since the RFC predates any types discussions, it would be good to cross-reference

boris-petrov:
I’ll fix the tests but just to make it clear - the runtime behavior is totally fine, correct and makes sense. I’m just saying here that the types following the runtime behavior to the letter is not useful - that’s what I want to fix here - to make the types more ergonomic.

NullVoxPopuli:
sounds good, and I now agree

boris-petrov:
Tests (and implementation) fixed.

test-packages/ts-gts-7-1-app/src/globals/element.gts
NullVoxPopuli, line 1: I just had a thought -- which _is_ contrary to our earlier discussion --- something I want to check on is if, since this whole PR is for the "dynamic string" case, if it makes sense to actually use a narrowed string, e.g.:

```glimmer-ts
function narrow(loose: string): loose is 'div' | 'button' {
  assert(`[BUG]: debug this`, loose === 'div' || loose === 'button');
  
  return loose;
}

class Demo extends Component {
  get tagName(): 'div' | 'button' {
    return narrow(this.args.tagName);
  }
  
  <template>
    {{#let (element this.tagName) as |Tag|}}
      <Tag /> div or button
    {{/let}}
  </template>
}  
```

NullVoxPopuli: or in TO:
```glimmer-ts
function narrow(loose: string): loose is 'input' | 'button' {
  assert(`[BUG]: debug this`, loose === 'input' || loose === 'button');
  
  return loose;
}


<template>
  {{#let (element (narrow @tagName)) as |Tag|}}
    <Tag value="hello there" /> div or button
  {{/let}}
</template>

```

^ this case in particular would fail without the narrowing, if we went forward with this PR, I think? 🤔 
(gonna pull down this branch tomorrow)

boris-petrov: No, why would that break? I've pushed a test for that case.

boris-petrov:
In addition to adding the test case for the comment above, I changed how `''` is handled - by narrowing-down to `never`. This allows `'video' | ''` to be typed as `HTMLVideoElement` which is the most useful for a programmer.

NullVoxPopuli:
I think I found a problem:


```glimmer-ts

import type { ComponentLike } from '@glint/template';
import type { TOC } from '@ember/component/template-only';

const Receiver: TOC<{ 
  Args: {
    nev: ComponentLike<{ Element: null | never }>,
    foo: ComponentLike<{ Element: Element }>,
    bar: ComponentLike<{ Element: HTMLElement }>,
    baz: ComponentLike<{ Element: HTMLVideoElement }>,
    zoo: ComponentLike<{ Element: HTMLImageElement }>,
  }}> = <template>
</template>;

const str = 'video';

export const Foo = <template>
  <Receiver 
    @nev={{element str}}
    @foo={{element str}}
    @bar={{element str}}
    @baz={{element str}}
    @zoo={{element str}}
  />
</template>;

export const Foo2 = <template>
  <Receiver 
    @nev={{element "video"}}
    @foo={{element "video"}}
    @bar={{element "video"}}
    @baz={{element "video"}}
    @zoo={{element "video"}}
  />
</template>;

// all of these pass null at runtime, yet don't error, and the Receiver
// is expecting a Component, which the type system thinks is happening,
// but it is not what is happening
export const Foo3 = <template>
  <Receiver 
    @nev={{element ""}}
    @foo={{element ""}}
    @bar={{element ""}}
    @baz={{element ""}}
    @zoo={{element ""}}
  />
</template>;
```

it seems like types got too wide

boris-petrov:
OK, I pushed a change for the types so that a lone `''` is handled as before - with `null` - but `'video' | ''` narrows down to `HTMLVideoElement`. I believe this is the best of both worlds.

NullVoxPopuli (approved):

boris-petrov:
Thanks for merging this!

The one thing I could discuss is your decision on the `'' extends K` case to return an `Element`. The problem with this, shown by my project, is that on `Element` tags I can't specify the `style` modifier for example - it requires `Element & ElementCSSInlineStyle`. Also, my thinking for the way I had done it was as follows:

`'video' | ''` - here the programmer's idea is that there either will be a video element or there will be no wrapping element. Narrowing this down to `Element` is better than `null` but it would be even better to narrow this to `HTMLVideoElement` - this way I can use everything I want and which would work at runtime on a video element. It would also allow me to use modifiers that are made to work on `HTMLElement` instances or `ElementCSSInlineStyle` as in the `style` modifier's case.

What do you think?

PR #1175, opened 2026-06-29
Fix `lt`, `lte`, `gt` and `gte` not accepting strings and dates

boris-petrov:
cc @NullVoxPopuli


NullVoxPopuli (changes requested):
seems good -- I had to double check the the RFC was ambiguous enough  for this _and_ that TS allowed it. 

I don't see any harm in allowing this... however, I think we should use a generic for the argument types rather than just number -- that way we say that both argumnts must of the same type, which mirrors the behavior TS has with number < string, for example

boris-petrov:
Thanks for the review! However, I think I've done exactly what you suggest - there is even a test for that - that it doesn't accept operands of different types. Am I missing something?

NullVoxPopuli (approved):

NullVoxPopuli:
naw, I can't read. thank you for making me double check <3

PR #1196, opened 2026-07-03
Fix #1068: keep curried generic components from poisoning consumer inference

NullVoxPopuli-ai-agent:
Fixes #1068 (the remaining "deep" case from [hlorellium's second report](https://github.com/typed-ember/glint/issues/1068#issuecomment-4065651544))

## Where the deep case stands on `main`

The 4-layer `DataTable → Row → Cell` repro no longer produces TS2589 (the v1.2.3 `bindInvokable` fix handled that), but it still fails: `Row`'s `T` collapses to the constraint `Identifiable`, so the yielded row item loses its type.

The mechanism: `bindInvokable` deliberately keeps the curried component's own type parameter **free** — that's what lets `(component PickerOption onSelect=@onSelect)` satisfy a generic-shaped `WithBoundArgs<typeof PickerOption, 'onSelect'>` target, and the existing #1068 test locks that in. But when the curried value is passed as an *arg* to another generic component, TypeScript instantiates that free generic to its constraint during inference, and the collapsed candidate beats the correct candidates from sibling args (`@item={{item}}`).

## Why not a type-level fix

I probed five signature shapes for an "instantiating binder" that would pin the curried generic from the bound named args (decomposing with constraint ties, `Given & Rest` intersection params, sequenced two-call currying to fix `GivenNamed` first, plus intersecting holistic & instantiated results). None work: TypeScript only performs generic-signature instantiation in context when the target signature is otherwise fully concrete, and any shape expressive enough to compute the rest-args necessarily has free outer type params, which disables it. The two goals — generic-shaped targets need the *holistic* type, inference-sensitive arg positions need *no* type — cannot be met by one value type.

They *can* be met positionally, which is what this PR does, building on the direction @johanrd explored in [johanrd/glint#1](https://github.com/johanrd/glint/pull/1):

- `ScopeStack` bindings carry metadata; `{{#let}}` block params initialized from `bind-invokable` subexpressions with named args are marked. Metadata lives in the scope frames themselves, so an inner shadowing binding (`{{#let Other as |Bound|}}`) correctly clears the mark — a bug in the ref-counted side-map approach.
- References to marked bindings are cast **in arg position only** to a new `InferenceInertInvokable` (`Invokable<(...args: any[]) => any>`) rather than bare `any`: the value contributes no inference candidates, but it's still invokable-shaped, so passing it where no invokable belongs remains an error. JSDoc-cast form covers untyped (`.gjs`) scripts.
- Yield positions (and everything else) keep the fully-typed reference, so `WithBoundArgs<typeof C, K>`-shaped targets keep working — both direct `{{yield (component ...)}}` and `{{#let ... as |C|}}{{yield (hash Option=C)}}`, covered in the new test.

## Known trade-off

At the cast positions, compatibility between the curried component's signature and the consuming arg's declared type is not checked (stated in the docs on `InferenceInertInvokable` and the emit). Sibling-arg checking is unaffected — the showcase test's `@glint-expect-error` cases (wrong `onSelect` callback, wrong `items` element type, and `user.price` on the yielded param) all still fire, which also proves `T=User` genuinely flows through the currying.

## Tests

- `component-curry-generic.test.gts`: hlorellium's full showcase (fails on `main` at the `Row` boundary) plus case-A guards for both yield shapes.
- Transform unit tests: cast in arg position with `direct:`/yield untouched, shadowing correctness, and the JSDoc form.
- All package typechecks, vitest suites, and lint pass (VS Code smoke timeout is the known pre-existing environmental failure).

### DIFF
--- a/packages/ember-tsc/src/transform/template/template-to-typescript.ts
+++ b/packages/ember-tsc/src/transform/template/template-to-typescript.ts
@@ -1207,7 +1207,34 @@ export function templateToTypescript(
         // component/helper, and returned as a value otherwise.
         let hasParams = Boolean(node.hash.pairs.length || node.params.length);
         if (!hasParams && position === 'arg' && !isGlobal(node.path)) {
-          emitExpression(node.path);
+          // A `{{#let}}`-bound curried component passed as an arg must not
+          // contribute inference candidates to the consuming component's type
+          // parameters: `bindInvokable` keeps the curried component's own
+          // generic free, and TypeScript instantiates a free generic to its
+          // constraint during inference, which then beats the correct
+          // candidates supplied by sibling args (#1068). The cast keeps the
+          // value invokable-shaped while making it inference-inert; in every
+          // other position (notably `{{yield}}`) the reference stays fully
+          // typed so generic-shaped targets like `WithBoundArgs<typeof C, K>`
+          // keep working.
+          if (
+            node.path.type === 'PathExpression' &&
+            node.path.head.type === 'VarHead' &&
+            node.path.tail.length === 0 &&
+            scope.isCurriedInvokable(node.path.head.name)
+          ) {
+            if (useJsDoc) {
+              mapper.text(`(/** @type {import("${typesModule}").InferenceInertInvokable} */ (`);
+              emitExpression(node.path);
+              mapper.text('))');
+            } else {
+              mapper.text('(');
+              emitExpression(node.path);
+              mapper.text(` as import("${typesModule}").InferenceInertInvokable)`);
+            }
+          } else {
+            emitExpression(node.path);
+          }
         } else if (position === 'top-level') {
           // e.g. top-level mustache `{{someValue}}`
           mapper.text('__glintDSL__.emitContent(');
@@ -1386,7 +1413,7 @@ export function templateToTypescript(
         mapper.text(');');
         mapper.newline();
 
-        emitBlock('default', node.program);
+        emitBlock('default', node.program, curriedLetBindings(node));
 
         if (node.inverse) {
           emitBlock('else', node.inverse);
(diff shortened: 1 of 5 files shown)


packages/ember-tsc/types/-private/dsl/index.d.ts
```diff
@@ -93,3 +93,30 @@ export declare function bindPositional<F, Bound extends unknown[]>(
...
+ * instantiation in context only fires when the target signature is otherwise
+ * fully concrete — so instead the transform casts the reference to this
+ * inference-inert type at exactly those use sites. The value contributes no
+ * inference candidates (so the consumer's type parameter is inferred from its
+ * other args), while remaining an `Invokable`, so passing it where no
+ * invokable belongs is still an error. The trade-off: compatibility between
+ * the curried component's signature and the consuming arg's declared type is
+ * not checked at that position.
+ */
+export type InferenceInertInvokable = Invokable<(...args: any[]) => any>;
```
NullVoxPopuli, line 122: would this not lose type information?

packages/ember-tsc/src/transform/template/template-to-typescript.ts
```diff
@@ -1207,7 +1207,34 @@ export function templateToTypescript(
...
+          // other position (notably `{{yield}}`) the reference stays fully
+          // typed so generic-shaped targets like `WithBoundArgs<typeof C, K>`
+          // keep working.
+          if (
+            node.path.type === 'PathExpression' &&
+            node.path.head.type === 'VarHead' &&
+            node.path.tail.length === 0 &&
+            scope.isCurriedInvokable(node.path.head.name)
+          ) {
+            if (useJsDoc) {
```
NullVoxPopuli: This is kind of a bonkers way to do this, I am not aware of any other intermediate format branch that uses JSDoc here

NullVoxPopuli (changes requested):
.

packages/ember-tsc/types/-private/dsl/index.d.ts
NullVoxPopuli-ai-agent: Yes — deliberately, and only at these exact positions (a `{{#let}}`-bound curried component passed as an arg). It's load-bearing rather than incidental: any type information the reference carries becomes an inference candidate, and for a curried *generic* component that candidate is the constraint-collapsed instantiation, which is precisely what breaks the consumer's `T` (`Row<T>` inferring `Identifiable` in the issue's repro).

I tried hard to avoid the loss — five signature shapes for a binder that would instead *pin* the curried generic from the bound args, summarized in the PR description. None can work: TS only instantiates a generic source signature in context when the target signature is otherwise fully concrete, and the two consumers have contradictory needs (generic-shaped `WithBoundArgs<typeof C, K>` targets require the free generic; arg-position inference requires no candidates). Positional treatment is the only split I found that satisfies both.

What's retained: the value is still `Invokable`-shaped (passing it where no invokable belongs errors), every other position including `{{yield}}` keeps the fully-typed reference, and sibling-arg checking is unaffected — the showcase test's `@glint-expect-error` cases all still fire. What's lost is only the check of the curried signature against the consuming arg's declared type at that position. Happy to explore a different trade if you see one I've missed.

packages/ember-tsc/src/transform/template/template-to-typescript.ts
NullVoxPopuli-ai-agent: Fair — dropped in be176dfb. Untyped scripts now emit the plain reference (no cast at all): the collapse this guards against requires a generic class component, which is a TypeScript-only concern in practice, so `.gjs` just keeps status quo behavior. The unit test now asserts the JSDoc path emits uncast.

NullVoxPopuli (approved):

Issue #1193, opened 2026-07-03
`eq`/`neq`: TS2367 "no overlap" errors silently dropped since 1.8.6 (regression via #1179)


NullVoxPopuli:
# `eq`/`neq`: TS2367 "no overlap" errors silently dropped since 1.8.6 (regression via #1179)

## Summary

Since `@glint/ember-tsc` 1.8.6, a `(eq x "literal")` comparison where the literal has no overlap with `x`'s type no longer reports TS2367. The error is still produced by TypeScript in the transformed module — it's dropped during diagnostic mapping back to the `.gts` source. As a side effect, existing `{{! @glint-expect-error }}` directives covering these comparisons now report TS2578 "Unused '@ts-expect-error' directive", tempting users to delete suppressions that still guard real type errors.

1.8.5 behaves correctly. Bisected to 1.8.6; the cause is #1179 (control-flow narrowing through operator special forms).

## Reproduction

```jsonc
// package.json
{
  "devDependencies": {
    "@glimmer/component": "^2.0.0",
    "@glint/ember-tsc": "1.8.6",
    "@glint/template": "1.7.8",
    "ember-source": "7.1.0",
    "typescript": "6.0.3"
  }
}
```

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "bundler",
    "strict": true,
    "noEmit": true,
    "skipLibCheck": true,
    "types": ["ember-source/types", "@glint/ember-tsc/types"]
  },
  "include": ["src"]
}
```

```gts
// src/demo.gts
import Component from '@glimmer/component';

export default class Demo extends Component {
  kind: 'a' | 'b' = 'a';

  <template>
    {{#if (eq this.kind "z")}}one{{/if}}

    {{! @glint-expect-error this.kind can never be "z" }}
    {{#if (eq this.kind "z")}}two{{/if}}
  </template>
}
```

Run `ember-tsc --noEmit`.

**With 1.8.5 (expected):**

```
src/demo.gts(7,15): error TS2367: This comparison appears to be unintentional because the types '"a" | "b"' and '"z"' have no overlap.
```

(1 error on the unsuppressed comparison; the directive on the second one is honored.)

**With 1.8.6 through 1.8.10 (actual):**

```
src/demo.gts(9,5): error TS2578: Unused '@ts-expect-error' directive.
```

(The TS2367 on line 7 is gone, and the directive flips to unused.)

[...]

PR #1203, opened 2026-07-25
Report dropped diagnostics on quoted attribute keys via `wideVerification`

johanrd:
Dashed attribute and component-arg names are emitted as *quoted* object keys; TS anchors TS2353 on the full quoted key, whose quotes are generated-only text, so Volar dropped the diagnostic — e.g. `<circle stroke-widht='2'>` and `<Inner @bogus-arg='x' />` both type-checked clean.

Fix: opt plain-attribute emission (`emitPlainAttributes`) and component-arg emission (`emitComponent`) into the existing `wideVerification` mechanism (#1168), mirroring #1193. Two failing tests included.

Note for reviewers: this surfaces TS2353 for valid SVG 2 attributes missing from the `svg-element-attributes` dataset (e.g. `vector-effect`) until the upstream data fix lands — may warrant sequencing after a dep bump.

Cowritten by Claude

### DIFF
--- a/packages/ember-tsc/src/transform/template/template-to-typescript.ts
+++ b/packages/ember-tsc/src/transform/template/template-to-typescript.ts
@@ -879,27 +879,39 @@ export function templateToTypescript(
             mapper.text('{ ');
 
             for (let attr of dataAttrs) {
-              mapper.forNode(attr, () => {
-                mapper.newline();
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):
looks good, needs conflict resolution tho

PR #1204, opened 2026-07-25
Add regression coverage for modifier diagnostics anchored on generated code

johanrd:
Test-only. Three modifier cases in the #1168/#1193 family (diagnostics anchored on generated scaffolding) that work today but had no coverage: element-type mismatch with and without args (TS2345 on `__glintY__.element`), and missing positional args (TS2554 on the `resolve(...)` callee).

Cowritten by Claude


NullVoxPopuli (approved):
looks good, needs conflict resolution tho

NullVoxPopuli (approved):

PR #1205, opened 2026-07-25
Report dropped {{yield}} diagnostics (arity + undeclared block target)

johanrd:
Two dropped-diagnostic bugs in `emitYieldExpression`, shown by failing tests in the first commit:

- `{{yield 'one' to='body'}}` against `Blocks: { body: [a: string, b: number] }` reported nothing — TS2554 anchors on the generated `yieldToBlock(...)` callee. Fixed with the `wideVerification` opt-in (#1168).
- `{{yield to='nope'}}` reported nothing — the block name is a generated string with no mapping. Fixed by mapping it to the `to='...'` literal, giving a precise anchor.

The `expected argument count` snapshot gains the two diagnostics its fixture always deliberately included but which were silently dropped.

Found via a coverage audit against vue-language-tools' tsc fixtures.

Cowritten by Claude

### DIFF
--- a/packages/ember-tsc/src/transform/template/template-to-typescript.ts
+++ b/packages/ember-tsc/src/transform/template/template-to-typescript.ts
@@ -1235,40 +1235,64 @@ export function templateToTypescript(
       node: AST.MustacheStatement | AST.SubExpression,
       position: InvokePosition,
     ): void {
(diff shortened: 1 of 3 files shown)


NullVoxPopuli (approved):
looks good, needs conflict resolutoin tho

PR #1210, opened 2026-07-25
Regression fix: Add missing SVG 2 / Filter Effects attributes to generated element types by bumping `svg-element-attributes` to `2.2.0`

johanrd:
#1203 shipped in `@glint/ember-tsc@1.8.13` before https://github.com/wooorm/svg-element-attributes/pull/5 was released, so valid attributes absent from the dataset started reporting TS2353 — `<circle vector-effect='non-scaling-stroke'>`, and `flood-color`/`flood-opacity` on `<feDropShadow>`.

`svg-element-attributes@2.2.0` crawls the SVG 2 presentation attributes table, so its `*` entry now carries the attributes valid on any element in the SVG namespace (12 → 73). Every generated SVG element interface extends `GlobalSVGAttributes`, so `<feDropShadow>` picks up `flood-color`/`flood-opacity` by inheritance — its own per-element list is unchanged upstream.

**The regression test in this PR fails on current `main` (three TS2353s) and passes with the bump.**

Note: since `elements.d.ts` lives in `@glint/template`, this needs a `@glint/template` release to reach consumers.

Cowritten by Claude

### DIFF
--- a/bin/build-elements.mjs
+++ b/bin/build-elements.mjs
@@ -42,25 +42,6 @@ const runtimeAdditionalProperties = new Map([
   ['SVGSVGElement', new Map([['xmlns', 'AttrValue']])],
 ]);
 
-// Valid attributes missing from svg-element-attributes@2.1.0, whose crawler
-// misses the SVG 2 presentation attributes list (fix proposed upstream:
-// https://github.com/wooorm/svg-element-attributes/pull/5 — remove these
-// once a release includes it and the dependency is bumped).
-const missingSpecAttributes = new Map([
-  // SVG 2 presentation attribute; may be specified on any SVG element:
-  // https://svgwg.org/svg2-draft/coords.html#VectorEffectProperty
-  ['GlobalSVG', new Map([['vector-effect', 'AttrValue']])],
-  // Filter Effects 1 applies these to <feDropShadow>, not just <feFlood>:
-  // https://drafts.fxtf.org/filter-effects/#feDropShadowElement
-  [
-    'SVGFEDropShadowElement',
-    new Map([
-      ['flood-color', 'AttrValue'],
-      ['flood-opacity', 'AttrValue'],
-    ]),
-  ],
-]);
-
 const GLOBAL_HTML_ATTRIBUTES_NAME = 'GlobalHTMLAttributes';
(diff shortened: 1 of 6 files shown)


bin/build-elements.mjs
```diff
@@ -42,6 +42,25 @@ const runtimeAdditionalProperties = new Map([
   ['SVGSVGElement', new Map([['xmlns', 'AttrValue']])],
 ]);
 
+// Valid attributes missing from svg-element-attributes@2.1.0, whose crawler
+// misses the SVG 2 presentation attributes list (fix proposed upstream:
+// https://github.com/wooorm/svg-element-attributes/pull/5 — remove these
+// once a release includes it and the dependency is bumped).
+const missingSpecAttributes = new Map([
```
NullVoxPopuli: we don't actually want to own any of these if the upstream package is willing to have them.

the above additional types are for ember-y reasons, but AttrValue (I think) is our default for everything, so we can simplify a lot.

Reason being is that we don't want to have churn on this repo when the responsibility for something is outside the repo.

NullVoxPopuli (changes requested):
see comment

bin/build-elements.mjs
johanrd: Yes, that is exactly why I added the upstream PR: https://github.com/wooorm/svg-element-attributes/pull/5

However, since #1203 introduced false positives, our options now are:

1) Revert #1203, and wait with the catching-typos-and-bogus-attrs fix until upstream is merged
2) Add a local temporary patch (This PR) that handles the FPs until the upstream PR is merged
3) Just wait for upstream, and by that accept false positives in `@glint/template 1.7.9`

NullVoxPopuli: what is @wooorm's average responsiveness on those kinds of things?

johanrd: I guess some days [[1]](https://github.com/wooorm/property-information/pull/26), [[2]](https://github.com/wooorm/html-element-attributes/pull/15), though the proposed fix in https://github.com/wooorm/svg-element-attributes/pull/5 is a different kind

johanrd:
@NullVoxPopuli https://github.com/wooorm/svg-element-attributes/releases/tag/2.2.0 is now released, so that simplifies🎉

NullVoxPopuli (approved):

Issue #81, opened 2021-03-29
Bug with Mixin usage


boris-petrov:
Check [this repo](https://github.com/boris-petrov/glint-issues) (actually [this](https://github.com/boris-petrov/glint-issues/commit/556fe399e1014e73746a24851a8095df50a56651) specific commit). It adds a mixin and a component that uses it.

Running `./node_modules/.bin/tsc --noEmit` leads to no errors. However, running `./node_modules/.bin/glint` causes:

```
app/components/test-component.ts:4:44 - error TS4020: 'extends' clause of exported class 'TestComponent' has or is using private name 'ICalendar'.

4 export default class TestComponent extends Component.extend(Calendar) {
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~

app/components/test-component.ts:4:44 - error TS4020: 'extends' clause of exported class 'TestComponent' has or is using private name 'ICalendar'.

4 export default class TestComponent extends Component.extend(Calendar) {
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

There shouldn't be a difference here as there are is no HBS involved so `Glint` should behave exactly the same as `tsc`.

P.S. Exporting the `ICalendar` interface fixes the errors however I still believe this is a bug because of the discrepancy between the two type-checkers.

dfreeman:
Usually those errors will only pop up with `--declaration`, because you wind up with an expression TypeScript doesn't know how to emit a type for. If you're seeing them from the `glint` CLI _without_ using `--declaration`, that's definitely a bug. Thanks for the report!

PR #1226, opened 2026-08-24
Add rewriteModuleStandalone, a TypeScript-free entry to the transform

### LINKED ISSUE #1224: Make rewriteModule usable without a TypeScript 5/6 JS API
`rewriteModule` takes a `TSLib` and parses the content-tag-preprocessed source with it. For consumers that pair the transform with TypeScript 7 (which ships no JS API), this forces a private TypeScript 5 dependency next to the compiler that does the real type-checking: ember-content-mapper carries `typescript@~5.9` purely so Glint's transform can parse (see NullVoxPopuli/ember-content-mapper#1 review discussion).

The transform uses the TS AST for three things:

1. Locating the synthesized `___T\`...\`` tagged templates and their positions (`calculateTaggedTemplateSpans`).
2. Deciding the `this` binding for each template (`templateThisBinding`: backing class vs lexical).
3. Reading import statements so lexical imports win over globals and special forms (`collectImportedBindings`, added for #1113).

(1) and most of (2) are already available without a parser: content-tag's own `parse` output reports each template's exact ranges and `type: 'expression' | 'class-member'`, and `preprocess.ts` computes `transformedStart`/`transformedEnd` from it today. The TS AST mostly re-derives what content-tag already said.

(3) needs an import scanner but not a type-checker-grade parser. `es-module-lexer` (the scanner Vite uses) reports import specifiers and named bindings from source text with no TypeScript dependency, and the script portions of a `.gts` are valid enough for it after content-tag preprocessing.

## Compatibility constraint

This has to be non-breaking: the tsserver-plugin and Volar language-server paths keep passing their `TSLib` and must keep working unchanged. Two shapes that satisfy that:

- Keep `rewriteModule(ts, input, environment, clientId)` as-is and add a parallel entry point (for example `rewriteModuleStandalone(input, environment)`) that derives template placement from content-tag and imports from a lightweight scanner. The existing callers never change; embedders opt into the new entry.
- Or make the internal script analysis pluggable: the current TS-AST analysis stays the default, and the same analysis interface can be fed from content-tag + `es-module-lexer`. `rewriteModule`'s signature and behavior stay identical for every current caller.

Either way `ember-tsc` itself keeps its TypeScript peer for type-checking; only embedders of the transform (content mappers, build plugins) shed the parser dependency. One thing to verify before choosing: whether the two analyses agree on edge cases the TS AST currently disambiguates (templates in arrow-function class fields, nested classes, static blocks) — the transform's existing test corpus covers most of these.

NullVoxPopuli-ai-agent:
Fixes #1224

Adds `rewriteModuleStandalone(input, environment)`, exported from a new `@glint/ember-tsc/transform/standalone` entrypoint. It produces the same `TransformedModule` as `rewriteModule(ts, input, environment)` without a TypeScript JS API, so embedders that type-check with TypeScript 7 (content mappers, build plugins) no longer need a private TypeScript 5 install just to run the transform.

**How it works.** `rewriteModule` reads three things off the TS AST: template placement, the template's `this` binding, and import bindings. The standalone entry gets them from [ember-estree](https://github.com/NullVoxPopuli/ember-estree) (content-tag + oxc-parser) through its `visitors`: `ImportDeclaration` for bindings, `GlimmerTemplate` for placement, walking the visitor path for the `this` binding. Template bodies are blanked (same length) before the call: only the script is analyzed there, the templates are parsed by glint itself, and a template that does not parse then cannot abort the analysis. That decision is shared with the TS path (`thisBindingFromAncestors`). Template ranges and the class-member flag still come from content-tag via `preprocess`. ember-estree is a regular dependency of `@glint/ember-tsc`, but only the `./transform/standalone` entrypoint imports it, so the language server, tsserver plugin and `ember-tsc` never load oxc (verified: importing the main and `./transform` entries loads neither ember-estree nor oxc-parser).

**Compatibility.** `rewriteModule` keeps its signature and behaviour. Internally, `calculateTemplateSpans` now takes a plain `EmbeddedTemplate` and `calculateTaggedTemplateSpans` is a thin adapter over the TS node, so both entries share the emit.

**Verification.** `rewrite-standalone.test.ts` runs both entry points over every `.gts`/`.gjs` in `test-packages` (56 files) plus targeted cases (nested classes, arrow fields, static blocks, heritage clauses, `satisfies`, ASI, confusing tokens in strings/regexes/template literals, gjs) and asserts identical `transformedContents`, errors, directives and `toDebugString()`. Same comparison over ember-primitives (132 files), kolay (48) and limber (360): all identical except limber's two deliberately broken tutorial prompts, see below.

**Mid-keystroke behaviour.** Completions while a file is invalid never depended on the transform's parser. Two things carry them: glint's template-level tolerance (handlebars errors inside `templateToTypescript` still emit the template) and, when the *script* is broken so content-tag fails, keeping the raw source with verification off so the type-checker's own recovering parser handles completions (regression-tested today). Both apply unchanged to the standalone entry, and TS 7's parser recovers like TS 5's.

oxc tolerates less than swc, so there was a third mode in between (templates emitted with no imports/placement). Removed: when ember-estree reports script errors (`ast.errors`, NullVoxPopuli/ember-estree#81) the standalone entry records a content-tag-style error and keeps the raw source, exactly like content-tag failure. One rule: script didn't parse → type-checker recovers, templates wait. Verified on `{{this.me`, `{{this.`, unclosed `{{this.`, `go = () => this.` before a template, and an unclosed call before a class: identical outcome to `rewriteModule` in every case.

**Upstream.** Depends on ember-estree 0.8.0, which carries the fixes the parity check found: #77 (content-tag result ordering), #78 (`void` placeholders for ASI), #82 (keep the `export default` wrapper) and #81 (`ast.errors`).

**Two preprocess fixes the parity check surfaced** (both also affect `rewriteModule` today):

1. content-tag reports a class's own template before a template that sits earlier in its heritage clause; `preprocess` assumed source order and spliced garbage. The parse output is now sorted.
2. A class-member template after a field initializer with no trailing semicolon (`isChecked = (v) => v === 1\n<template>…`) preprocessed to `1[___T\`…\`]`, an element access, so the class's own template was emitted as a context-bound `templateExpression` instead of `templateForBackingValue(this, …)`. Four files in limber's tutorial hit this. `preprocess` now emits `;[…]` (a `;` is an empty class element). Regression test added in `rewrite.test.ts`.

https://claude.ai/code/session_01FsXgxh6vNRWMCNcAQ3g6TN


packages/ember-tsc/src/transform/template/inlining/script-scanner.ts
NullVoxPopuli, line 1: what in the world are you doing here?


what is this for? it looks like your parsing JS/TS... why?

NullVoxPopuli-ai-agent: It re-derives what the TS AST gave `rewriteModule` (class nesting for the `this` binding, statement position for the implicit default export, import bindings) without a parser. Agreed it should not exist: switching this to ember-estree, which gives a real oxc ESTree with the templates spliced in, and deleting the scanner.

packages/ember-tsc/src/transform/template/inlining/tagged-strings.ts
```diff
@@ -9,134 +9,171 @@ import { GlintEnvironment } from '../../../config/index.js';
...
+  if (!resolveTagInfo(importedBindings, tag, environment)) {
+    return { errors: [], directives: [], partialSpans: [] };
+  }
+
+  assert(
+    ts.isNoSubstitutionTemplateLiteral(node.template),
+    'No interpolated values in template strings',
+  );
+
+  // environment-specific transforms may emit templateLocation in meta, in
```
NullVoxPopuli: we only have one environment specific transform -- this is not a generic behavior

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
```diff
@@ -0,0 +1,244 @@
...
+      visitors: (file) => {
+        let program = (file as unknown as { program: Program }).program;
+        collectImportedBindings(program, analysis.imports);
+        walk(program, [], analysis.placements);
+        analyzed = true;
+        return null;
+      },
+    });
+  } catch (error) {
+    // After the visitor factory has run, ember-estree parses each template's
```
NullVoxPopuli: is this actually a bug? I don't know what this comment is trying to say\


why do we set analyzed true on the preparation of the visitors?

and why do we collect importing bindings? does oxc not have a fault tolerant parse mode? 

does ember-estree need to add fault-tolerance?

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
```diff
@@ -0,0 +1,244 @@
...
+    // error itself is reported by `calculateTemplateSpans`.
+    if (!analyzed) throw error;
+  }
+
+  return analysis;
+}
+
+const SKIPPED_KEYS = new Set(['parent', 'loc', 'range', 'tokens', 'comments']);
+
+function walk(node: Node, ancestors: Array<Ancestor>, placements: Map<number, Placement>): void {
```
NullVoxPopuli: why are we walking here? we already have a means to do this with ember-estree

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
```diff
@@ -0,0 +1,244 @@
...
+    } else if (isNode(child)) {
+      walk(child, ancestors.concat({ node, key }), placements);
+    }
+  }
+}
+
+// ember-estree stands each expression template in with a same-length
+// template literal (`void `...`` from 0.7.1, a bare backtick literal before) and
+// each class member with a `static{`...`}` block.
+function isPlaceholderShape(node: Node): boolean {
```
NullVoxPopuli: this function name implies that we get something definite out of it, however, it will match many things that are not placeholders (just real code)

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
```diff
@@ -0,0 +1,244 @@
...
+// expression statement, optionally through a `satisfies`.
+function isExpressionStatement(ancestors: Array<Ancestor>): boolean {
+  let parent = ancestors[ancestors.length - 1]?.node;
+  if (parent?.type === 'TSSatisfiesExpression') {
+    parent = ancestors[ancestors.length - 2]?.node;
+  }
+  return parent?.type === 'ExpressionStatement';
+}
+
+// Mirrors `collectImportedBindings` in `inlining/tagged-strings.ts`.
```
NullVoxPopuli: why mirror? we don't like duplicate code if we don't need it

packages/ember-tsc/src/transform/template/inlining/tagged-strings.ts
NullVoxPopuli-ai-agent: Right, that fallback was pre-existing and dead. Replaced with an assert in be745d1.

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
NullVoxPopuli-ai-agent: Not a bug in glint, a workaround for ember-estree throwing out of `toTree` mid-walk on a template syntax error (the factory had run, so the analysis was complete, hence the flag). Gone in be745d1: NullVoxPopuli/ember-estree#80 adds `onTemplateError`, so the walk continues and this path just gets the failing template's location like any other.

Import bindings are needed for the emit itself, not tolerance: `import { on } from '@ember/modifier'` must win over the `on` keyword global (#1113), and imports of special forms (`fn`, `hash`, ...) need resolving to their source. Same as `collectImportedBindings` on the TS path.

oxc has no tolerant mode: on an unrecoverable script error `program.body` is `[]` (TypeScript's parser keeps going). That's the one documented divergence (covered by a test); the type-checker reports the syntax error either way.

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
NullVoxPopuli-ai-agent: Agreed. be745d1 uses `visitors` (`ImportDeclaration`, `GlimmerTemplate`) plus `onTemplateError`'s path; the walk is gone. That needed two ember-estree fixes (NullVoxPopuli/ember-estree#80): keep the `ExportDefaultDeclaration` wrapper (today `export default <template>` loses the export in the tree and in `print()`), and don't abort on one bad template.

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
NullVoxPopuli-ai-agent: Removed along with the walk in be745d1.

packages/ember-tsc/src/transform/template/rewrite-module-standalone.ts
NullVoxPopuli-ai-agent: Shared now: `thisBindingFromAncestors(kinds)` in `inlining/index.ts` is the one decision; `templateThisBinding` (TS AST) and `ancestorsOf` (ESTree path) only map their own AST to `static-block | heritage-clause | class`. Import reading stays per-AST since the node shapes differ, but both produce the same `ImportedBindings`.

NullVoxPopuli (approved):

PR #1228, opened 2026-08-25
VS Code extension: stand down on TypeScript 7 workspaces

NullVoxPopuli-ai-agent:
When the workspace's installed `typescript` is version 7 or newer, the Glint extension now disables itself at activation instead of starting a language server that cannot work.

Glint's language server and tsserver plugin need the TypeScript 5/6 JS API. The TypeScript 7 package is the native compiler and ships no JS API, so on a TS 7 workspace Glint can only produce broken output (in practice: syntax-error squiggles across every `<template>` tag, because the plugin never loads and tsserver parses `.gts` as plain TS). Template type-checking on TS 7 comes from a content mapper run by TypeScript itself ([ember-content-mapper](https://github.com/NullVoxPopuli/ember-content-mapper)).

What this PR does at activation:

- Resolves `typescript/package.json` from the first workspace folder, honoring `glint2.libraryPath`. Both the TS 5/6 and TS 7 packages export `./package.json`, so the resolution works across versions.
- If the major version is >= 7: logs to the "Glint2 Language Server" output channel that Glint requires TypeScript 5 or 6 and that template type-checking now comes from `ember-content-mapper` via `contentMappers` in tsconfig.json with `tsc --runExternalCode`, then returns early. The language server, the tsserver plugin configuration, and the Ember TSC status bar item never start.
- Otherwise: behavior is unchanged.

Notes for review:

- The check runs once at activation; switching the workspace TypeScript major requires a reload, which matches how the extension already treats server-affecting changes.
- The early return also skips registering `glint2.restart-language-server` and the Ember TSC source command. Since no server exists to restart or configure on these workspaces, that seemed right, but registering no-op commands instead is a small change if preferred.
- A follow-up could have the extension call the Native Preview extension's `registerContentMappers` API on these workspaces (microsoft/typescript-go#4712's LSP-activation section), so `.gts` files get TS 7 language features without opening a `.ts` file first. That needs a Native Preview build newer than 2026-08-19, which the marketplace does not have yet.

`pnpm compile` passes in `packages/vscode`.

### DIFF
--- a/packages/vscode/src/extension.ts
+++ b/packages/vscode/src/extension.ts
@@ -78,6 +78,23 @@ export const { activate, deactivate } = defineExtension(() => {
   let pendingRestart = false;
   let lastActivationReason: string | undefined;
 
+  // Glint's language server and tsserver plugin require the TypeScript 5/6
+  // JS API, which the TypeScript 7 package does not ship. On a TS 7 workspace
+  // they would only produce broken diagnostics, so Glint stands down entirely:
+  // template type-checking comes from a content mapper run by TypeScript
+  // itself instead.
+  const workspaceTypeScriptVersion = detectWorkspaceTypeScriptVersion(getLibraryPathSetting());
+  if (workspaceTypeScriptVersion && parseInt(workspaceTypeScriptVersion, 10) >= 7) {
+    outputChannel.appendLine(
+      `[Activation] TypeScript ${workspaceTypeScriptVersion} is installed in this workspace. ` +
+        `Glint requires TypeScript 5 or 6, so the Glint language server and tsserver plugin will not start. ` +
+        `With TypeScript 7, template type-checking comes from a content mapper instead: add ` +
+        `ember-content-mapper (https://github.com/NullVoxPopuli/ember-content-mapper) to "contentMappers" ` +
+        `in tsconfig.json and run tsc with --runExternalCode.`,
+    );
+    return volarLabs.extensionExports;
+  }
+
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

NullVoxPopuli-ai-agent:
Pushed a second commit (897545a) that makes the stand-down useful rather than just quiet: on TypeScript 7 workspaces the extension now activates TypeScript (Native Preview) and calls its `registerContentMappers('typed-ember.glint2-vscode', [{ extensions: ['.gts', '.gjs'] }])`.

Why it matters: the native server only tracks documents whose extensions were registered with it, so without this a content-mapped `.gts` file never gets `didOpen` and shows no hover, completions, or diagnostics in the editor even though `tsc --runExternalCode` checks it fine on the CLI (confirmed against ember-content-mapper's example app: Neovim, which sets filetypes directly, worked; VS Code showed nothing until the registration). Registering also lets the server discover the project's tsconfig from a `.gts` file alone, so users no longer need to open a `.ts` file first.

Outcomes are logged to the Glint output channel: Native Preview not installed, installed build too old to support registration (needs a build newer than 2026-08-19; the marketplace is still on 0.20260708.2, so today this requires a build from typescript-go main), or registered. The API is typed loosely in the extension (`registerContentMappers?` on the activate() result) to avoid a dependency on Native Preview's types.

NullVoxPopuli-ai-agent:
Correction: the commit mentioned above was pushed after this PR had already merged, so it never landed here. The registration change is now its own PR: https://github.com/typed-ember/glint/pull/1230.

